PR #231, opened 2015-01-16
Fix debug output to always include config file

lehmannro:
Previously, `location` was a NULL pointer when the check succeeded in `/etc`.  This led to the strange message:
```
--looking for config file at /home/robert/etc/libvmi.conf
--looking for config file at /root/etc/libvmi.conf
--looking for config file at /etc/libvmi.conf
**Using config file at
```

### DIFF
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -55,7 +55,7 @@ open_config_file(
     /* first check home directory of sudo user */
     if ((sudo_user = getenv("SUDO_USER")) != NULL) {
         if ((pw_entry = getpwnam(sudo_user)) != NULL) {
-            location = g_malloc0(snprintf(NULL,0,"%s/etc/libvmi.conf",
+            location = safe_malloc(snprintf(NULL,0,"%s/etc/libvmi.conf",
                                           pw_entry->pw_dir)+1);
             sprintf(location, "%s/etc/libvmi.conf",
                      pw_entry->pw_dir);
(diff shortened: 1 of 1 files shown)


libvmi/core.c
```diff
@@ -77,22 +77,26 @@ open_config_file(
...
 
     if (f) {
         goto success;
     }
+    free(location);
 
     /* finally check in /etc */
     dbprint(VMI_DEBUG_CORE, "--looking for config file at /etc/libvmi.conf\n");
-    f = fopen("/etc/libvmi.conf", "r");
+    location = g_malloc0(strlen("/etc/libvmi.conf")+1);
```
bdpayne: If you use g_malloc0 then you need to use g_free to cleanup that memory later (not free).  I see you have used free a couple of times below.

For the most part, we use safe_malloc in libvmi (a macro unique to this project).  However, I just realized that several g_malloc0 calls have slipped in as well.  And those other g_malloc0's appear to be using the wrong free.  Alas, I can clean those up shortly.

For the purposes of this PR, it would be nice to just stick with safe_malloc and then the calls to free are all ok as they are.

bdpayne:
Thanks for the contribution!  I have one comment on your code.  Beyond that.. please note that our automated PR checking system is currently down due to some data center issues.  Once that is up and running, we'll run the PR checker on this PR and then we can proceed with merging it.

libvmi/core.c
tklengyel: AFAIK the only thing g_free does that differs from free is that it checks if the pointer is null first. So IMHO there isn't a hard requirement to use it.

bdpayne: Matching g_free with g_malloc provides better portability and allows one to benefit from some memory tracking features for debugging memory leaks more easily.

lehmannro:
I have replaced all `g_malloc0`s with `safe_malloc`.

bdpayne:
Thanks, this looks good.  Hopefully I can get the test infrastructure running again shortly and then we can proceed with merging this.  Cheers!

bdpayne:
retest this please

bdpayne:
(note the above comment should kick in the PR testing system... let's see if it works)

bdpayne:
retest this please

bdpayne:
PR tester is happy, I'll proceed with the merge now.  Thanks again.

PR #234, opened 2015-02-13
Fixes qemu-kvm-patch

clarkb7:
Removes memory corruption
Fixes write requests


bdpayne:
this is ok to test

bdpayne:
thanks for the contribution clarkb7, nice catch

PR #239, opened 2015-03-13
Invalidate page cache with Xen after a page is written to

tklengyel:


libvmi/driver/memory_cache.c
```diff
@@ -297,6 +314,15 @@ memory_cache_insert(
     }
 }
 
+void memory_cache_remove(
```
bdpayne: Why is this defined a second time here?

awsaba: I think there has to be 2 versions of these functions, one for with page cache and one without, it's just hard to see the #else at L285.  Maybe a block comment or something around there will help call it out?

bdpayne: Ah yes, you're right.  Funny, I was looking for that else and just scanned right over it.  Must be early still ;-)

bdpayne:
+1, would like to see one more set of eyes before we merge

libvmi/driver/memory_cache.c
tklengyel: @awsaba is correct.

PR #241, opened 2015-03-21
Sanitize public headers

tklengyel:
The goal of this PR is to clean up the public headers and split sub-functionality out from libvmi.h into separate headers. The two extra headers split from libvmi.h are events.h that contains the events API, and shm.h which contains the public SHM interface. The shm API was erroneously just wrapped into an ifdef that relied on LibVMI internal config.h. Thus in this PR we change the build system to only install these extra headers if LibVMI was built with the appropriate flags.


bdpayne:
Changes look good to me, I'll merge once the PR tester is green.

PR #244, opened 2015-03-25
Fix Windows initialization when the CR3 points to a table that hasn't be...

tklengyel:
...en fully populated yet


awsaba:
+1.  I think this is similar to #178, but for init_from_sysmap.

tklengyel:
@awsaba: yea.. the way I was finally able to trigger it is by crafting a call to CreateProcessA and pausing the VM right when the new CR3 is loaded. At this stage I couldn't get LibVMI to reinit because the kernel pages don't seem to be in this new pagetable yet.

PR #246, opened 2015-04-01
Fix some more coverity bugs

tklengyel:
CID 54271 & CID 54278

### DIFF
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -158,7 +158,6 @@ status_t read_config_string(vmi_instance_t vmi,
     sprintf(config_str, "%s %s", vmi->image_type, config);
 
     config_file = fmemopen(config_str, length, "r");
-
     ret = read_config_file(vmi, config_file);
 
     free(config_str);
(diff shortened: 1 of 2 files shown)


tklengyel:
This is still WIP..

tklengyel:
So basically VMI_CONFIG_STRING can only be used if one inits with VMI_INIT_PARTIAL first and then goes to vmi_init_complete/vmi_init_custom_complete. I'm wondering if we would want to deprecate the config string option to begin with as it has been superseded by the GHashTable method (which is a lot more intuitive IMHO).

awsaba:
I've never liked the way that vmi_instance.image_type was reused (misused?) for VMI_CONFIG_STRING, so I'd be in favor of removing the current implementation.

awsaba:
Is this still WIP?  I say we go ahead and merge it as-is and target removing `VMI_CONFIG_STRING` for the next release.

tklengyel:
It's good to go, let's open an issue for deprecating VMI_CONFIG_STRING.

PR #247, opened 2015-04-02
0 can be a valid physical address.

awsaba:
Leave it to the driver to fail if it's not.

I was playing around with other ways to dump guest memory and noticed the discrepancy at 0, which `xc_map_foreign_range` will map just fine, but can't be read by the `dump-memory` example.

I'd like to get a couple pairs of eyes on this in case there is some place where non-zero is assumed and not otherwise handled.  Using 0 as the key for the memory_cache seems to work fine.

### DIFF
--- a/libvmi/read.c
+++ b/libvmi/read.c
@@ -97,13 +97,13 @@ vmi_read(
 
         if(dtb) {
             paddr = vmi_pagetable_lookup(vmi, dtb, start_addr + buf_offset);
+            if (!paddr) {
+                return buf_offset;
+            }
         } else {
             paddr = start_addr + buf_offset;
         }
 
-        if (!paddr) {
-            return buf_offset;
-        }
 
         /* access the memory */
         pfn = paddr >> vmi->page_shift;
@@ -331,12 +331,11 @@ vmi_read_str(
         addr += len;
         if(dtb) {
             paddr = vmi_pagetable_lookup(vmi, dtb, addr);
+            if (!paddr) {
+                return rtnval;
+            }
         } else {
             paddr = addr;
-        }
-
-        if (!paddr) {
-            return rtnval;
         }
 
         /* access the memory */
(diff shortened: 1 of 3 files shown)


libvmi/read.c
```diff
@@ -97,13 +97,13 @@ vmi_read(
 
         if(dtb) {
             paddr = vmi_pagetable_lookup(vmi, dtb, start_addr + buf_offset);
+            if (!paddr) {
```
tklengyel: Of course what happens if the vaddr is actually mapped to paddr=0. Not likely to happen on any sane OS but possible..

libvmi/read.c
```diff
@@ -97,13 +97,13 @@ vmi_read(
...
+                return buf_offset;
+            }
         } else {
             paddr = start_addr + buf_offset;
         }
 
-        if (!paddr) {
```
tklengyel: On the other hand we have been restricting access to paddr 0 even wtih physical access directly so this is clearly an issue that needs to be fixed either way.

libvmi/read.c
tklengyel: Probably a better approach would be to have vmi_pagetable_lookup return an errno in case something went wrong during the translation, instead of relying on a potentially valid paddr=0 being an error condition.

awsaba: I considered changing it to use `vmi_pagetable_lookup_extended` directly, but that also uses `!info->paddr` to determine it's return status right now, so that issue affects far more code paths.

tklengyel: In that case we might need to extend the extended lookup to also return an error code for failed translations (when PTE ENTRY_PRESENT() is not true)

tklengyel: We just need to thread carefully to remain compatible with both x86 and ARM paging (ENTRY_PRESENT is for x86 only, though I'm sure ARM has something that's equivalent)

awsaba: That probably means pushing that logic down to the `v2p_arch` functions, so I propose merging this PR as it stands now and I'll work those changes into another PR.

tklengyel: +1, that's reasonable, let's just open an issue for this to not forget that v2p could result in p=0.

tklengyel:
I think the two patches in this PR could be merged into one, afterwards it would be good to.

awsaba:
Squashed and tests passing.

PR #251, opened 2015-04-03
Return a status_t from arch v2p functions.

awsaba:
This is a lot work for the edge of a virtual address being mapped to physical page 0.

I also changed `vmi_uv2p()` to call `vmi_pagetable_lookup_extended()` instead of `vmi_uv2p_nocache()`, which was calling `vmi_pagetable_lookup()`, which does try the cache.  This does not change any API, so callers would have to call `vmi_pagetable_lookup_extended()` to know that 0 was the valid paddr.

`vmi_pagetable_lookup_cache()` added for cases where:
 1. A return status is desirable (`vmi_pagetable_lookup()` does not do this)
 2. A cached value from the v2p cache is acceptable (`vmi_pagetable_lookup_extended()` does not do this)

This is used 3 times in various read/write functions to replace the call to `vmi_pagetable_lookup()`.

### DIFF
--- a/libvmi/accessors.c
+++ b/libvmi/accessors.c
@@ -335,17 +335,46 @@ GSList* vmi_get_va_pages(vmi_instance_t vmi, addr_t dtb) {
 
 addr_t vmi_pagetable_lookup (vmi_instance_t vmi, addr_t dtb, addr_t vaddr)
 {
+    addr_t paddr = 0;
 
-    page_info_t info = {0};
+    status_t status = vmi_pagetable_lookup_cache(vmi, dtb, vaddr, &paddr);
 
-    /* check if entry exists in the cachec */
-    if (VMI_SUCCESS == v2p_cache_get(vmi, vaddr, dtb, &info.paddr)) {
+    if (status != VMI_SUCCESS) {
+        return 0;
+    }
+
+    return paddr;
+}
+
+/*
+ * Return a status when page_info is not needed, but also use the cache,
+ * which vmi_pagetable_lookup_extended() does not do.
+ *
+ * TODO: Should this eventually replace vmi_pagetable_lookup() in the API?
+ */
+status_t vmi_pagetable_lookup_cache(
+    vmi_instance_t vmi,
+    addr_t dtb,
+    addr_t vaddr,
+    addr_t *paddr)
+{
+    status_t ret = VMI_FAILURE;
+    page_info_t info = { .vaddr = vaddr,
+                         .dtb = dtb
+                       };
+
+    if(!paddr) return ret;
+
+    *paddr = 0;
+
+    /* check if entry exists in the cache */
+    if (VMI_SUCCESS == v2p_cache_get(vmi, vaddr, dtb, paddr)) {
 
         /* verify that address is still valid */
         uint8_t value = 0;
 
-        if (VMI_SUCCESS == vmi_read_8_pa(vmi, info.paddr, &value)) {
-            return info.paddr;
+        if (VMI_SUCCESS == vmi_read_8_pa(vmi, *paddr, &value)) {
+            return VMI_SUCCESS;
         }
         else {
             v2p_cache_del(vmi, vaddr, dtb);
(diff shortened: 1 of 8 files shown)


tklengyel:
@awsaba: could you recap in the PR message what vmi_pagetable_lookup_cache is for? Also, what happens if we are compiling without the cache being enabled at all?

awsaba:
It just calls the empty cache functions, so it should do the same thing that all the other existing code does (e.g., `vmi_pagetable_lookup()`). There shouldn't be any direct references.

libvmi/accessors.c
```diff
@@ -335,10 +335,10 @@ GSList* vmi_get_va_pages(vmi_instance_t vmi, addr_t dtb) {
 
 addr_t vmi_pagetable_lookup (vmi_instance_t vmi, addr_t dtb, addr_t vaddr)
```
tklengyel: IMHO we should change this function to return status_t as well.

libvmi/accessors.c
```diff
@@ -380,15 +381,67 @@ status_t vmi_pagetable_lookup_extended(
...
+ * TODO: Should this eventually replace vmi_pagetable_lookup() in the API?
+ */
+status_t vmi_pagetable_lookup_cache(
+    vmi_instance_t vmi,
+    addr_t dtb,
+    addr_t vaddr,
+    addr_t *paddr)
+{
+    status_t ret = VMI_FAILURE;
+    page_info_t info = {0};
```
tklengyel: I think it's better to initialize the struct here with {.dtb=dtb, .vaddr=vaddr}. The memset to 0 is really not required (and btw you already did that once when initialized the struct with {0}).

tklengyel:
So I looked into the ARM paging docs, simply the switch case b00 indicates that the vaddr is not mapped into the process. So we could just make the default case in the arm v2p to be VMI_FAILURE, and update it to VMI_SUCCESS in the switch cases that are already defined.

libvmi/accessors.c
awsaba: That would make it the same as the `_cache()` version, so that could go away.  It's fine with me as long we can get the +1's for changing external API.

awsaba: ...and the code shouldn't be duped in any case, so making this just a wrapper to `vmi_pagetable_lookup_cache()` if we decide not to change the external API.

libvmi/arch/arm_aarch32.c
```diff
@@ -78,11 +78,15 @@ void get_fine_second_level_descriptor(vmi_instance_t vmi, uint32_t vaddr, page_i
...
-addr_t v2p_aarch32 (vmi_instance_t vmi,
+status_t v2p_aarch32 (vmi_instance_t vmi,
     addr_t dtb,
     addr_t vaddr,
     page_info_t *info)
 {
+    status_t status = VMI_FAILURE;
+
+    info->paddr = 0;
+    info->size = VMI_PS_UNKNOWN;
```
tklengyel: page_info_t is already memset to zero by the callers so no need to reset it there

libvmi/arch/intel.c
```diff
@@ -88,33 +90,41 @@ uint64_t pdba_base_pae (uint64_t pdpe)
...
+    *pgd_value = 0;
     *pgd_entry = pdba_base_nopae(pdpe) + pgd_index_nopae(vaddr);
-    dbprint(VMI_DEBUG_PTLOOKUP, "--PTLookup: pgd_entry = 0x%.8"PRIx64"\n", *pgd_entry);
     if(VMI_FAILURE == vmi_read_32_pa(instance, *pgd_entry, &value)) {
-        value = 0;
+        dbprint(VMI_DEBUG_PTLOOKUP, "--PTLookup: failed to read pgd_entry at = 0x%.8"PRIx64"\n", *pgd_entry);
+        return VMI_FAILURE;
     }
-    return value;
+    *pgd_value = value;
```
tklengyel: So while the end-result PA could be 0, I'm not sure if these values midway through the translation can. We tend to have an ENTRY_PRESENT check on all of these anyway later on, but a pure 0 entry here is probably a cause for a VMI_FAILURE.

awsaba: Since we only have one failure result, I think letting these fall through the the ENTRY_PRESENT checks is probably still fine.  If we were to do anything in the case that 0 is read, I'd say it should be an `errprint()` because it probably means the entry address was miscalculated or there is something unsupported by libvmi going on.

tklengyel: Yea sure, it is being caught either way so we can leave it as it is.

libvmi/accessors.c
awsaba: @bdpayne @smaresca I think this PR is close to getting merged, but we need some input on pushing the following change to the external API: making vmi_pagetable_lookup return status_t and use an output parameter for the return value: 

`status_t vmi_pagetable_lookup (vmi_instance_t vmi, addr_t dtb, addr_t vaddr, addr_t *paddr)`.

smaresca: @awsaba I'm not against changing the external API, especially for a case such as this given that it helps to remove ambiguity.  +1 from me, but will defer to @bdpayne

tklengyel:
@awsaba: can you rebase this PR? IMHO it's good to be merged afterwards.

PR #252, opened 2015-04-05
Turn on ARM Supersection detection

tklengyel:


libvmi/arch/arm_aarch32.c
```diff
@@ -127,9 +127,8 @@ addr_t v2p_aarch32 (vmi_instance_t vmi,
                 info->paddr = (info->arm_aarch32.fld_value & VMI_BIT_MASK(20,31)) | (vaddr & VMI_BIT_MASK(0,19));
             } else {
                 dbprint(VMI_DEBUG_PTLOOKUP, "--ARM AArch32 PTLookup: the entry is a supersection descriptor for its associated modified virtual addresses\n");
-                // TODO: supersections are unsupported right now (breaks ptlookup when included)
```
awsaba: @tklengyel What/how was this breaking before that is not an issue now?

tklengyel: Not sure, it works now in every case I tested it with.

awsaba:
If I am reading the v5 manual correctly, either
 * supersections are available if sub-pages are enabled, or
 * fine page tables are supported
Since bit 18, should be 0 in the not supported cases and the fields for sections are the same, I think this will be OK to merge.

awsaba:
I've got 2 other questions about the existing code for fine page tables that now might be a good time ask about:

https://github.com/libvmi/libvmi/pull/252/files#diff-20908eeda83b03d307be38c4d1ee5e23R71
Should this be `<< 2` instead of '| 0b11'?

https://github.com/libvmi/libvmi/pull/252/files#diff-20908eeda83b03d307be38c4d1ee5e23R145
Should this reference `sld_value` instead of `fld_value`?

tklengyel:
@awsaba: yeap and yeap. Both seem to be bugs =)

PR #255, opened 2015-04-13
Make domids explicitly 64-bit wide

tklengyel:
LibVMI interprets the incoming domid as unsigned long (which may be 64-bit wide). Thus, if the user inits via GHashTable input with a 32-bit domid, and the 32-bit after the domid is not 0, LibVMI will fail to init. Making domid explicitly 64-bit within LibVMI avoids this confusion.

### DIFF
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -239,7 +239,7 @@ static status_t
 set_driver_type(
     vmi_instance_t vmi,
     vmi_mode_t mode,
-    unsigned long id,
+    uint64_t id,
     const char *name)
 {
     if (VMI_AUTO == mode) {
@@ -277,7 +277,7 @@ static status_t
 set_id_and_name(
     vmi_instance_t vmi,
     vmi_mode_t mode,
-    unsigned long id,
+    uint64_t id,
     const char *name)
 {
 
@@ -357,7 +357,7 @@ static status_t
 vmi_init_private(
     vmi_instance_t *vmi,
     uint32_t flags,
-    unsigned long id,
+    uint64_t id,
     const char *name,
     vmi_config_t config)
 {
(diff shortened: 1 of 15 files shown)


libvmi/core.c
```diff
@@ -565,13 +565,13 @@ vmi_init_custom(
     } else if (VMI_CONFIG_GHASHTABLE == config_mode) {
 
         char *name = NULL;
-        unsigned long domid = VMI_INVALID_DOMID;
+        uint32_t domid = VMI_INVALID_DOMID;
         GHashTable *configtbl = (GHashTable *)config;
         gpointer idptr = NULL;
 
         name = (char *)g_hash_table_lookup(configtbl, "name");
         if(g_hash_table_lookup_extended(configtbl, "domid", NULL, &idptr)) {
```
awsaba: This is kind of general question about the use of `VMI_CONFIG_GHASHTABLE`, but is there a definitive place other than looking at the internal code for someone to find out what the expected types of the values should be?

tklengyel: Not really at the moment..

smaresca: I nominate for a PR in parallel with this one, Maybe dropped in notes/  (??)

In libvmi/core.c
<https://github.com/libvmi/libvmi/pull/255#discussion_r28254189>:

> @@ -565,13 +565,13 @@ vmi_init_custom(
>      } else if (VMI_CONFIG_GHASHTABLE == config_mode) {
>
>          char *name = NULL;
> -        unsigned long domid = VMI_INVALID_DOMID;
> +        uint32_t domid = VMI_INVALID_DOMID;
>          GHashTable *configtbl = (GHashTable *)config;
>          gpointer idptr = NULL;
>
>          name = (char *)g_hash_table_lookup(configtbl, "name");
>          if(g_hash_table_lookup_extended(configtbl, "domid", NULL, &idptr)) {

Not really at the moment..

—
Reply to this email directly or view it on GitHub
<https://github.com/libvmi/libvmi/pull/255/files#r28254189>.

tklengyel:
So the basic config parser just converts the values defined in libvmi.conf to uint64_t, so that's what the pointers should point to when passed via GHashTable input as well. I guess we ought to make domid 64-bit wide as well just to make it more coherent with the rest of the values.

awsaba:
There are places where some Xen APIs declare domid to be uint16_t, so the value is going to get cast down somewhere eventually, so I think consistency in the sized types used in the GHashTable is a reasonable compromise.  Drivers could warn or error if they get a non-sane value, I guess.

tklengyel:
@awsaba: added the check in the Xen driver to see if the domainid is invalid (larger then what domid_t can carry)

PR #259, opened 2015-04-15
Fixing format warning in va-pages example

awsaba:
I believe `enum` should always be native int sized.


tklengyel:
-1, all page size declaration (page_size_t) are with ULL, thus the underlying storage is forced to be 64-bit. Without ULL it's up to the compiler to choose a storage mechanism, which could really be anything and not just int.

awsaba:
I changed this to extend whatever the compiler chooses to uint64_t to match the PRIx64, but FYI, the ULL in the enum declaration are apparently not forcing my version of gcc to do anything: https://gist.github.com/awsaba/b23386ee18bf0830558e

tklengyel:
Huh interesting and good to know =) Thanks!

Issue #260, opened 2015-04-16
Error mapping PE Export Table


tklengyel:
On certain Win7 SP1 x64 images mapping the export table of DLLs fails with the following log:

--PEParse: magic is 0x20b
--PEParse: DLL base [VA] 0x0000000076c10000. Export header [RVA] 0x000000000009fffc. Size 44279.
--PID cache hit 2132 -- 0x0000000047f33000
--PTLookup: lookup vaddr = 0x0000000076cafffc
--PTLookup: dtb = 0x0000000047f33000
--PTLookup: pml4e_address = 0x0000000047f33000
--MEMORY cache hit 0x47f33000
--PTLookup: pml4e = 0x00c0000047f37867
--PTLookup: pdpte_address = 0x0000000047f37008
--MEMORY cache hit 0x47f37000
--PTLookup: pdpte = 0x01500000485be867
--PTLookup: pde_address = 0x00000000485bedb0
--MEMORY cache hit 0x485be000
--PTLookup: pde = 0x07100000492ec867
--PTLookup: pte_address = 0x00000000492ec578
--MEMORY cache hit 0x492ec000
--PTLookup: pte = 0x0000000000000000
--PTLookup: paddr = 0x0000000000000000
--PEParse: failed to map export header
--PEParse: failed to get export table

The VA where the export table should be doesn't seem to be mapped into the process, verified via Volatility as well: 

In [1]: cc(pid=2132)
Current context: taskmgr.exe @ 0xfffffa8001ed3330, pid=2132, ppid=1052 DTB=0x47f33000
In [2]: dd(0x0000000076cafffc)
Memory unreadable at 76cafffc

However, when dumping the DLL via Volatility, the carved DLL does contain the correct export table. Root cause of the issue is TBD.

kirschju:
.rdata:0000000076CAFFFC ; Export directory for KERNEL32.dll
.rdata:0000000076CAFFFC ;
.rdata:0000000076CAFFFC                 dd 0                    ; Characteristics
.rdata:0000000076CB0000                 dd 53158EADh            ; TimeDateStamp: Tue Mar 04 08:28:29 2014
.rdata:0000000076CB0004                 dw 0                    ; MajorVersion
.rdata:0000000076CB0006                 dw 0                    ; MinorVersion
.rdata:0000000076CB0008                 dd rva aKernel32_dll    ; Name
.rdata:0000000076CB000C                 dd 1                    ; Base
[...]

PR #261, opened 2015-04-16
Fix cross-page export table mapping issue

tklengyel:
Fixes issue #260.

### DIFF
--- a/libvmi/os/windows/peparse.c
+++ b/libvmi/os/windows/peparse.c
@@ -513,10 +513,28 @@ peparse_get_export_table(
          export_header_va, ((windows_instance_t)vmi->os_data)->ntoskrnl_va,
          export_header_rva);
 
-    nbytes = vmi_read_va(vmi, export_header_va, pid, et, sizeof(*et));
+    nbytes = vmi_read_va(vmi, export_header_va, pid, et, sizeof(struct export_table));
     if (nbytes != sizeof(struct export_table)) {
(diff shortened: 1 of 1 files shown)


libvmi/os/windows/peparse.c
```diff
@@ -516,7 +516,25 @@ peparse_get_export_table(
...
     if (nbytes != sizeof(struct export_table)) {
         dbprint(VMI_DEBUG_PEPARSE, "--PEParse: failed to map export header\n");
-        return VMI_FAILURE;
+
+        /*
+         * Sometimes Windows maps the export table on page-boundaries,
+         * such that the first export_flags field (which is reserved) is.
+         * not actually accessible (the page is not mapped). See Issue #260.
+         */
+        if (export_header_va >> 12 != (export_header_va+4) >> 12) {
```
awsaba: Is there anything to check that the addresses here are 4-byte aligned?  If not, maybe something like `(export_header_va + 4) & ~(0xfff)` could be used to include that as a sanity check?

tklengyel: It's not really the alignment that matters - it's whether the struct is is on a page-boundary with the first uint32_t mapped into another page.

awsaba: If the special case is "export_header_va is exactly 4 bytes below a page boundary", then that's what the check should be for, not "4 or less".

tklengyel: @awsaba: sure. The check you propose is a bit more restrictive but it does match what we currently see so I'll make the adjustment.

PR #263, opened 2015-04-28
Xen 4.6 support

tklengyel:
Still WiP, RFCing mainly to see if I have broken anything on different Xen versions.


smaresca:
Regarding x86_registers_t, I'm assuming it's the receiving side for the stuff provided w/ a given vm_event request, ultimately populated by hvm_event_fill_regs()?

tklengyel:
@smaresca: yeap, want to eventually add the same feature to Xen on ARM too..

secana:
Does LibVMI work with XEN 4.5? We are running at with 4.4.1 and backported patches from 4.5 which is very unstable and not a long time solution.

tklengyel:
So this branch is currently working 100% for my usecase with DRAKVUF on Xen 4.6 RC1.

tklengyel:
So to ease the reviewing a bit, I think it's easier to check xen_events.c as a wholly new file, while the current xen_events.c has just been renamed xen_events_legacy.c (which is what gets built for everything up to Xen 4.5). It looks a bit ugly here on Github.

awsaba:
@tklengyel Are there any other changes that need to be made for the final Xen 4.6 release?

tklengyel:
This branch already works pretty well. There are a bunch of optimizations we could do on Xen 4.6 but those are not critical.

PR #266, opened 2015-05-05
Enable building LibVMI with KVM support on machines that don't have the ...

tklengyel:
...kernel module loaded

More robust test should be devised to check if the required KVM patches are in place then this one.


awsaba:
This is a great idea, and should make it easier for downstream builds, so I suggest pulling it into the current for-release PR.

PR #268, opened 2015-05-21
Make symbol in vmi_translate_sym2v const char*

tklengyel:


smaresca:
+1, will merge after test pass

PR #270, opened 2015-05-29
added kernel address to symbol translation for linux

bentau:
* the rva cache implementation did not work for me -> changed hash function (seems to be a bugfix?)

### DIFF
--- a/libvmi/os/linux/symbols.c
+++ b/libvmi/os/linux/symbols.c
@@ -65,7 +65,6 @@ get_symbol_row(
             while (curpos < MAX_ROW_LENGTH) {
                 if (isspace(row[curpos])) {
                     row[curpos] = '\0';
-                    break;
                 }
                 ++curpos;
             }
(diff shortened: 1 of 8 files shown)


libvmi/os/linux/symbols.c
```diff
@@ -140,3 +139,67 @@ linux_system_map_symbol_to_address(
...
+    FILE *f = NULL;
+    char *row = NULL;
+    char* address_str;
+    char* it;
+    char* symbol = NULL;
+    int size = 0;
+
+
+    if (pid != 0) {
+	//Not implemented, yet
```
tklengyel: Print an error message for this.

libvmi/os/linux/symbols.c
```diff
@@ -140,3 +139,67 @@ linux_system_map_symbol_to_address(
...
+    size = snprintf(NULL,0,"%"PRIx64"", address) + 1; 
+    address_str = g_malloc0(size);
+    snprintf(address_str, size, "%"PRIx64"", address);
+    if (get_symbol_row(f, row, address_str, 0) == VMI_FAILURE) {
+        goto done;
+    }
+
+    for(it=row; *it!=0; it++);
+    for(it++; *it!=0; it++);
+    it++;
```
tklengyel: Add a comment for this block describing what you are doing here.

tklengyel:
Also, please update the comment in libvmi.h regarding the use of v2sym, highlighting that it now supports looking up addresses for the Linux kernel using System.map, when the PID passed is 0.

tklengyel:
this is OK to test

tklengyel:
this is ok to test

bentau:
comments should be fixed

tklengyel:
One more thing, the v2sym function has thus far taken RVA, but for the System.map lookup it seems like you are just searching for the full VA (base_vaddr is completely unused). I think that makes sense in this case, but then the comment in the header explain that the function can either take a base_vaddr + address combination to parse the PE header found at address to search for the RVA, or simply an address with base_vaddr=0, pid=0 for lookup in System.map in Linux. This would also warrant renaming os_interface->os_rva2sym to os_interface->os_v2sym as it is no longer just for rva.

bentau:
I am not sure whether os_rva2sym or os_v2sym is the correct naming as I am not familiar with the windows naming. I guess that you should decide how the function should be called.
I fixed the comment in libvmi.h to fit your comments.

tklengyel:
The naming of the function on the os_interface right now reflected that it only supports RVAs. Since that's no longer the case, matching it with the public v2sym naming would be appropriate. Please rename it as part of this patch.

bdpayne:
lgtm

@tklengyel If you are happy with this then we can go ahead and merge.

tklengyel:
Yeap, looks good! Thanks! ;)

Issue #258, opened 2015-04-15
Issues accessing memory in guests with > 2GB of RAM


bdpayne:
From a mailing list conversation, it sounds like we may have an issue accessing memory in guests with more than 2GB of RAM.  I'm not sure of the root cause, but this is a placeholder to investigate.

Mailing list discussion is at:
https://groups.google.com/forum/#!topic/vmitools/mJKxLohpJ78

awsaba:
[This gist](https://gist.github.com/awsaba/5c807a8d07582c0f5770) shows the issue for Xen HVMs.  In particular Win 7 x64 uses pages in the 0x00100000-0010fbff range, which fail in get_memory() as being above the size of the RAM calculated using memkb. Changing that check for live VMs to use the result of "xc_get_maximum_gpfn()" should fix the issue for live VMs.

I'm not sure how to address the API to expose this for tools such as `dump-memory` itself, since using `max_gpfn` almost doubles the size for the 2GB VM, for not a lot of pages.  Options I see are:
1. Use the E820 map to stop dumping when the remaining regions are reserved
2. Support (and encourage the use of) non-raw, formatted files that leave out the 0's for unmapped pages.
3. Warn people in the documentation, and let them truncate stuff they don't need from the end of the file if they don't need it.

tklengyel:
Is there an API to access the E820 map?

awsaba:
I saw it while skimming `xenctrl.h` (which is how I found `xc_get_maximum_gpfn()`), which is the extent of what I know about, or if it will be populated for guests using EFI based BIOS, etc.

tklengyel:
Ah, excellent: xc_get_machine_memory_map

tklengyel:
Yea, we should just use this map and return VMI_FAILURE for regions known to be inaccessible.

smaresca:
VMI_FAILURE or a new status_t to indicate inaccessible memory?

tklengyel:
What I meant is that there is no reason to actually try to map memory when we already know its inaccessible. We should use xc_get_maximum_gpfn() for the upper bound and skip the holes automatically for performance boost. We have similar holes on guest memory in ARM as well that we could automatically return VMI_FAILURE for (with probably a debug message saying the access is skipped because its a known hole). I think for dump-memory it's fine to have the memory with the holes in it, even if that means that the file is going to be bigger then necessary.

smaresca:
Right certainly, but useful to indicate to the caller so (for example)
[...]

PR #282, opened 2015-08-12
Add max_physical_addressable value distinct from memsize.

awsaba:
This is an attempt to address problems like #258, by keeping track of an additional attribute of the guest, maximum_physical_address, which is different from the maximum amount of RAM allocated to a guest depending on how the hypervisor maps it into the guests physical address space.

### DIFF
--- a/examples/dump-memory.c
+++ b/examples/dump-memory.c
@@ -69,7 +69,7 @@ main(
         goto error_exit;
     }
 
-    size = vmi_get_memsize(vmi);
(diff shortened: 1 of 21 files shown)


awsaba:
So there is some different behavior with respect to the max page counts for Xen HV vs PV domains, and there doesn't seem to be an easy answer for KVM, so I'm hoping to settle for improving things for Xen HVM and having all other targets work about the same as they did before.

tklengyel:
LGTM +1

PR #297, opened 2015-11-23
added size_t size to vmi_instance struct

richardjuan:
vmi_instance->size  was missing for xen with snapshot enabled


tklengyel:
vmi->size has been deprecated and apparently we forgot to update some references to it in non-standard (experimental) builds. It has been replaced by max_physical_address and allocated_ram_size, either of which should be used in the appropriate context depending on what is actually being checked.

awsaba:
Closing this PR and opened #302 to track.

PR #299, opened 2015-12-08
Fix to pyvmi va* and pa* write functions

mbushou:
Ref the "Unable to write memory at specified address" exception:  I think the likely cause of the error was a 64-bit vaddr/paddr being crammed into a unsigned int by PyArg_ParseTuple.  This fix replaces the PyArg_ParseTuple "I" format string (unsigned int) with "K" (unsigned PY_LONG_LONG) to use with memory addresses.  I used "K" for consistency because other functions in pyvmi.c use this type for the same purpose (see pyvmi_read_va for example).

### DIFF
--- a/tools/pyvmi/pyvmi.c
+++ b/tools/pyvmi/pyvmi.c
@@ -510,7 +510,7 @@ pyvmi_write_pa(
     void *buf;
     int count;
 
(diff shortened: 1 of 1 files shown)


awsaba:
this is ok to test

awsaba:
retest this please

PR #301, opened 2015-12-09
Allow setting MEMACCESS permissions in specified VMM pagetable

tklengyel:


tklengyel:
retest this please

libvmi/driver/xen/xen_events.c
```diff
@@ -862,7 +862,10 @@ status_t xen_set_mem_access(vmi_instance_t vmi, mem_access_event_t *event,
     dbprint(VMI_DEBUG_XEN, "--Setting memaccess for domain %"PRIu64" on physical address: %"PRIu64" npages: %"PRIu64"\n",
         dom, event->physical_address, npages);
 
-    rc = xc_set_mem_access(xch, dom, access, page_key, npages);
+    if ( !altp2m_idx )
```
awsaba: I'm pretty sure 0 is a valid altp2m index.  I think this needs some other value (-1?) or an additional flag based on the presence of `VM_EVENT_FLAG_ALTERNATE_P2M`.

Unless xc_set_mem_access() is equivalent to `xc_altp2m_set_mem_access(...,..., 0, ..., ...)`, which is unclear.  I'm not finding these APIs very well documented.

tklengyel: 0 is treated specially by Xen as the base pagetable. All other altp2m views take 0 as the base and get populated lazily in a copy-on-write fashion when using xc_altp2m_set_mem_access. So for 0 you should keep using the same xc_set_mem_access routines as there is no difference for that view from a mem_access perspective.

Some documentation is available at http://lists.xenproject.org/archives/html/xen-devel/2015-06/msg01319.html

tklengyel: Also see http://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=xen/arch/x86/mm/p2m-ept.c#l854
When you change permissions on the hostp2m (0) it gets automatically propagated to all altp2m views as well.

tklengyel: Furthermore, you can't destroy altp2m 0:
http://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=xen/arch/x86/mm/p2m.c#l2335

Interestingly you can send 0 for xc_altp2m_set_mem_access but that actually may be a bug..

libvmi/driver/xen/xen_events.c
```diff
@@ -376,7 +376,7 @@ status_t process_unhandled_mem(vmi_instance_t vmi,
     mem_access_event_t event = { 0 };
     event.physical_address = page->key << 12;
     event.npages = 1;
-    xen_set_mem_access(vmi, &event, VMI_MEMACCESS_N);
+    xen_set_mem_access(vmi, &event, VMI_MEMACCESS_N, req->altp2m_idx);
```
awsaba: Is `req->altp2m_idx` guaranteed to be 0 if VM_EVENT_ALTERNATE_P2M is cleared?  "may be used" (from http://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=xen/include/public/vm_event.h;h=9270d52592af52de51cfde07bc492a504412edb2;hb=HEAD#l241) sounds underspecified to me.

tklengyel:
retest this please

libvmi/driver/xen/xen_events.c
tklengyel: By "may be" what I meant there is that it "can be" used for altp2m. It's always 0 if VM_EVENT_ALTERNATE_P2M is not set.

tklengyel:
IMHO this is also good to be merged.

awsaba:
retest this please

PR #300, opened 2015-12-08
Fix event response flag setting in xen events driver

tklengyel:

PR #303, opened 2015-12-11
Ensure safety of vmi_clear_event when issued from event callbacks

tklengyel:
When we are monitoring a VM with multiple vCPUs there could be multiple requests triggered by an event on the ringpage already. When the handler issues a vmi_clear_event in the callback of the first request, the second request will be left without a handler and LibVMI will shutdown as it doesn't know what to do. With this PR we queue all vmi_clear_event requests that happen in event callbacks so that the driver can safely process them when it can ensure that no more requests are left on the ring. This can only be done by pausing all vCPUs, processing any leftover requests that may have been placed there since last time we checked, then removing each event that was queued.

### DIFF
--- a/libvmi/driver/xen/xen_events.c
+++ b/libvmi/driver/xen/xen_events.c
@@ -370,7 +370,9 @@ status_t process_register(vmi_instance_t vmi,
         }
 
         event->vcpu_id = req->vcpu_id;
+        vmi->event_callback = 1;
         process_response ( event->callback(vmi, event), event, rsp );
+        vmi->event_callback = 0;
 
         return VMI_SUCCESS;
     }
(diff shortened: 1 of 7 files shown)


tklengyel:
retest this please

tklengyel:
retest this please

libvmi/driver/xen/xen_events.c
```diff
@@ -1159,10 +1134,81 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+
+        ghashtable_foreach(vmi->clear_events, i, key, cb) {
+            vmi_clear_event(vmi, *key, cb);
+        }
+
+        g_hash_table_destroy(vmi->clear_events);
+        vmi->clear_events =
+            g_hash_table_new_full(g_int64_hash, g_int64_equal, free, NULL);
+
+        vmi_resume_vm(vmi);
```
tklengyel: This actually may be problematic because it will also unpause the vCPU that was paused by the vm_event, before the response was pulled form the ring on the Xen side..

tklengyel:
This will have to be rebased once PR #300 is merged.

tklengyel:
I've rebased this branch so it's now good to be merged.

PR #305, opened 2015-12-21
Enabled tinyints as input to pyvmi write_8_* functions

mbushou:
Ref Bryan's mailing list post on 12/08/15:
Question: The va, pa, and ksym 8-bit write functions accept single characters, should they also accept 8-bit integers?
Bryan: Yes, they should take 8-bit integers.

This patch adds a second PyArg_ParseTuple function call inside pyvmi.c "write_8_*" functions.  The second PyArg_ParseTuple tries to format the incoming value as a tinyint, and only executes if the first (existing) call fails to successfully extract a character.

See attached for before and after behavior.

[write_8_testing.py.txt](https://github.com/libvmi/libvmi/files/68657/write_8_testing.py.txt)

### DIFF
--- a/tools/pyvmi/pyvmi.c
+++ b/tools/pyvmi/pyvmi.c
@@ -1012,7 +1012,9 @@ pyvmi_write_8_pa(
     addr_t paddr;
     uint8_t value;
 
-    if (!PyArg_ParseTuple(args, "Kc", &paddr, &value)) {
+    if (!PyArg_ParseTuple(args, "Kc", &paddr, &value) &&
+        !PyArg_ParseTuple(args, "KB", &paddr, &value)
+    ) {
         PyErr_SetString(PyExc_ValueError,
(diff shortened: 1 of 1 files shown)


awsaba:
@mbushou @tklengyel Am I correct in thinking the thing holding up this PR is the `configure.ac` change that was addressed elsewhere?

tklengyel:
@awsaba: yes it seems like

mbushou:
Yes.  I made a mistake and the configure.ac PR should have been separate.
Anything I need to do on my end?

awsaba:
@mbushou Yes, you'll need to either:
 * Reset the branch to just include the first commit, or
 * Revert the bad commit, then squash them.
And then force push your branch, which will update the PR.

mbushou:
@awsaba Thanks, done, ready for another review.

bdpayne:
Thanks!

PR #308, opened 2016-01-05
Properly issue clear_event callback

tklengyel:


awsaba:
Creating a typedef for `vmi_event_free_t` may help with readability, even if no more parameters are added.

tklengyel:
@awsaba: good call, made the change

PR #312, opened 2016-02-03
Ensure safety of shutting down xen events

tklengyel:
Failure to do so can cause a hypervisor crash.


tklengyel:
retest this please

PR #313, opened 2016-02-11
Allow using Rekall profiles for Linux guests

tklengyel:
In this patch we move the Rekall profile component from the Windows driver to common and convert the JSON parsing from JANSSON to JSON-C, as JANSSON is unable to handle large VAs stored in the Linux profile. The Rekall profile for Linux needs to be generated with Rekall after patch 7668d50b129b4cc38048819e38ee0b0cee794f75 (https://github.com/google/rekall/pull/113).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -218,7 +218,7 @@ update is made to the kernel, the profile has to be re-generated, thus it's a bi
 as the standard LibVMI configuration entries which are generally stable for specific releases
 of Windows.
 
-Rekall is available at https://code.google.com/p/rekall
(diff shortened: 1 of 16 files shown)


awsaba:
1. How easy is it to get a compatible copy of json-c?  That is, can I yum/apt-get install it on popular distros?
2. We should probably add a note for optional dependencies in the `README.md`.

tklengyel:
@awsaba: it's in  all of the standard repositories. On debian it's `apt-get install libjson-c-dev` and yes, I'll update the README.md anyway as the Rekall commands shown there are now outdated.

tklengyel:
@awsaba: I've updated the README so this should be good to go.

PR #315, opened 2016-02-24
Update outdated Google Code link to GitHub link in README.rst

ntddk:


README.rst
```diff
@@ -19,7 +19,7 @@ kernel symbols from Windows and Linux.
 
 If you would like higher level semantic information, then we suggest using 
 LibVMI with PyVMI (python wrapper, included with LibVMI) and Volatility.
-Volatility (http://code.google.com/p/volatility/) is a forensic memory analysis
+Volatility (https://github.com/volatilityfoundation/volatility) is a forensic memory analysis
```
awsaba: @tklengyel Rekall and Volatility seem pretty separate at this point, does it make more sense to mention them both or be more helpful to stick to Rekall since that what is used for its profiles?

tklengyel: We should probably mention both.

README.rst
```diff
@@ -260,11 +260,3 @@ make install
 
 The default installation prefix is /usr/local.  You may need to run
 'ldconfig' after performing a 'make install'.
-
```
awsaba: @bdpayne This was before my time, I have no idea if people are getting here by coming from XenAccess.  Maybe mention it in a new **History** section of the README?

README.rst
bdpayne: Agreed on both.  We use each one in different ways.  Not sure if that's "a good thing", but it is where we are at today.

README.rst
bdpayne: XenAccess is far enough in the past that I agree it is ok to remove it from here.  I don't think we need a history section... we can just remove it.

bdpayne:
When updating this for the comments above, please go ahead and rebase because https://github.com/libvmi/libvmi/pull/318 did some white space cleanup that is causing conflicts here.  Thanks!

smaresca:
@ntddk thanks again for the PR. Have you had a chance to rebase? If needed due to time constraints on your end, we could use your branch and make a separate PR. Let us know.

PR #317, opened 2016-02-24
Enable Xen events by default

tklengyel:


awsaba:
Looks like there's an unrelated bugfix in here, too, but let's get them both in, since the event API is stable enough at this point.

PR #320, opened 2016-03-16
Added vmi_get_linux_sysmap function

dalevy:
the new accessor function will return the value that a particular linux vmi instance has for the location of the system map file or NULL if otherwise. This should require the libvmi public api reference site to be updated as well as it is a new addition

### DIFF
--- a/libvmi/accessors.c
+++ b/libvmi/accessors.c
@@ -474,8 +474,9 @@ vmi_get_linux_sysmap(
 {
     linux_instance_t linux_instance = NULL;
 
-    if(VMI_OS_LINUX != vmi->os_type || (VMI_INIT_PARTIAL & vmi->init_mode))
+    if(VMI_OS_LINUX != vmi->os_type || (VMI_INIT_PARTIAL & vmi->init_mode)){
 	return NULL;
+    }
 
     if(!vmi->os_data){
 	return NULL;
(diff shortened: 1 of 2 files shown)


tklengyel:
This is ok to test

libvmi/accessors.c
```diff
@@ -467,3 +467,22 @@ addr_t vmi_translate_uv2p (vmi_instance_t vmi, addr_t virt_address, vmi_pid_t pi
...
     return paddr;
 }
+
+const char *
+vmi_get_linux_sysmap(
+	vmi_instance_t vmi)
+{
+    linux_instance_t linux_instance = NULL;
+
+    if(VMI_OS_LINUX != vmi->os_type || (VMI_INIT_PARTIAL & vmi->init_mode))
```
bdpayne: Nit: could you put curly brackets around this if clause (like you do in the following one)?

dalevy: Will do - going to submit another merge request.

bdpayne: FYI, you should be able to fix, commit, rebase / merge, and then push the updated branch into this same PR.  That is the preferred workflow.

dalevy: I did a git...add, commit, and push after the change. I hope that works out correct - I'm not too good at big git projects.

bdpayne:
@dalevy The benefit of doing the rebase is that you can squash this back down into 1 commit.  Having said that, it isn't critical in this situation, so no worries.

Thanks for the PR... merging now.

dalevy:
@bdpayne Thanks for the heads up. I'll keep that in mind next time :)

PR #326, opened 2016-03-29
Check for VM state before resuming it on events shutdown

tklengyel:
Without this check we may end up resuming a VM that was already paused.

### DIFF
--- a/libvmi/driver/xen/xen_events.c
+++ b/libvmi/driver/xen/xen_events.c
@@ -578,7 +578,7 @@ status_t process_single_step_event(vmi_instance_t vmi,
 
 void xen_events_destroy(vmi_instance_t vmi)
 {
-    int rc;
+    int rc, resume = 0;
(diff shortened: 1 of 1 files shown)


awsaba:
I think this may not be necessary because from at least 4.4 forward, whether or not a domain is paused/unpaused is controlled by a counter, but I haven't checked to see how far back it has been implemented that way: http://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=xen/common/domain.c;h=19e8470bc66e217de3a5cd504c6036377e2ab32a;hb=a611ed5d040a764e01edcc8b6d99342c393589d9#l832

tklengyel:
Thats what I thought too but on 4.6 and 4.7 this has been an active issue
that this PR fixes.

awsaba:
I'm going to go ahead and merge this.  The impact is minor and shouldn't have any affect if not necessary.

PR #330, opened 2016-04-18
Don't redefine xc_domain_maximum_gpfn

tklengyel:
On Xen 4.6 ARM builds we end up erroneously defining our own xc_domain_maximum_gpfn because the xen_cr3_to_pfn_x86_32 guard is not present and __XEN_INTERFACE_VERSION__ is >= 40600.

Thanks to @kittel for the bug-report.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -180,6 +180,7 @@ xen_event_space=' '
         enable_xen='no'
         have_xen='missing xenstore'
     [else]
+        AC_CHECK_LIB(xenctrl, xc_domain_maximum_gpfn, [AC_DEFINE([HAVE_XC_DOMAIN_MAXIMUM_GPFN],[1],"xenctrl has xc_domain_maximum_gpfn")],[missing="no"])
         AC_CHECK_LIB(xenctrl, xc_interface_open, [], [missing="yes"])
         [if test "$missing" = "yes"]
         [then]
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -180,6 +180,7 @@ xen_event_space=' '
         enable_xen='no'
         have_xen='missing xenstore'
     [else]
+        AC_CHECK_LIB(xenctrl, xc_domain_maximum_gpfn, [AC_DEFINE([HAVE_XC_DOMAIN_MAXIMUM_GPFN],[1],"xenctrl has xc_xc_domain_maximum_gpfn")],[missing="no"])
```
awsaba: Typo?  There is an extra `xc_` in the message.

libvmi/driver/xen/xen_private.h
```diff
@@ -55,8 +55,10 @@
 #define xen_pfn_to_cr3_x86_32(pfn) (((unsigned)(pfn) << 12) | ((unsigned)(pfn) >> 20))
 #define xen_cr3_to_pfn_x86_32(cr3) (((unsigned)(cr3) >> 12) | ((unsigned)(cr3) << 20))
 
+#ifndef HAVE_XC_DOMAIN_MAXIMUM_GPFN
```
awsaba: Since this is now it's own configure check, it might be less confusing to pull this outside of the `#ifndef xen_cr3_to_pfn_x86_32`

tklengyel: Sounds good.

Issue #331, opened 2016-05-12
Failed to read UNICODE_STRING buffer Windows x64


masthoon:
Sometimes when I try to read a Windows _\_UNICODE_STRING_ using _vmi_read_unicode_str_va_ function, it fails : 

> windows_read_unicode_struct: failed to read UNICODE_STRING buffer

My virtual machine is Windows 7 SP1 x64.

I think, this is due to the definition of _win64_unicode_string_t_ and the trick used in [_windows_read_unicode_struct_](https://github.com/libvmi/libvmi/blob/bf68c6e239b52c89504e30ff88d9de449a0f35f3/libvmi/os/windows/unicode.c#L55) to read unicode string for 64 bits Windows guest.

The definition of _win64_unicode_string_t_ in [libvmi/private.h](https://github.com/libvmi/libvmi/blob/master/libvmi/private.h#L185) can induce errors because of the **packed** attribute.

```c
/** Windows' UNICODE_STRING structure (x64) */
    typedef struct _windows_unicode_string64 {
        uint16_t length;
        uint16_t maximum_length;
        uint64_t pBuffer;   // pointer to string contents
    } __attribute__ ((packed))
    win64_unicode_string_t;
```

Currently, LibVMI reads raw unicode_string structure from memory (virtual address _vaddr_), then compute pBuffer virtual address using read pBuffer high order bytes (shifted) and high order bytes of _\_UNICODE_STRING_ virtual address (_vaddr_).

```c
buffer_va = (vaddr & 0xFFFFFFFF00000000) + (us64.pBuffer >> 32);
```

But high order bytes of _\_UNICODE_STRING_ structure can be different from the real address of pBuffer (especially in Kernel).

From Rekall for Win7SP1 x64, I got : (Buffer offset is 8)
```json
"_UNICODE_STRING": [16, {
   "Buffer": [8, ["Pointer", {
    "target": "unsigned short"
    }]], 
   "Length": [0, ["unsigned short", {}]], 
   "MaximumLength": [2, ["unsigned short", {}]]
   }]
```

By removing **packed** attribute to the structure and setting `buffer_va = us64.pBuffer;`, it's working as intended at least for my setup.

Do you think that my fix is okay for all versions of Windows ? I can do a pull request if you want.

Why does LibVMI reuse high order bytes of structure address in pBuffer virtual address ?

Thanks !

tklengyel:
Yes, please an open a PR for this! Thanks.

PR #332, opened 2016-05-13
Fix read UNICODE_STRING buffer x64

masthoon:
[Related issue](https://github.com/libvmi/libvmi/issues/331)

### DIFF
--- a/libvmi/private.h
+++ b/libvmi/private.h
@@ -183,11 +183,12 @@ typedef struct _windows_unicode_string32 {
     win32_unicode_string_t;
 
 /** Windows' UNICODE_STRING structure (x64) */
-    typedef struct _windows_unicode_string64 {
-        uint16_t length;
-        uint16_t maximum_length;
-        uint64_t pBuffer;   // pointer to string contents
-    } __attribute__ ((packed))
+typedef struct _windows_unicode_string64 {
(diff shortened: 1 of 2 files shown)


tklengyel:
This is ok to test.

libvmi/private.h
```diff
@@ -183,12 +183,11 @@ typedef struct _windows_unicode_string32 {
...
 /** Windows' UNICODE_STRING structure (x64) */
-    typedef struct _windows_unicode_string64 {
-        uint16_t length;
-        uint16_t maximum_length;
-        uint64_t pBuffer;   // pointer to string contents
-    } __attribute__ ((packed))
-    win64_unicode_string_t;
+typedef struct _windows_unicode_string64 {
+    uint16_t length;
+    uint16_t maximum_length;
```
tklengyel: We should have an explicit 32-bit padding field here so that we don't rely on the compiler replicating the padding of the microsoft compiler exactly.

masthoon:
I have added the padding field to the packed structure.

tklengyel:
Thanks, could you squash the two commits in this PR to just one? Then it's good to merge.

masthoon:
Please let me know if it's ok for you.

PR #339, opened 2016-05-21
Deprecate byte-granularity memaccess events

tklengyel:
Byte-granularity events have a particular weakness in that events triggered up to 8-bytes prior to the actual byte where the event is registered would not trigger a callback. Thus, the API as it is right now may encourage the development of insecure applications. As only the user is aware if such events are permissible or not depending on the use-case, it is better to leave it up to the user to determine what to do.


smaresca:
definitely in favor of this. Tested fine on my end. Merging

PR #353, opened 2016-07-25
Update PyVMI example to support linux

maxking:
This updates the pyvmi example to check for ostype and then get offsets specific to the os. Currently there is no check for the os type and the example errors out when calling `vmi.get_offset("win_tasks")` .


tklengyel:
This is ok to test.

maxking:
@tklengyel Looks like the tests pass :)

tklengyel:
Thanks, it looks good to me!

PR #357, opened 2016-08-15
LibVMI Xen backwards compatibility

### LINKED ISSUE #356: LibVMI backwards compatibility
Currently compiling LibVMI on a Xen 4.6+ host will automatically disable event support for earlier Xen versions. Wiring up the LibVMI events API should be done at run-time based on the running Xen version rather then during compile time.

tklengyel:
Currently LibVMI determines Xen compatibility during compile-time, requiring a user to compile a version of LibVMI for each version of Xen it needs to be deployed on. While this problem mostly affects the Xen events API which went through big changes, other libxc functions also have changed between various versions, making for example a LibVMI compiled on a Xen 4.6 incompatible with Xen 4.5 libraries and with the 4.5 events ABI.

In this PR we introduce a libxc wrapper using the dl library to load the appropriate functions for the appropriate Xen version we detect during run-time. With this we deprecate a lot of the preprocessor magic.

To properly add backwards compatibility support for Xen events we deprecate including the actual Xen event headers from the host and define the Xen events ABI locally. This allows us to compile support for all versions of Xen from Xen 4.2 upwards. However, the Xen ring API has changed substantially between Xen 4.1 and Xen 4.2, thus rather than defining a local copy of the Xen ring API as well we just deprecate support for Xen 4.1 events.

We also deprecate support for Xen 4.0 and older.

This fixes #356.

### DIFF
--- a/libvmi/driver/xen/libxc_wrapper.c
+++ b/libvmi/driver/xen/libxc_wrapper.c
@@ -1,0 +1,129 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+ * memory in a target virtual machine or in a file containing a dump of
+ * a system's physical memory.  LibVMI is based on the XenAccess Library.
+ *
+ * Author: Tamas K Lengyel <tamas.lengyel@zentific.com>
+ *
+ * This file is part of LibVMI.
+ *
+ * LibVMI is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.
+ *
+ * LibVMI is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <string.h>
+
+#include "xen_private.h"
+
+static inline status_t sanity_check(xen_instance_t *xen)
+{
+    status_t ret = VMI_FAILURE;
+    libxc_wrapper_t *w = &xen->libxcw;
+
+    if ( xen->major_version != 4 )
+        return ret;
+
+    switch ( xen->minor_version )
+    {
+        case 1:
+            if ( !w->xc_domain_maximum_gpfn )
+                break;
+
+            ret = VMI_SUCCESS;
+            break;
+        case 2:
+        case 3:
+        case 4:
+            if ( !w->xc_domain_maximum_gpfn || !w->xc_map_foreign_batch ||
+                 !w->xc_mem_access_enable || !w->xc_mem_access_disable ||
+                 !w->xc_mem_access_resume || !w->xc_hvm_set_mem_access ||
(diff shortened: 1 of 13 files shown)


awsaba:
One potential issue is that `xc_version()` returns the version of the running hypervisor, but not the version of the userspace tools that `dlopen()` might have found and loaded.  Given the Frankenstien's monster nature of dev machines when people start trying to use libvmi, is there a sanity check somewhere to make sure that some expected per-version APIs were actually found before calling them?

tklengyel:
@awsaba: That's true but usually if you deploy a version of libxc that doesn't match the version of Xen you tend to have bigger problems then LibVMI. We can do a per-version sanity check though to see if the expected functions are found for the reported Xen version, that would make sense.

tklengyel:
The PR has been tested on Xen 4.5 and Xen 4.8-unstable using events, everything is working as expected.

tklengyel:
I've added a couple more smaller fixes and cleanups as well as allowing setting the response flags on Xen 4.5 which first introduced this feature.

tklengyel:
Retest this please.

smaresca:
@tklengyel  you said "We can do a per-version sanity check though to see if the expected functions are found for the reported Xen version, that would make sense." -- what were you thinking for this?

libvmi/driver/xen/libxc_wrapper.c
```diff
@@ -0,0 +1,141 @@
...
+
+#include "xen_private.h"
+
+#define LIBXC_PATHS_TO_TRY 2
+static const char* libxc_paths[LIBXC_PATHS_TO_TRY] = {
+    "/usr/local/lib/libxenctrl.so",
+    "/usr/lib/libxenctrl.so",
+};
+
+static inline status_t sanity_check(xen_instance_t *xen)
```
tklengyel: @smaresca

libvmi/driver/xen/libxc_wrapper.c
```diff
@@ -0,0 +1,141 @@
...
+
+#include <string.h>
+
+#include "xen_private.h"
+
+#define LIBXC_PATHS_TO_TRY 2
+static const char* libxc_paths[LIBXC_PATHS_TO_TRY] = {
+    "/usr/local/lib/libxenctrl.so",
+    "/usr/lib/libxenctrl.so",
+};
```
smaresca: I'm concerned that we will run afoul of package maintainers with this approach..using LIBXC_PATHS_TO_TRY like above basically requires keeping a big list of possible paths for all distributions we want to support.  

Case in point: As-is this excludes Red Hat/Centos using the Citrix maintained Xen RPMs because they install libxenctrl into /usr/lib64. Also, those specific packages actually use versioned files too (libxenctrl.so.4.6 for example) so multiple can be present in parallel.

Any prefix of course is possible per compile-time directives for the xen libs.

I recommend looking at ld.so(8) for dynamic linker search order. Relying on system defined resolution (especially given the flexibility of /etc/ld.so.conf and /etc/ld.so.conf.d) is probably the safest route.

tklengyel:
@smaresca: right, there is no need to define paths for dlopen at all, it works in tandem with ld.so so it can find libxenctrl.so without the path listing here.

smaresca:
@tklengyel  addressed my comment about absolute, static paths used for dlopen() in create_libxc_wrapper(). The function is now dlopen()-ing libxenctrl.so using the standard lib resolution (relying upon ld.so.conf etc and LD_* for), so this will respect any system library prefix path used. Make this distribution agnostic.

smaresca:
+1, Merging

Issue #354, opened 2016-08-01
libvmi does not always initialize correctly with Windows 10 using Rekall profiles


CMF-Marodeur:
I added the required information for Windows 10 (analogue to tklengyels pull request #314), which generally works.
The problem is that libvmi does not always initialize (a problem i already faced when attaching to a Windows 7 image, but I did not investigate any further).

The problem seems to missing page table entries:

Working:
LibVMI Version 0.11.0
--found Xen
LibVMI Mode 2
--completed driver init.
--got id from name (domU-4 --> 33)
**set image_type = domU-4
**set hvm to true (HVM).
--xen: setup live mode
**set allocated_ram_size = 7fdde000, max_physical_address = 0xff000000
**set pae = 1
**set pse = 1
**set lme = 1
**IA-32e paging
**sanity checking cr3 = 0x000000002b75f000
--succesfully completed architecture init.
--looking for config file at (null)/etc/libvmi.conf
--looking for config file at /etc/libvmi.conf
**Using config file at /etc/libvmi.conf
**set os_type to unknown.
**Trying to init from Rekall profile
--PTLookup: lookup vaddr = 0xfffff801c4886000
--PTLookup: dtb = 0x0000000071162000
--MEMORY cache set 0x71162000
--xen_get_memory_pfn success on pfn=0x71162
--PTLookup: pml4e_address = 0x0000000071162f80
--MEMORY cache set 0x584000
--xen_get_memory_pfn success on pfn=0x584
--PTLookup: pdpte_address = 0x0000000000584038, pdpte_value = 0x0000000000585063
--MEMORY cache set 0x585000
--xen_get_memory_pfn success on pfn=0x585
--PTLookup: pde_address = 0x0000000000585120, pde_value= 0x0000000000593063
--MEMORY cache set 0x593000
--xen_get_memory_pfn success on pfn=0x593
--PTLookup: pte_address = 0x0000000000593430, pte_value = 0x8000000002c86963
--PTLookup: paddr = 0x0000000002c86000
--V2P cache set 0xfffff801c4886000 -- 0x0000000002c86000 (0x0000000071162000/0xfffff801c4886000)
**KernBase PA=0x2c86000
**KernBase VA=0xfffff801c4886000
--MEMORY cache set 0x2ed3000
--xen_get_memory_pfn success on pfn=0x2ed3
**init from Rekall profile success

Not working:
LibVMI Version 0.11.0
--found Xen
LibVMI Mode 2
--completed driver init.
[...]

PR #358, opened 2016-08-18
Added support for specifying win_ntoskrnl_va in the libvmi.conf.

CMF-Marodeur:
This is useful when restoring a virtual machine from a state file.

*grammar.y added win_ntoskrnl_va
*lexicon.l added win_ntoskrnl_va
*core.c added win_ntoskrnl_va, added sanity check to init_from_rekall_profile, try fallback initialisation if init_from_rekall_profile fails
*windows.h added defines for sanity check when using init_from_rekall_profile
*kdbg.c Allow initialisation with ntoskrnl + ntoskrnl_va

### DIFF
--- a/libvmi/os/windows/core.c
+++ b/libvmi/os/windows/core.c
@@ -84,14 +84,16 @@ get_ntoskrnl_base(
 {
     uint8_t page[VMI_PS_4KB];
     addr_t ret = 0;
+    access_context_t ctx = {
+        .translate_mechanism = VMI_TM_NONE,
+        .addr = page_paddr
+    };
 
-    for(; page_paddr + VMI_PS_4KB < vmi->max_physical_address; page_paddr += VMI_PS_4KB) {
+    for(; ctx.addr + VMI_PS_4KB < vmi->max_physical_address; ctx.addr += VMI_PS_4KB) {
 
         uint8_t page[VMI_PS_4KB];
-        status_t rc = peparse_get_image_phys(vmi, page_paddr, VMI_PS_4KB, page);
-        if(VMI_FAILURE == rc) {
+        if(VMI_FAILURE == peparse_get_image(vmi, &ctx, VMI_PS_4KB, page))
             continue;
-        }
 
         struct pe_header *pe_header = NULL;
         struct dos_header *dos_header = NULL;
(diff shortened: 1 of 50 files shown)


libvmi/os/windows/windows.h
```diff
@@ -23,6 +23,10 @@
 
 #include "private.h"
 
+#define LOWEST_KERNEL_ADDRESS_64BIT 0xffff800000000000
+#define LOWEST_KERNEL_ADDRESS_32BIT 0x80000000
```
tklengyel: The problem with this on 32bit machines is that its configurable. You can define at boot time what the boundary should be, so while most of the time this its fine it can also easily break.

CMF-Marodeur: Acutally I don`t think it is needed at all.
This is used to sanity check the kpcr ( [Issue #354](https://github.com/libvmi/libvmi/issues/354) ) There should be no problem with the kpcr in the case of VMI_PM_LEGACY or VMI_PM_PAE (in contrast to VMI_PM_IA32E where a wrong kpcr is read if the currently running process is a wow64 process).

tklengyel: Right. I made this correction in my win10 pr already btw. I'll have to check if there are any other situations where the kernel va is used but I think it's really only used during a sanity check (in which case its a bit of an overkill to have a separate config entry).

libvmi/os/windows/core.c
```diff
@@ -523,6 +528,7 @@ init_from_rekall_profile(vmi_instance_t vmi)
 
     status_t ret = VMI_FAILURE;
     windows_instance_t windows = vmi->os_data;
+    addr_t lowest_kernel_address = 0x0;
     dbprint(VMI_DEBUG_MISC, "**Trying to init from Rekall profile\n");
```
tklengyel: If the user already defined ntoskrln and ntoskrln_va via the config we should skip this part in the Rekall init process that tries to automatically find them. We can skip all the way ahead to the part with `    // Let's do some sanity checking`

libvmi/os/windows/windows.h
tklengyel: So IMHO what would make sense is to reduce this PR to just the parts that introduce the ntoskrnl_va config option and adjusting the init process to skip the parts where the kernel is being searched for.

tklengyel:
This is ok to test.

tklengyel:
This looks good, but could you rebase it on the current master branch and squash the two commits to just one? Thanks!

CMF-Marodeur:
Obviously I don`t know how to git...
I tried to follow this tutorial but failed.
https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request
How the hell can I rebase / squash the commits? :-D

tklengyel:
I usually do git `git rebase -i HEAD~2` then just "fix" the second commit. That squashes it into the previous one seemlessly. You can do HEAD~3 or more if you want to squash more in a single swoop, or just repeat HEAD~2 till you get back to the first commit. Once you are done squashing the commits you can push to your branch with the `--force` option to update the PR.

tklengyel:
Oh, and to rebase it on the current master, you do `git rebase -i origin/master` (provided origin is github.com/libvmi/libvmi). That will pickup your branch and apply it on top of the current master. There shouldn't be any conflicts here so it should be a seamless process.

CMF-Marodeur:
Ok, I think this is completely wrong now... :-\
Should I just create a new commit from scratch? ^^

tklengyel:
You can, just delete your old branch, create a new one with the same name, then force push it =) Sorry this turned out to be so complicated.

namidairo:
Is there any particular reason why the maintainers aren't using the github squash and merge button? Or is it the case that they aren't aware of it because I believe it might be default disabled in the repo settings. Seems like it would be quite a but less complicated than the current workflow.

https://help.github.com/articles/about-pull-request-merge-squashing/

tklengyel:
No, we are aware of it, it's just been the workflow thus far to go with normal merges that also create merge-commits. It also helps with reviewing the code if the PRs are up-to-date with the latest master, even if there are no conflicts.

CMF-Marodeur:
Deleting and recreating the new branch didn`t work either.
So i just created a new PR ( [PR366](https://github.com/libvmi/libvmi/pull/366) ). Seemed to be the best solution.
This was my first time trying to do a PR, so sorry all for the screw-up...

PR #363, opened 2016-08-19
Add example of using Xen's emulate on response feature

tklengyel:


smaresca:
Looks good, thanks Tamas

PR #355, opened 2016-08-13
Add new events from Xen 4.8

tklengyel:
New types of events:
 - Guest-requested event (available on Xen via HVMOP Hypercall since 4.6).
 - Debug exception events
 - CPUID events

PR #364, opened 2016-08-19
Properly clear new events on shutdown

tklengyel:
PR #355 missed clearing up the events on shutdown and wiring up the clear functions to vmi_clear_event.

### DIFF
--- a/libvmi/driver/xen/xen_events.c
+++ b/libvmi/driver/xen/xen_events.c
@@ -1154,6 +1154,9 @@ status_t xen_set_guest_requested_event(vmi_instance_t vmi, bool enabled) {
     if ( xen->major_version != 4 || xen->minor_version < 5 )
         return VMI_FAILURE;
 
+    if ( !enabled && !vmi->guest_requested_event )
+        return VMI_SUCCESS;
+
     rc  = xen->libxcw.xc_monitor_guest_request(xen_get_xchandle(vmi),
                                                xen_get_domainid(vmi),
                                                enabled, 1);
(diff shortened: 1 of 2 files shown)


libvmi/driver/xen/xen_events.c
```diff
@@ -668,6 +669,9 @@ void xen_events_destroy_new(vmi_instance_t vmi)
     rc = xen->libxcw.xc_monitor_write_ctrlreg(xch, dom, VM_EVENT_X86_CR4, false, false, false);
     rc = xen->libxcw.xc_monitor_write_ctrlreg(xch, dom, VM_EVENT_X86_XCR0, false, false, false);
     rc = xen->libxcw.xc_monitor_software_breakpoint(xch, dom, false);
+    rc2 = xen_set_guest_requested_event(vmi, 0);
```
smaresca: @tklengyel any negative implications that you can see here w/ ignoring the rc of any of these calls, like anything that could leave a guest in an odd state?  I understand why the rc variables are just overwritten most of the time, especially given that this function could be considered a best-effort cleanup/error path anyway.

libvmi/driver/xen/xen_events.c
```diff
@@ -632,6 +632,7 @@ status_t process_debug_event(vmi_instance_t vmi,
 void xen_events_destroy_new(vmi_instance_t vmi)
```
smaresca: *_new as just as a distinction from _legacy?

libvmi/driver/xen/xen_events.c
tklengyel: Not really, it's safe to issue these anytime, even if they were not enabled previously.

libvmi/driver/xen/xen_events.c
tklengyel: Correct.

libvmi/driver/xen/xen_events.c
smaresca: ok

PR #372, opened 2016-08-24
Altp2m support

K-Mayer:
This adds support for the direct control of the altp2m-features offered by Xen.

I don`t know if exposing these features to a user without any sanity checks fits in with the spirit of libvmi.
But since I use them constantly I hope they can be usefull for others.

Any comments / suggestions?

Cheers

Maro

### DIFF
--- a/libvmi/Makefile.am
+++ b/libvmi/Makefile.am
@@ -2,6 +2,7 @@ SUBDIRS = config
 
 h_public =  libvmi.h \
             libvmi_extra.h \
+            slat.h \
             x86.h
 
 h_private = \
@@ -28,6 +29,7 @@ c_sources = \
     performance.c \
     pretty_print.c \
     read.c \
+    slat.c \
     strmatch.c \
     write.c \
     memory.c \
@@ -66,7 +68,9 @@ endif
 
 if HAVE_XEN
 h_public    += events.h
-drivers     += driver/xen/xen.h \
+drivers     += driver/xen/altp2m.c \
+               driver/xen/altp2m_private.h \
+               driver/xen/xen.h \
                driver/xen/xen_private.h \
                driver/xen/xen.c \
                driver/xen/xen_events.h \
--- a/libvmi/accessors.c
+++ b/libvmi/accessors.c
@@ -495,16 +495,16 @@ addr_t vmi_translate_uv2p (vmi_instance_t vmi, addr_t virt_address, vmi_pid_t pi
 
 const char *
 vmi_get_linux_sysmap(
-	vmi_instance_t vmi)
+    vmi_instance_t vmi)
 {
     linux_instance_t linux_instance = NULL;
 
     if(VMI_OS_LINUX != vmi->os_type || (VMI_INIT_PARTIAL & vmi->init_mode)){
-	return NULL;
+        return NULL;
     }
 
     if(!vmi->os_data){
-	return NULL;
+        return NULL;
     }
 
     linux_instance = vmi->os_data;
--- a/libvmi/driver/driver_interface.h
+++ b/libvmi/driver/driver_interface.h
@@ -138,6 +138,26 @@ typedef struct driver_interface {
     status_t (*set_debug_event_ptr)(
         vmi_instance_t,
         bool enabled);
+    status_t (*slat_get_domain_state_ptr)(
+        vmi_instance_t vmi,
+        bool *state);
+    status_t (*slat_set_domain_state_ptr)(
+        vmi_instance_t vmi,
+        bool state);
+    status_t (*slat_create_ptr)(
+        vmi_instance_t vmi,
+        uint16_t *view);
+    status_t (*slat_destroy_ptr)(
+        vmi_instance_t vmi,
+        uint16_t view);
+    status_t (*slat_switch_ptr)(
+        vmi_instance_t vmi,
+        uint16_t view);
+    status_t (*slat_change_gfn_ptr)(
+        vmi_instance_t vmi,
+        uint16_t slat_idx,
+        addr_t old_gfn,
+        addr_t new_gfn);
 
     /* Driver-specific data storage. */
     void* driver_data;
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -32,6 +32,7 @@
 #include "driver/xen/xen_events.h"
 #include "driver/driver_interface.h"
 #include "driver/memory_cache.h"
+#include "driver/xen/altp2m_private.h"
 
 //----------------------------------------------------------------------------
 // Helper functions
@@ -927,6 +928,11 @@ xen_init_vmi(
         goto _bail;
 
     ret = xen_init_events(vmi);
+
+    if ( VMI_FAILURE == ret )
+        goto _bail;
+
+    xen_init_altp2m(vmi);
 
 _bail:
     return ret;
--- a/libvmi/driver/xen/xen_events.c
+++ b/libvmi/driver/xen/xen_events.c
@@ -556,7 +556,7 @@ status_t process_cpuid_event(vmi_instance_t vmi,
     vmi->event_callback = 0;
 
     if ( !rsp->flags || rsp->flags == (1 << VM_EVENT_FLAG_VCPU_PAUSED) )
-        dbprint(VMI_DEBUG_XEN, "%s warning: CPUID events require the callback to specify how to handle it, we are likely to be going into a CPUID loop right now\n"
+        dbprint(VMI_DEBUG_XEN, "%s warning: CPUID events require the callback to specify how to handle it, we are likely to be going into a CPUID loop right now\n",
                 __FUNCTION__);
 
     return VMI_SUCCESS;
--- a/libvmi/slat.c
+++ b/libvmi/slat.c
@@ -1,0 +1,54 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+* memory in a target virtual machine or in a file containing a dump of
+* a system's physical memory.  LibVMI is based on the XenAccess Library.
+*
+* Author: Kevin Mayer (kevin.mayer@gdata.de)
+*
+* This file is part of LibVMI.
+*
+* LibVMI is free software: you can redistribute it and/or modify it under
+* the terms of the GNU Lesser General Public License as published by the
+* Free Software Foundation, either version 3 of the License, or (at your
+* option) any later version.
+*
+* LibVMI is distributed in the hope that it will be useful, but WITHOUT
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+* FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+* License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "private.h"
+#include "driver/driver_wrapper.h"
+
+status_t vmi_slat_get_domain_state (vmi_instance_t vmi, bool *state)
+{
+    return driver_slat_get_domain_state (vmi, state);
+}
+
+status_t vmi_slat_set_domain_state (vmi_instance_t vmi, bool state)
+{
+    return driver_slat_set_domain_state (vmi, state);
+}
+
+status_t vmi_slat_create (vmi_instance_t vmi, uint16_t *slat_id)
+{
+    return driver_slat_create (vmi, slat_id);
+}
+
+status_t vmi_slat_destroy (vmi_instance_t vmi, uint16_t slat_idx)
+{
+    return driver_slat_destroy (vmi, slat_idx);
+}
+
+status_t vmi_slat_switch (vmi_instance_t vmi, uint16_t slat_idx)
+{
+    return driver_slat_switch (vmi, slat_idx);
+}
+
+status_t vmi_slat_change_gfn (vmi_instance_t vmi, uint16_t slat_idx, addr_t old_gfn, addr_t new_gfn)
+{
+    return driver_slat_change_gfn (vmi, slat_idx, old_gfn, new_gfn);
+}
--- a/libvmi/slat.h
+++ b/libvmi/slat.h
@@ -1,0 +1,118 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+* memory in a target virtual machine or in a file containing a dump of
+* a system's physical memory.  LibVMI is based on the XenAccess Library.
+*
+* Author: Kevin Mayer (kevin.mayer@gdata.de)
+*
+* This file is part of LibVMI.
+*
+* LibVMI is free software: you can redistribute it and/or modify it under
+* the terms of the GNU Lesser General Public License as published by the
+* Free Software Foundation, either version 3 of the License, or (at your
+* option) any later version.
+*
+* LibVMI is distributed in the hope that it will be useful, but WITHOUT
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+* FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+* License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/**
+* @file slat.h
+* @brief The LibVMI second level address translation API is defined here.
+*
+* The slat (Second Level Address Translation) is used to circumvent the
+* overhead of shadow page tables. By modifying the pointer to the
+* translation tables they can also be used to switch between completely
+* different sets of access rights for memory pages.
+*/
+
+#ifndef LIBVMI_SLAT_H
+#define LIBVMI_SLAT_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#pragma GCC visibility push(default)
+
+/**
+* Checks if slat is enabled on the domain
+*
+* @param[in] vmi LibVMI instance
+* @param[out] state slat state of the domain
+* @return VMI_SUCCESS or VMI_FAILURE
+*/
+status_t vmi_slat_get_domain_state (
+    vmi_instance_t vmi,
+    bool *state);
+
+/**
+* Enables or disables slat for the domain
+*
+* @param[in] vmi LibVMI instance
+* @param[in] state slat state of the domain
+* @return VMI_SUCCESS or VMI_FAILURE
+*/
+status_t vmi_slat_set_domain_state (
+    vmi_instance_t vmi,
+    bool state);
+
+/**
+* Creates a new slat slat_id
+*
+* @param[in] vmi LibVMI instance
+* @param[out] slat_id Number of the newly created slat_id
+* @return VMI_SUCCESS or VMI_FAILURE
+*/
+status_t vmi_slat_create (
+    vmi_instance_t vmi,
+    uint16_t *slat_id);
+
+/**
+* Destroys an slat slat_id
+*
+* @param[in] vmi LibVMI instance
+* @param[in] slat_id Number of the slat_id which is to be destroyed
+* @return VMI_SUCCESS or VMI_FAILURE
+*/
+status_t vmi_slat_destroy (
+    vmi_instance_t vmi,
+    uint16_t slat_idx);
+
+/**
+* Switches to a specific slat slat_id
+*
+* @param[in] vmi LibVMI instance
+* @param[in] slat_id Number of the slat_id which to which to switch to
+* @return VMI_SUCCESS or VMI_FAILURE
+*/
+status_t vmi_slat_switch (
+    vmi_instance_t vmi,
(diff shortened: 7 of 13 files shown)


CMF-Marodeur:
Just saw an error when declaring / implementing the altp2m-functions.
Were should the different functions be declared / implemented?
Maybe like this:
vmi_altp2m_get_domain_state -> accessors + libvmi.h
vmi_altp2m_set_domain_state -> accessors + libvmi.h
vmi_altp2m_create_p2m -> ?
vmi_altp2m_destroy_p2m -> ?
vmi_altp2m_switch_p2m -> ?
vmi_altp2m_set_mem_access -> events
vmi_altp2m_change_gfn -> accessors + libvmi.h

tklengyel:
So I had been debating about adding this wrapper to LibVMI as well. The thing to consider here is that there is no support for altp2m-type setup in KVM or any other hypervisor but Xen. As such, adding this wrapper which simply forwards to libxc 1:1 adds little value. So I'm torn whether we should have this added to the LibVMI API or not at this time.

@smaresca @awsaba any opinion?

smaresca:
Even if no other hypervisor implements a feature, I think it's likely fine to expose it so other future drivers can make use of it.

tklengyel:
@smaresca Sure. In that case I would prefer a hypervisor-neutral naming for it though. We already use VMM Pagetable ID instead of altp2m_idx for this reason in the events API. Maybe vmi_vmmpt_*?

smaresca:
I feel like the expression of such a capability in the API will need a more generic name though

tklengyel:
This is ok to test.

smaresca:
@CMF-Marodeur could you describe your current use-case? We're definitely familiar with alt-p2m and its purpose, but knowing how you make use of it could help shape the API.

awsaba:
I'd say include it, but just barely, because the EPTP is architectural feature that may or may not be supported by the hardware itself.  If included, it should get it's own header,  vmi_intel (or vmi_x86, does SVM have any equivalent?) because being generic is nice, but having libvmi keep track of an xch and domid is definitely helpful.

tklengyel:
@awsaba: it's not necessarily architectural. We are actually adding altp2m support to Xen on ARM right now. But the name altp2m is definitely Xen-specific.

tklengyel:
@CMF-Marodeur: to answer your question, I think this could go into a separate header that get's installed with the other libvmi headers. It doesn't really belong to the core libvmi API or the events API. The one altp2m function (altp2m_set_mem_access) that belongs to the events API already is already exposed via the standard LibVMI events API.

CMF-Marodeur:
@smaresca I use the altp2m-feature to quickly switch between two sets of access rights. Basically I divide the guest memory into OS-specific memory and third-party memory because I`m interestested in the transition between those memory regions.
Something a bit like CXPInspector by Carsten Willems et al..

In my use case I also discarded the libvmi event handling (to much overhead since I have dozens / hundreds of memory pages which should all call the same callback) and instead I just defined a default handler for write and one for execute events. For this I implemented another set_mem_access which does not rely on the libvmi events:
status_t vmi_altp2m_set_mem_access ( vmi_instance_t vmi , uint16_t view, vmi_mem_access_t access, addr_t gfn )
Problem ist that this approach is a pretty dirty hack (in the event handling) and would require quite some work for a clean and configurable implementation.

Name:
I assume something containing EPT is discarded because it is not vendor agnostic?

tklengyel:
@CMF-Marodeur: indeed, the current event API has some overhead in that it requires separate event structures for each page it is registered on. We could probably make it more flexible to have a subset of the mem_access event type that registers to a type of violation rather then specific gfns.

CMF-Marodeur:
@tklengyel : That is exactly what I needed (in combination with a way to set the access rights without having to create an event struct first).

Concerning this PR:
So I basically create a new header which exposes the functions. The implementation stays in xen_events.c.
Since you guys already use VMM Pagetable ID i could stick with vmi_vmmpt_*. For example

vmi_vmmpt_create
vmi_vmmpt_switch
vmi_vmmpt_switch

As an alternative I would suggest vmi_slat_*.
[slat](https://en.wikipedia.org/wiki/Second_Level_Address_Translation)

tklengyel:
Sure, let's go with slat, that is at least standard naming.

tklengyel:
And as for naming within the xen-driver, it should stick to altp2m, only in the public-header and driver interface should it follow the hypervisor neutral-naming.

smaresca:
+1

CMF-Marodeur:
So I moved the slat-code to a dedicated header / source file and changed the names.
Additionally I extendet the makefile etc. so programs using libvmi can access the new functions.

And right after the commit I saw two minor typos / wrong text messages. ^^

libvmi/Makefile.am
```diff
@@ -2,6 +2,7 @@ SUBDIRS = config
 
 h_public =  libvmi.h \
             libvmi_extra.h \
+			slat.h \
```
tklengyel: alignment doesn't match, are you using tabs instead of spaces?

libvmi/Makefile.am
```diff
@@ -28,6 +29,7 @@ c_sources = \
     performance.c \
     pretty_print.c \
     read.c \
+	slat.c \
```
tklengyel: alignment

libvmi/accessors.c
```diff
@@ -511,4 +511,4 @@ vmi_get_linux_sysmap(
 
     return linux_instance->sysmap;
 
-}
+}
```
tklengyel: Unnecessary change.

libvmi/driver/driver_interface.h
```diff
@@ -138,6 +138,26 @@ typedef struct driver_interface {
     status_t (*set_debug_event_ptr)(
         vmi_instance_t,
         bool enabled);
+    status_t ( *slat_get_domain_state_ptr )(
```
tklengyel: Style doesn't match the others here (ie. no spaces before and after the function name)

libvmi/events.c
```diff
@@ -754,4 +754,4 @@ status_t vmi_shutdown_single_step(vmi_instance_t vmi)
     {
         return VMI_FAILURE;
     }
-}
+}
```
tklengyel: Unnecessary change.

libvmi/slat.c
```diff
@@ -0,0 +1,70 @@
...
+* You should have received a copy of the GNU Lesser General Public License
+* along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "private.h"
+#include "driver/driver_wrapper.h"
+
+status_t vmi_slat_get_domain_state ( vmi_instance_t vmi , bool *state )
+{
+    status_t rc = VMI_FAILURE;
```
tklengyel: No need for defining status_t here, just return the call to the driver directly.

libvmi/slat.h
```diff
@@ -0,0 +1,119 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+* memory in a target virtual machine or in a file containing a dump of
+* a system's physical memory.  LibVMI is based on the XenAccess Library.
+*
+* Copyright 2011 Sandia Corporation. Under the terms of Contract
+* DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government
+* retains certain rights in this software.
+*
+* Author:
```
tklengyel: Author missing.

libvmi/slat.h
```diff
@@ -0,0 +1,119 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+* memory in a target virtual machine or in a file containing a dump of
+* a system's physical memory.  LibVMI is based on the XenAccess Library.
+*
+* Copyright 2011 Sandia Corporation. Under the terms of Contract
```
tklengyel: I don't think you are under contract with Sandia, so remove this block from here.

libvmi/slat.c
```diff
@@ -0,0 +1,70 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+* memory in a target virtual machine or in a file containing a dump of
+* a system's physical memory.  LibVMI is based on the XenAccess Library.
+*
+* Copyright 2011 Sandia Corporation. Under the terms of Contract
```
tklengyel: Not under Sandia contract, no need to add this.

libvmi/slat.c
```diff
@@ -0,0 +1,70 @@
...
+* DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government
+* retains certain rights in this software.
+*
+* Author:
```
tklengyel: Author missing.

libvmi/slat.h
```diff
@@ -0,0 +1,119 @@
...
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/**
+* @file slat.h
+* @brief The LibVMI second level address translation API is defined here.
+*
+* More detailed description can go here.
```
tklengyel: More detailed description can go here. ;)

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
     return vrc;
 }
+
+status_t xen_altp2m_get_domain_state ( vmi_instance_t vmi , bool *state )
+{
+    int rc;
+    xen_instance_t *xen = xen_get_instance ( vmi );
+    xc_interface * xch = xen_get_xchandle ( vmi );
+    unsigned long domain_id = xen_get_domainid ( vmi );
+    printf ( "in xen_altp2m_get_domain_state with domain_id: %lx\n" , domain_id );
```
tklengyel: Debug message should be printed with dbprint().

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
 
     return vrc;
 }
+
+status_t xen_altp2m_get_domain_state ( vmi_instance_t vmi , bool *state )
+{
+    int rc;
+    xen_instance_t *xen = xen_get_instance ( vmi );
+    xc_interface * xch = xen_get_xchandle ( vmi );
+    unsigned long domain_id = xen_get_domainid ( vmi );
```
tklengyel: This should be domid_t

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+    printf ( "in xen_altp2m_get_domain_state with domain_id: %lx\n" , domain_id );
+    if ( !xch ) {
+        errprint ( "%s error: invalid xc_interface handle\n" , __FUNCTION__ );
+        return VMI_FAILURE;
+    }
+    if ( domain_id == VMI_INVALID_DOMID ) {
```
tklengyel: needs casting (ie. (domid)VMI_INVALID_DOMID ).

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+    return VMI_SUCCESS;
+
+}
+
+status_t xen_altp2m_set_domain_state ( vmi_instance_t vmi , bool state )
+{
+    int rc;
+    xen_instance_t *xen = xen_get_instance ( vmi );
+    xc_interface * xch = xen_get_xchandle ( vmi );
+    unsigned long domain_id = xen_get_domainid ( vmi );
```
tklengyel: This should be domid_t

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+    if ( !xch ) {
+        errprint ( "%s error: invalid xc_interface handle\n" , __FUNCTION__ );
+        return VMI_FAILURE;
+    }
+    if ( domain_id == VMI_INVALID_DOMID ) {
```
tklengyel: needs casting (ie. (domid_t)VMI_INVALID_DOMID ).

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+{
+    int rc;
+    xen_instance_t *xen = xen_get_instance ( vmi );
+    xc_interface * xch = xen_get_xchandle ( vmi );
+    domid_t domain_id = xen_get_domainid ( vmi );
+    if ( !xch ) {
+        errprint ( "%s error: invalid xc_interface handle\n" , __FUNCTION__ );
+        return VMI_FAILURE;
+    }
+    if ( domain_id == VMI_INVALID_DOMID ) {
```
tklengyel: needs casting (ie. (domid_t)VMI_INVALID_DOMID ).

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+    return VMI_SUCCESS;
+
+}
+
+status_t xen_altp2m_create_p2m ( vmi_instance_t vmi , uint16_t *altp2m_idx )
+{
+    int rc;
+    xen_instance_t *xen = xen_get_instance ( vmi );
+    xc_interface * xch = xen_get_xchandle ( vmi );
+    domid_t domain_id = xen_get_domainid ( vmi );
```
tklengyel: This should be domid_t

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+    return VMI_SUCCESS;
+
+}
+
+status_t xen_altp2m_destroy_p2m ( vmi_instance_t vmi , uint16_t altp2m_idx )
+{
+    int rc;
+    xen_instance_t *xen = xen_get_instance ( vmi );
+    xc_interface * xch = xen_get_xchandle ( vmi );
+    domid_t domain_id = xen_get_domainid ( vmi );
```
tklengyel: This should be domid_t

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+    if ( !xch ) {
+        errprint ( "%s error: invalid xc_interface handle\n" , __FUNCTION__ );
+        return VMI_FAILURE;
+    }
+    if ( domain_id == VMI_INVALID_DOMID ) {
```
tklengyel: casting

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
...
+    int rc;
+    xen_instance_t *xen = xen_get_instance ( vmi );
+    xc_interface * xch = xen_get_xchandle ( vmi );
+    domid_t domain_id = xen_get_domainid ( vmi );
+
+    if ( !xch ) {
+        errprint ( "%s error: invalid xc_interface handle\n" , __FUNCTION__ );
+        return VMI_FAILURE;
+    }
+    if ( domain_id == VMI_INVALID_DOMID ) {
```
tklengyel: casting missing

libvmi/driver/xen/xen_events.c
```diff
@@ -1410,3 +1417,147 @@ status_t xen_events_listen(vmi_instance_t vmi, uint32_t timeout)
 
     return vrc;
 }
+
+status_t xen_altp2m_get_domain_state ( vmi_instance_t vmi , bool *state )
```
tklengyel: These functions should be in a separate file in the Xen folder, say altp2m.c, as these don't belong to the events system.

libvmi/driver/xen/xen_events.c
```diff
@@ -728,6 +728,13 @@ status_t xen_init_events_new(vmi_instance_t vmi)
     vmi->driver.set_cpuid_event_ptr = &xen_set_cpuid_event;
     vmi->driver.set_debug_event_ptr = &xen_set_debug_event;
 
+    vmi->driver.slat_get_domain_state_ptr = &xen_altp2m_get_domain_state;
```
tklengyel: Wiring up these functions should be done in xen.c and only if the Xen version is 4.6 or newer.

libvmi/driver/xen/xen_events_private.h
```diff
@@ -144,6 +144,15 @@ status_t xen_set_cpuid_event(vmi_instance_t vmi, bool enabled);
 status_t xen_start_single_step(vmi_instance_t vmi, single_step_event_t *event);
 status_t xen_stop_single_step(vmi_instance_t vmi, uint32_t vcpu);
 status_t xen_shutdown_single_step(vmi_instance_t vmi);
+status_t xen_altp2m_get_domain_state (vmi_instance_t vmi, bool *state);
```
tklengyel: Should be moved to another header, altp2m_private.h which gets included in xen.c

libvmi/Makefile.am
CMF-Marodeur: Actually yes (unintentionally).
My IDE seems to be forgetting that it should use 4 spaces instead of tabs.

libvmi/slat.c
CMF-Marodeur: How much information about the author do you need?
Is a name (real name) sufficent?

libvmi/slat.h
CMF-Marodeur: I saw this in the libvmi.h (and other headers), so I thought it would be ok for now. ^^
I will try to think of a more detailed description.

libvmi/driver/xen/xen_events_private.h
CMF-Marodeur: Does the new header require the copyright/licences like the xen_events_private.h
[...]

> * Portions of this header and dependent code is based upon that in xen-access,
> *    from the official Xen source distribution.  That code carries the
> *    following copyright notices and license.
> *
> * Copyright (c) 2011 Virtuata, Inc.
> * Copyright (c) 2009 by Citrix Systems, Inc. (Patrick Colp), based on
> *   xenpaging.c
[...]

CMF-Marodeur:
Baby steps, but I think we are getting there.

Author is still intentionally left blank.
Do you guys need a real name for that?

Concerning the xen_init_altp2m in xen.c:
Technically it can not fail so the return status_t is of no importance there. Not sure if it should stay there or not.

libvmi/accessors.c
```diff
@@ -495,16 +495,16 @@ addr_t vmi_translate_uv2p (vmi_instance_t vmi, addr_t virt_address, vmi_pid_t pi
 
 const char *
 vmi_get_linux_sysmap(
-	vmi_instance_t vmi)
+    vmi_instance_t vmi)
 {
```
CMF-Marodeur: unrelated fixes (tabs were used there). Should they stay in?

libvmi/driver/xen/xen.c
```diff
@@ -835,6 +836,26 @@ xen_init_events(
 }
 
 status_t
+xen_init_altp2m (
```
tklengyel: So since there is no separate flag the user sends to specifically activate the altp2m parts this function doesn't need to return a status_t and can just be void. You could also make it into a static inline and move it to altp2m_private.h

libvmi/accessors.c
tklengyel: Fine by me.

libvmi/Makefile.am
```diff
@@ -66,7 +68,8 @@ endif
 
 if HAVE_XEN
 h_public    += events.h
-drivers     += driver/xen/xen.h \
+drivers     += driver/xen/altp2m.c \
```
tklengyel: Also add altp2m_private.h here.

tklengyel:
This is shaping up fine so with the last two things fixed +1 from me.

libvmi/driver/xen/xen.c
CMF-Marodeur: I followed the given scheme, because I think that maybe there should be a flag there.
But switching it to void is fine with me. Any user will see the dbprint from driver_wrapper.h if he tries to use the functions with the wrong xen-version.

CMF-Marodeur:
What about the Author?
Is CMF-Marodeur (no email) fine with you guys?

K-Mayer:
Any comments?

tklengyel:
We will need another +1 from a maintainer and then we can merge. PRs that add/change the API need two maintainers to ack it.

awsaba:
Needs a squash, but otherwise LGTM.

@tklengyel should #382 be merged first and this PR rebased to prevent new warnings from popping up?

tklengyel:
@awsaba: either way works for me

smaresca:
@K-Mayer thanks for the squash -- merging now

PR #373, opened 2016-08-24
Add generic mem_access event types

tklengyel:
The current mem_access event API requires the user to create a vmi_event_t for each page they want to set an event on. This can incur heavy memory overhead when the same type of access restriction is placed on multiple pages.

In this PR we introduce a generic mem event type which allows the user to subscribe to  vmi_mem_access_t events regardless of which page the access violation occurred. We further introduce vmi_set_mem_event which allows the user to change specific page permissions once a handler has been registered.

Notice: we currently don't allow to mix generic mem events with gfn-specific mem events. Doing so may be possible in the future if there is need for it, however for now we choose to reduce the complexity of the code this way.

### DIFF
--- a/libvmi/driver/xen/xen_events_legacy.c
+++ b/libvmi/driver/xen/xen_events_legacy.c
@@ -417,24 +417,47 @@ status_t process_mem(vmi_instance_t vmi, bool access_r, bool access_w, bool acce
         return VMI_FAILURE;
     }
 
-    vmi_event_t *event = g_hash_table_lookup(vmi->mem_events, &gfn);
+    vmi_event_t *event;
     vmi_mem_access_t out_access = VMI_MEMACCESS_INVALID;
     if(access_r) out_access |= VMI_MEMACCESS_R;
     if(access_w) out_access |= VMI_MEMACCESS_W;
     if(access_x) out_access |= VMI_MEMACCESS_X;
 
-    if (event && event->mem_event.in_access & out_access) {
-        event->mem_event.gla = gla;
-        event->mem_event.gfn = gfn;
-        event->mem_event.offset = offset;
-        event->mem_event.out_access = out_access;
-        event->vcpu_id = vcpu_id;
+    if ( g_hash_table_size(vmi->mem_events_on_gfn) ) {
+        event = g_hash_table_lookup(vmi->mem_events_on_gfn, &gfn);
 
-        vmi->event_callback = 1;
-        process_response ( event->callback(vmi, event), rsp_flags );
-        vmi->event_callback = 0;
+        if (event && event->mem_event.in_access & out_access) {
+            event->mem_event.gla = gla;
+            event->mem_event.gfn = gfn;
+            event->mem_event.offset = offset;
+            event->mem_event.out_access = out_access;
+            event->vcpu_id = vcpu_id;
+            process_response ( event->callback(vmi, event), rsp_flags );
+            return VMI_SUCCESS;
+        }
+    }
 
-        return VMI_SUCCESS;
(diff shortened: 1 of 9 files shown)


tklengyel:
Tested on Xen 4.7 and works as expected.

libvmi/driver/xen/xen_events_legacy.c
```diff
@@ -925,7 +947,7 @@ xen_set_mem_access_legacy(vmi_instance_t vmi, mem_access_event_t *event,
 
         hvmmem_access_t access = memaccess_conversion[page_access_flag];
 
-        rc = xen->libxcw.xc_hvm_set_mem_access(xch, dom, access, page_key, npages);
+        rc = xen->libxcw.xc_hvm_set_mem_access(xch, dom, access, gpfn, 1);
```
CMF-Marodeur: Small suggestion:
Maybe you could use a define here (and in similar cases).
It may be a matter of personal taste, but I find
`rc = xen->libxcw.xc_hvm_set_mem_access(xch, dom, access, gpfn, SINGLE_PAGE);`
much easier to read, especially when the definition of the function only states " uint64_t nr" as the type / name (which is not very helpfull for people who are new to xen / libvmi).

tklengyel: Gotcha. It doesn't really worth a define but I'll add a comment to say that effectively ;)

K-Mayer:
@tklengyel 
I don't know where to put this, since I there is no general conversation possible and this also isn't a real issue, so I'll append it here:
When converting my code to the newest libvmi-version I noticed the 
`vmi_set_mem_event(vmi_instance_t` vmi, addr_t physical_address[....])
I think libvmi deprecated the byte-wise granularity with PR  #339 so the last page offset of the address is always ignored.
Wouldn't it make sense to switch to a gfn-based addressing?
At least in my usecase I work with gfn most of the time (since a finer granularity is not supported by hardware) resulting in my code shifting the gfn << 12 (to convert it to the address) just to have libvmi shift it back again when setting the memory accces.

There are good reasons for both approaches.
1. For people not familiar with the hardware constrains the gfn concept may seem a bit confusing. => hide the constrains and allow the user to supply a complete address. 
2. The hardware does not support byte granularity => use the gfn addressing
As fare as I can see in the code the user has to check if the event really occoured at the desired address anyway, since libvmi only returns "an event on page X happened with address Y". Is this correct? If so this means that the constrains are not really hidden from the user, invalidating point 1.

tklengyel:
I was thinking about this as well but as you say it is simpler for users not having to know and keep in mind that they need to shift their PAs to get gfn's. Especially as we don't currently define any gfn_t or macros/helpers to do this conversion. So for a beginner this would fall into the "dark magic" category.

As for two, we do report the full address in "gla". We additionally also report the page and the offset, but not exclusively.

K-Mayer:
Thats what I meant with "an event on page X happened with address Y".
The user gets any event on the page and has to decide if it is the event they wanted based on the gla / offset. This does not hide the page granularity imposed by the hardware, but rather obfuscates it. The old byte events tried to hide it from the user.
In my opinion the current approach is the same black magic for beginners (at least it was for me ^^).

> I set an event at address 0x1234 but got an event at 0x1337. Wtf? oO

While I like the gfn approach better because once understood it provides a reason for the strange event addresses (and some other confusing behaviour), this pretty much seems to be a matter of taste.

Providing vmi_set_mem_event_on_address as well as vmi_set_mem_event_on_gfn would just increase the confusion I suppose.

tklengyel:
Yea, at some point there is no way around having the user understand what he/she is doing =)

PR #376, opened 2016-08-26
LibVMI events ABI stabilization

### LINKED ISSUE #375: LibVMI events ABI compatibility
The current events ABI is not backward or forward compatible, meaning a binary compiled with one version of the events structures has no way of safely interacting with an older or newer version of LibVMI. Given that LibVMI is a shared library it is quite possible that a binary will be compiled with one version of LibVMI but will get deployed on a system that has another version installed. Similar issues were addressed with Xen when the vm_events ABI has been rearchitected.

To address this issue the following steps need to be taken:
 - Avoid the use of enums in event structures as a compiler is free to interpret its underlying container, which means a binary and the library may have different opinion about the structures memory layout
 - Introduce versioning to the events structures so that LibVMI will be able to deal with binaries compiled with older versions
 - Make event structures explicitly packed

tklengyel:
Fixes #375

### DIFF
--- a/examples/event-example.c
+++ b/examples/event-example.c
@@ -294,6 +294,7 @@ int main (int argc, char **argv)
      *  us to follow as various tasks are scheduled and run upon the CPU)
      */
     memset(&cr3_event, 0, sizeof(vmi_event_t));
+    cr3_event.version = VMI_EVENTS_VERSION;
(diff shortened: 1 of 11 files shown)


smaresca:
+1 from me
Tony could you give a +1 also?

PR #382, opened 2016-08-29
Compile with -Werror -Wall -Wextra and fix associated warnings

tklengyel:
The changes are mechanic fixing of warnings issued mostly by -Wunused-parameter and -Wunused-variable

### DIFF
--- a/libvmi/driver/kvm/kvm.c
+++ b/libvmi/driver/kvm/kvm.c
@@ -1115,11 +1115,12 @@ kvm_get_memory_patch(
     }
     else {
         // get the data from kvm
-        nbytes =
-            read(kvm_get_instance(vmi)->socket_fd, buf, length + 1);
-        if (nbytes != (length + 1)) {
+        nbytes = read(kvm_get_instance(vmi)->socket_fd, buf, length + 1);
+        if ( nbytes <= 0 )
             goto error_exit;
-        }
+
+        if ( (uint32_t)nbytes != (length + 1) )
+            goto error_exit;
 
         // check that kvm thinks everything is ok by looking at the last byte
         // of the buffer, 0 is failure and 1 is success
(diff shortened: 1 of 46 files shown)


libvmi/driver/kvm/kvm.c
```diff
@@ -1115,11 +1115,12 @@ kvm_get_memory_patch(
     }
     else {
         // get the data from kvm
-        nbytes =
-            read(kvm_get_instance(vmi)->socket_fd, buf, length + 1);
-        if (nbytes != (length + 1)) {
+        nbytes = read(kvm_get_instance(vmi)->socket_fd, buf, length + 1);
+        if ( nbytes <= 0 )
```
smaresca: What if nbytes = length/2 ? Was the original condition intended to check that data of proper size was returned, or just for an invalid rc from read()?

smaresca: nevermind, I see the condition below now

libvmi/performance.c
```diff
@@ -53,52 +53,6 @@ print_measurement(
            (long int) ktv_end.tv_usec, *diff);
 }
 
-static double
```
smaresca: No harm in keeping these around, no?

libvmi/driver/kvm/kvm.c
tklengyel: Yea, the problem was comparing signed and unsigned variables but just casting it to unsigned is not safe without first checking if it's less then 0.

smaresca:
I like tagging deliberately unused parameters in the function prototype. 

GCC/Clang 
http://stackoverflow.com/a/12891181 

MSCV
#define UNUSED(x) __pragma(warning(suppress: 4100 4101)) x

libvmi/performance.c
tklengyel: They break compiling with -Werror so if we want to keep them we can just comment them out.

tklengyel:
Retest this please.

tklengyel:
@smaresca: sure, looks fine to me that way too.

tklengyel:
Retest this please.

PR #384, opened 2016-09-08
Refactor cache interface

tklengyel:
Rather then defining empty functions in the c files when the cache is disabled, do it right in the header and disable compiling cache.c when not needed. Also, relocate SHM cache functions to shm.c.


smaresca:
+1, merging

PR #385, opened 2016-09-14
Support out of tree builds for examples

mneilsen:
When building examples out of tree, include files relative to the top
source directory rather than the top build directory.


tklengyel:
This is ok to test.

PR #400, opened 2016-10-07
Issue event callbacks on shutdown as well

tklengyel:
When LibVMI is shutting down, the event queue is simply being looped and all events issued a simple response to allow the vCPU to continue. However, without processing events appropriately this may leave the VM in an unstable condition. For example, failure to reinject a breakpoint already queued on the ring is a serious problem. To be consistent with the handling we will process all events the same way and issue callbacks to their handler.

### DIFF
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -687,8 +687,8 @@ vmi_destroy(
         return VMI_FAILURE;
 
     vmi->shutting_down = TRUE;
-    events_destroy(vmi);
     driver_destroy(vmi);
+    events_destroy(vmi);
     if (vmi->os_interface) {
         os_destroy(vmi);
     }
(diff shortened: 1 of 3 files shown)


tklengyel:
Retest this please.

libvmi/core.c
```diff
@@ -687,8 +687,8 @@ vmi_destroy(
         return VMI_FAILURE;
 
     vmi->shutting_down = TRUE;
-    events_destroy(vmi);
     driver_destroy(vmi);
+    events_destroy(vmi);
```
smaresca: [echoing sidebar conversation] i'm curious about the swapped order here

tklengyel: Necessary so that the callback tables are not destroyed before being able to issue the final callbacks.

PR #406, opened 2016-11-09
Fix argument count check so dump-memory runs correctly

mattshockl:
Fix argc check to verify 3 arguments instead of 2 so we can read from argv[2] later on in the program


tklengyel:
This is ok to test.

PR #410, opened 2016-11-18
Correct argument count check in map-symbol example

mattshockl:
argc must be at least 3 for argv[2] to not be nil


smaresca:
+1, will merge

PR #412, opened 2016-12-02
Fix memory leak when trying to init events on PV domains

tklengyel:
Destroying events is only called if xen->hvm is set, so we should mirror that for the initialization.

### DIFF
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -901,7 +901,7 @@ xen_init_vmi(
     if ( VMI_FAILURE == ret )
         goto _bail;
 
-    if(vmi->init_mode & VMI_INIT_EVENTS)
+    if(xen->hvm && (vmi->init_mode & VMI_INIT_EVENTS))
     {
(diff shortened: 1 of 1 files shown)


libvmi/driver/xen/xen.c
```diff
@@ -901,7 +901,7 @@ xen_init_vmi(
     if ( VMI_FAILURE == ret )
         goto _bail;
 
-    if(vmi->init_mode & VMI_INIT_EVENTS)
+    if(xen->hvm && (vmi->init_mode & VMI_INIT_EVENTS))
```
bdpayne: This coding style is kind of error prone.  Could you put brackets around the outer if statement?

smaresca: +1

tklengyel: Not sure which outer if you are referring to.

bdpayne: Oh, sorry, I read the diff wrong ;-)  This actually looks just fine!

PR #418, opened 2016-12-05
Add missing event_callback flip before issuing callbacks

tklengyel:


bdpayne:
Could you add a brief comment explaining why this is needed?

tklengyel:
It's required to be able to work with multi-vCPU events. When you are in a callback with multi-vCPU VMs and you clear an event in a callback you are at risk of crashing because there might already be another event queued for that same event on the ring. If you just blindly clear that event in the callback the next event will be pulled from the ring without a handler and we crash because we don't know what to do. This flip signals to LibVMI that vmi_clear_events needs to be queued because we are in an active callback. Once all events are pulled from the ring the VM gets paused and events get cleared without risking a race-condition.

bdpayne:
Sorry, I meant a comment in the code :-)

tklengyel:
There is already a comment in the code to this effect (see https://github.com/libvmi/libvmi/blob/master/libvmi/events.c#L789).

bdpayne:
Ok, but that's in a different file.  No need for the long explanation again.  Just a quick reference.  I make the suggestion b/c I think code like this isn't always obvious 6 months from now.

tklengyel:
IMHO there are too many places we issue callbacks to add such a comment at in the driver level.

bdpayne:
Fair enough :-)

PR #419, opened 2016-12-05
Rework parts of the events ABI

### LINKED ISSUE #417: LibVMI doesn't clear memory events properly in callback
This may just be my lack of knowledge with LibVMI internals, but I can't seem to find a way to clear a memory event without entering into an infinite loop.  I modified the msr_event_example to simply create an event that signals when the page that the MSR_LSTAR register points to executes, print something, then clear the event.  The expected output of this program should be one print statement, then continued execution of the kernel since, per the API, clear_event should reset the page permissions and continue on the current instruction.  However, even though I clear the event, once the kernel begins executing again, it signals the event again.  This just creates a loop that freezes the guest completely and it seems there is no way to break out of it.  Even running the example code given in the examples directory can't print out the system calls without freezing the guest and entering into an infinite loop.  I'm using a Windows 7-64 bit guest running on Xen if that matters.  The poc code is below:

```
#include <stdlib.h>
#include <string.h>
#include <errno.h>
#include <sys/mman.h>
#include <stdio.h>
#include <inttypes.h>
#include <signal.h>

#include <libvmi/libvmi.h>
#include <libvmi/events.h>

vmi_event_t syscall_event;

event_response_t syscall_event_callback(vmi_instance_t vmi, vmi_event_t *event) {
    fprintf(stderr, "%s\n", "Syscall Reached");
    vmi_clear_event(vmi, event, NULL);
    return 0;
}

static int interrupted = 0;
static void close_handler(int sig){
    interrupted = sig;
}

int main (int argc, char **argv) {
    status_t status = VMI_SUCCESS;
    vmi_instance_t vmi;
    struct sigaction act;
    act.sa_handler = close_handler;
    act.sa_flags = 0;
    sigemptyset(&act.sa_mask);
    sigaction(SIGHUP,  &act, NULL);
    sigaction(SIGTERM, &act, NULL);
    sigaction(SIGINT,  &act, NULL);
    sigaction(SIGALRM, &act, NULL);

    char *name = NULL;

    if(argc < 2){
        fprintf(stderr, "Usage: syscall_events_example <name of VM>\n");
        exit(1);
    }

    // Arg 1 is the VM name.
    name = argv[1];

    // Initialize the libvmi library.
    if (vmi_init(&vmi, VMI_XEN | VMI_INIT_COMPLETE | VMI_INIT_EVENTS, name) == VMI_FAILURE){
        printf("Failed to init LibVMI library.\n");
        if (vmi != NULL ) {
            vmi_destroy(vmi);
        }
        return 1;
    }
    else{
        printf("LibVMI init succeeded!\n");
    }

    vmi_pause_vm(vmi);

    reg_t lstar = 0;
    addr_t phys_lstar = 0;

    vmi_get_vcpureg(vmi, &lstar, MSR_LSTAR, 0);
    phys_lstar = vmi_translate_kv2p(vmi, lstar);

    /* Register event to track any writes to a MSR. */
    memset(&syscall_event, 0, sizeof(vmi_event_t));
    SETUP_MEM_EVENT(&syscall_event, phys_lstar, VMI_MEMACCESS_X, syscall_event_callback, 0);
    syscall_event.mem_event.generic = 0;

    vmi_register_event(vmi, &syscall_event);

    vmi_resume_vm(vmi);

    printf("Waiting for events...\n");

    while(!interrupted){
        status = vmi_events_listen(vmi,500);
        if (status != VMI_SUCCESS) {
            printf("Error waiting for events, quitting...\n");
            interrupted = -1;
        }
    }

    printf("Finished with test.\n");

    // cleanup any memory associated with the libvmi instanc
    vmi_destroy(vmi);

    return 0;
}
```

### LINKED ISSUE #420: LibVMI overwriting VCPU data
It seems as LibVMI is overwriting step data in between calls to register_singlestep_event and clear_singlestep_event, causing errors when CHECK_VCPU_SINGLESTEP can't find the correct VCPU number.  I instrumented events.c to print the event->ss_event.vcpus value in register and clear and got the following output:

```
Syscall Reached: LSTAR=0xfffff800026d2640 RIP=0xfffff800026d28d3
register_singlestep_event vcpus: 1
clear_singlestep_event: 40708315
Syscall Reached: LSTAR=0xfffff800026d2640 RIP=0xfffff800026d28db
register_singlestep_event vcpus: 1
clear_singlestep_event: 40708322
Syscall Reached: LSTAR=0xfffff800026d2640 RIP=0xfffff800026d28e2
```

You can see that the value is getting overwritten (or something?) in between those calls.  The code I used to run this is:

```
event_response_t reset_event_callback(vmi_instance_t vmi, vmi_event_t *event) {
    event->version = VMI_EVENTS_VERSION;
    vmi_register_event(vmi, event);

    return VMI_EVENT_RESPONSE_NONE;
}

event_response_t syscall_event_callback(vmi_instance_t vmi, vmi_event_t *event) {
    fprintf(stderr, "Syscall Reached: LSTAR=0x%lx RIP=0x%lx\n", lstar, event->x86_regs->rip);

    event->mem_event.generic = 0;
    event->mem_event.physical_address = event->mem_event.gfn << 12;
    vmi_clear_event(vmi, event, NULL);
    
    vmi_step_event(vmi, event, event->vcpu_id, 1, reset_event_callback);

    return VMI_EVENT_RESPONSE_NONE;
}
```

And the code I added to instrument is:
```
status_t register_singlestep_event(vmi_instance_t vmi, vmi_event_t *event) {
    ...
    fprintf(stderr, "register_singlestep_event vcpus: %u\n", event->ss_event.vcpus);
    ...
}

status_t clear_singlestep_event(vmi_instance_t vmi, vmi_event_t *event) {
    ...
    fprintf(stderr, "clear_singlestep_event vcpus: %u\n", event->ss_event.vcpus);
    ...
}
```

tklengyel:
Fixes #417 
Fixes #420 

The physical_address and the generic bits are required if the user calls vmi_clear_event in a callback. We simply retire physical_address and use simple gfn going forward, eliminating the need for the union of IN/OUT structs.

Convert some IN fields to CONST IN so that it's more intuitive and conforms to the intended use of vmi_step_event and avoids having to reset fields before calling vmi_clear_event.

Bumping events version as this is an events ABI change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,8 @@ examples/step-event-example
 examples/step-event-example.o
 examples/va-pages
 examples/va-pages.o
+examples/xen-emulate-response
+examples/xen-emulate-response.o
 install-sh
 libtool
 libvmi.pc
--- a/examples/event-example.c
+++ b/examples/event-example.c
@@ -50,8 +50,7 @@ vmi_event_t kernel_vsyscall_event;
 vmi_event_t kernel_sysenter_target_event;
 
 void print_event(vmi_event_t event){
-    printf("PAGE %"PRIx64" ACCESS: %c%c%c for GFN %"PRIx64" (offset %06"PRIx64") gla %016"PRIx64" (vcpu %"PRIu32")\n",
-        event.mem_event.physical_address,
+    printf("PAGE ACCESS: %c%c%c for GFN %"PRIx64" (offset %06"PRIx64") gla %016"PRIx64" (vcpu %"PRIu32")\n",
         (event.mem_event.out_access & VMI_MEMACCESS_R) ? 'r' : '-',
         (event.mem_event.out_access & VMI_MEMACCESS_W) ? 'w' : '-',
         (event.mem_event.out_access & VMI_MEMACCESS_X) ? 'x' : '-',
@@ -334,20 +333,17 @@ int main (int argc, char **argv)
     memset(&msr_syscall_sysenter_event, 0, sizeof(vmi_event_t));
     msr_syscall_sysenter_event.version = VMI_EVENTS_VERSION;
     msr_syscall_sysenter_event.type = VMI_EVENT_MEMORY;
-    msr_syscall_sysenter_event.mem_event.physical_address = phys_sysenter_ip;
-    msr_syscall_sysenter_event.mem_event.npages = 1;
+    msr_syscall_sysenter_event.mem_event.gfn = phys_sysenter_ip >> 12;
 
     memset(&kernel_sysenter_target_event, 0, sizeof(vmi_event_t));
     kernel_sysenter_target_event.version = VMI_EVENTS_VERSION;
     kernel_sysenter_target_event.type = VMI_EVENT_MEMORY;
-    kernel_sysenter_target_event.mem_event.physical_address = phys_ia32_sysenter_target;
-    kernel_sysenter_target_event.mem_event.npages = 1;
+    kernel_sysenter_target_event.mem_event.gfn = phys_ia32_sysenter_target >> 12;
 
     memset(&kernel_vsyscall_event, 0, sizeof(vmi_event_t));
     kernel_vsyscall_event.version = VMI_EVENTS_VERSION;
     kernel_vsyscall_event.type = VMI_EVENT_MEMORY;
-    kernel_vsyscall_event.mem_event.physical_address = phys_vsyscall;
-    kernel_vsyscall_event.mem_event.npages = 1;
+    kernel_vsyscall_event.mem_event.gfn = phys_vsyscall >> 12;
 
     while(!interrupted){
         printf("Waiting for events...\n");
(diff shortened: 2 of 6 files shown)


bdpayne:
Looks good to me overall, but this part of the code base isn't my strong suit.  I'll let others review before we merge.

Please squash the multiple commits.

smaresca:
Simplifying the in/out aspect is a net improvement for sure.

@tklengyel while you're at it, could you add a comment underscoring the meaning of ‘generic’ ? I do not believe it is clearly defined anywhere. Probably best adjacent to mem_access_event_t 

Just a general observation:  I think we’ve flip-flopped between page # and physical addr one or twice.  No matter where we land here, I think it’s crucial that we stick to the decision.. As @K-Mayer noted in PR#373, it ends up being a matter of taste ( https://github.com/libvmi/libvmi/pull/373#issuecomment-248656774 ) .  Both are correct representations, and both require the library user to understand the implications.

tklengyel:
@smaresca: more verbose comment for 'generic' is added.

PR #424, opened 2016-12-07
Wrap all hypervisor library calls

### LINKED ISSUE #423: No hard library dependencies for drivers
Currently when LibVMI is compiled it will link with shared libraries to support Xen and/or KVM. This results in a LibVMI shared library that requires that the target system have the same libraries deployed. However, Xen hosts generally lack libvirt libraries and KVM hosts lack the Xen libraries. This effectively requires LibVMI to be compiled with either of the drivers disabled, making it either libvmi-xen or libvmi-kvm.

To alleviate this issue LibVMI should check at runtime which of the libraries are present and make drivers available when the supporting libraries have been found.

tklengyel:
Fixes #423 

This PR allows us to compile all supported driver code into LibVMI but only activate the drivers when the run-time environment supports it. With this PR the library no longer has a hard-dependency on all shared libraries being present at once.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -176,7 +176,7 @@ xen_space='      '
 
     [if test "$with_xenstore" = "yes"]
     [then]
-        AC_CHECK_LIB(xenstore, xs_read, [], [missing="yes"])
+        AC_CHECK_LIB(xenstore, xs_read, [missing="no"], [missing="yes"])
         AC_CHECK_HEADERS([xenstore.h])
         AC_CHECK_HEADERS([xs.h])
     [fi]
@@ -188,7 +188,12 @@ xen_space='      '
         enable_xen='no'
         have_xen='missing xenstore'
     [else]
-        AC_CHECK_LIB(xenctrl, xc_interface_open, [], [missing="yes"])
+        [if test "$with_xenstore" = "yes"]
+        [then]
+            AC_DEFINE(HAVE_LIBXENSTORE, [1], [Define to 1 if enable xenstore access])
+        [fi]
+
+        AC_CHECK_LIB(xenctrl, xc_interface_open, [missing="no"], [missing="yes"])
         [if test "$missing" = "yes"]
         [then]
             AC_DEFINE([ENABLE_XEN], [0], [Define to 1 to enable Xen support.])
--- a/configure.ac
+++ b/configure.ac
@@ -245,7 +245,7 @@ kvm_space='      '
         enable_kvm='no'
         have_kvm='libm missing'
     [else]
-        AC_CHECK_LIB(virt, virConnectOpen, [], [missing="yes"])
+        AC_CHECK_LIB(virt, virConnectOpen, [missing="no"], [missing="yes"])
         [if test "$missing" = "yes"]
         [then]
             AC_DEFINE([ENABLE_KVM], [0], [Define to 1 to enable KVM support.])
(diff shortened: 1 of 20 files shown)


tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

MikePetullo:
This type of thing almost always benefits binary packages, because users are sometimes hesitant to install a package which depends on all of the libraries required by all of its various compile-time backends. I had originally built the Fedora libvmi package to support only Xen, but the package reviewer asked that I build the KVM support too. (I myself use only Xen.) Thus the Fedora package does indeed pull in more dependencies than a given user probably strictly needs.

On the technical side, I had thought that libvirt already abstracted away Xen's interfaces. I suppose it is possible that libvirt provides access only to the "lowest common denominator" features of Xen. My impression was that there were some SLAT features in Xen/libvmi that are not available on other platforms, but I might be wrong.

In any case, we should benchmark the proposal to ensure its runtime overhead is negligible.

To help with the decision, here is a bit on the impact of the present design. The Fedora libvmi package requires:

        /sbin/ldconfig
        /sbin/ldconfig
        libc.so.6()(64bit)
        libdl.so.2()(64bit)
        libglib-2.0.so.0()(64bit)
        libjson-c.so.2()(64bit)
        libm.so.6()(64bit)
        libvirt.so.0()(64bit)
        libxenctrl.so.4.7()(64bit)
        libxenstore.so.3.0()(64bit)
        rpmlib(CompressedFileNames) <= 3.0.4-1
        rpmlib(FileDigests) <= 4.6.0-1
        rpmlib(PayloadFilesHavePrefix) <= 4.0-1
        rpmlib(PayloadIsXz) <= 5.2-1
        rtld(GNU_HASH)

The libvirt-libs package itself requires a number of other libraries as  does the xen-libs package. In both cases, most of these further dependencies are common, such as liblzma and libext2fs. I would suspect this adds up to a few MB.

Related to this is the debugging features of libvmi. I would like to see an environment variable which would turn on the debugging features which presently require a compile-time flag. I do not think I have yet publicly proposed this. For an existing exemplar, see:

        https://developer.gnome.org/glib/stable/glib-running.html

tklengyel:
Thanks for the input! That's pretty much the idea behind this approach - don't force the user to install packages obviously unrelated to his current environment. Sometimes that is simple not possible, or not easily. OTOH maintaining multiple versions of packages built for specific environments is a PITA. So this is middle-way out. There are some problems of course, like the Xen package on CentOS doesn't actually install the symlink for libxenctrl.so, only for libxenctrl.so.<MAJOR>.<MINOR>.. A workaround is to just make the libvmi package provide this symlink to whatever libxenctrl version is present.

As for libvirt, libvmi being a fairly low-level library will need access to Xen-functions that libvirt currently doesn't expose, SLAT being just one.

Benchmarking the proposal is up for grabs, but I think we are on par with linking normally with the library (aka. filling in the import table at launch time). We just effectively have our own version of the import table that is selectively filled up based on the environment.

As for the debugging output being dependent upon an environment variable I'm not too much in favor as it adds a lot of if checks and bitmasking ops which I would not want to see in a production-version. There really should be no reason a user would need that output. LibVMI already prints critical errors to stderr which aid the user with common misconfiguration related issues. All the other debug output IMHO is only relevant for actual LibVMI development.

smaresca:
@MikePetullo Thank you for the feedback, much appreciated.

Agree with @tklengyel, overhead here should be negligible. Function pointers here are essentially doing the same job as the ELF PLT. Instead, libVMI does the work at init rather than the dynamic linker. At worst, that would requiring an additional jmp or indirect branch.  I think I'm interested in a benchmark too, but I wouldn't consider it a blocker for merge given that this PR improves overall compatibility.

PR #425, opened 2016-12-13
get offset of symbol+subsymbol

scorpionzezz:
This patch is to get the symbol + sub-symbol offset from the rekall profile.

### DIFF
--- a/libvmi/accessors.c
+++ b/libvmi/accessors.c
@@ -141,6 +141,16 @@ vmi_get_offset(
     return vmi->os_interface->os_get_offset(vmi, offset_name);
 }
 
+status_t
+vmi_get_kernel_struct_offset(
+    vmi_instance_t vmi,
+    const char* symbol,
+    const char* member,
+    addr_t *addr)
+{
+    return vmi->os_interface->os_get_kernel_struct_offset(vmi,symbol,member,addr);
+}
+
 uint64_t
 vmi_get_memsize(
     vmi_instance_t vmi)
--- a/libvmi/libvmi.h
+++ b/libvmi/libvmi.h
@@ -1731,6 +1731,21 @@ uint64_t vmi_get_offset(
     const char *offset_name);
 
 /**
+ * Get the memory offset associated with the given symbol and subsymbol in the struct
+ * @param[in] vmi LibVMI instance
+ * @param[in] struct_name String name for desired symbol
+ * @param[in] subsymbol String name for desired subsymbol
+ * @param[out] offset of subsymbol in struct (symbol)
+ *
+ * @return VMI_SUCCESS or VMI_FAILURE
+ */
+status_t vmi_get_kernel_struct_offset(
+    vmi_instance_t vmi,
+    const char* struct_name,
+    const char* member,
+    addr_t *addr);
+
+/**
  * Gets the memory size of the guest or file that LibVMI is currently
  * accessing.  This is the amount of RAM allocated to the guest, but
  * does not necessarily indicate the highest addressable physical address;
--- a/libvmi/os/linux/core.c
+++ b/libvmi/os/linux/core.c
@@ -354,6 +354,7 @@ status_t linux_init(vmi_instance_t vmi) {
     os_interface = safe_malloc(sizeof(struct os_interface));
     bzero(os_interface, sizeof(struct os_interface));
     os_interface->os_get_offset = linux_get_offset;
+    os_interface->os_get_kernel_struct_offset = linux_get_kernel_struct_offset;
     os_interface->os_pid_to_pgd = linux_pid_to_pgd;
     os_interface->os_pgd_to_pid = linux_pgd_to_pid;
     os_interface->os_ksym2v = linux_symbol_to_address;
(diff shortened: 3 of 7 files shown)


libvmi/os/windows/core.c
```diff
@@ -762,6 +762,7 @@ windows_init(
     /* Need to provide this functions so that find_page_mode will work */
     os_interface = safe_malloc(sizeof(struct os_interface));
     bzero(os_interface, sizeof(struct os_interface));
+    os_interface->os_get_symbol_offset = NULL;
```
tklengyel: Any particular reason why this is skipped for Windows?

libvmi/accessors.c
```diff
@@ -141,6 +141,16 @@ vmi_get_offset(
     return vmi->os_interface->os_get_offset(vmi, offset_name);
 }
 
+status_t
+vmi_get_kernel_offset(
+    vmi_instance_t vmi, 
+    const char* symbol, 
+    const char* subsymbol,
```
tklengyel: I'm not particularly happy with this API. For internal use in LibVMI it's fine but it is really not intuitive to be publicly supported. Ie., when you don't specify a subsymbol it is looking for a kernel function / static variable, whereas with a subsymbol it looks for an offset in struct. These should really be two separate functions on the public side, as "vmi_get_kernel_offset" doesn't describe it accurately.

scorpionzezz:
Hey there, thanks for your feedback :)

I did fix it, and gave the appropriate name and also implemented it for windows. May you take a look it again please?

tklengyel:
This is ok to test.

tklengyel:
Retest this please.

tklengyel:
@smaresca @bdpayne @awsaba looks fine to me but needs another +1 because of the public API addition

libvmi/libvmi.h
```diff
@@ -1731,6 +1731,34 @@ uint64_t vmi_get_offset(
...
+ * @param[in] subsymbol String name for desired subsymbol
+ * @param[out] offset of subsymbol in struct (symbol)
+ *
+ * @return VMI_SUCCESS or VMI_FAILURE
+ */
+status_t vmi_get_kernel_struct_offset(
+    vmi_instance_t vmi,
+    const char* struct_name,
+    const char* subsymbol,
+    addr_t *rva);
```
smaresca: Parameters named subsymbol and rva seem strange in the context of this function. I understand that they originate at a Rekall layer, but for public exposure in the LibVMI API, please keep them as generic as possible.

libvmi/libvmi.h
```diff
@@ -1731,6 +1731,34 @@ uint64_t vmi_get_offset(
...
+
+/**
+ * Get the memory offset of a kernel function or a symbol
+ * @param[in] vmi LibVMI instance
+ * @param[in] symbol name for the desired symbol
+ * @param[out] offset of the kernel function or the symbol
+ *
+ * @return VMI_SUCCESS or VMI_FAILURE
+ */
+status_t vmi_get_constant_offset(
```
smaresca: Would this be more accurate to describe as returning a va of a kernel symbol? If so, could this instead be worked into the vmi_translate_ksym2v (e.g., if symbol not found elsewhere and rekall profile exists, call vmi->os_interface->os_get_constant_offset(..) ?)

libvmi/libvmi.h
tklengyel: Any suggestions what to name them?

libvmi/libvmi.h
tklengyel: This would make sense actually, +1

libvmi/libvmi.h
scorpionzezz: How about I change rva to addr, to make it more generic?

libvmi/libvmi.h
scorpionzezz: Ah yes, this make sense, so I remove the vmi_get_constant_offset in my latest commit.

libvmi/os/linux/core.c
```diff
@@ -433,6 +434,16 @@ void linux_read_config_ghashtable_entries(char* key, gpointer value,
     _done: return;
 }
 
+status_t linux_get_kernel_struct_offset(vmi_instance_t vmi, const char* symbol, const char* subsymbol, addr_t *addr){
+    status_t ret = VMI_SUCCESS;
+    linux_instance_t linux_instance = vmi->os_data;
+    if (VMI_FAILURE == rekall_profile_symbol_to_rva(linux_instance->rekall_profile,symbol,subsymbol,addr)){
```
tklengyel: You can just return the call to rekall_profile_symbol_to_rva directly, no need, for the local variable ret.

libvmi/os/windows/core.c
```diff
@@ -414,6 +414,16 @@ get_kpgd_method0(
     return VMI_FAILURE;
 }
 
+status_t windows_get_kernel_struct_offset(vmi_instance_t vmi, const char* symbol, const char* subsymbol, addr_t *addr){
+    status_t ret = VMI_SUCCESS;
+    windows_instance_t windows = vmi->os_data;
+    if (VMI_FAILURE == rekall_profile_symbol_to_rva(windows->rekall_profile,symbol,subsymbol,addr)){
```
tklengyel: You can just return the call to rekall_profile_symbol_to_rva directly, no need, for the local variable ret.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

libvmi/os/linux/core.c
scorpionzezz: okay, I changed it in my latest commit

libvmi/os/windows/core.c
scorpionzezz: okay, I changed it in my latest commit

libvmi/libvmi.h
smaresca: rva to addr makes sense sure. Please use 'member' insted of subsymbol as well

scorpionzezz:
done, changed subsymbol -> member

smaresca:
Thanks @scorpionzezz  !

scorpionzezz:
no problem, thanks @smaresca  and @tklengyel  for developing this amazing tool :)

PR #426, opened 2016-12-13
Add guest page-table walk fault information to mem_access events

tklengyel:
Also, only copy gla if GLA_VALID is set, otherwise set it to ~0ull.

Bumping events version as this is an ABI change.


tklengyel:
Retest this please.

libvmi/driver/xen/msr-index.h
```diff
@@ -0,0 +1,127 @@
...
+
+    [MSR_IA32_FEATURE_CONTROL]   = 0x0000003a,
+
+    [MSR_IA32_SYSENTER_CS]       = 0x00000174,
+    [MSR_IA32_SYSENTER_ESP]      = 0x00000175,
+    [MSR_IA32_SYSENTER_EIP]      = 0x00000176,
+
+    [MSR_IA32_MISC_ENABLE]       = 0x000001a0,
+
+    [MSR_HYPERVISOR]             = 0x40000000
```
smaresca: Contextual information for others (which might be useful in a comment?): The name of this MSR entry, MSR_HYPERVISOR, is not really defined anywhere outside of LibVMI, though I believe it makes sense. 

In general, the MSR range 0x40000000 - 0x400000FF was designated by Intel&AMD for software defined use, so it is possible that it could be used for something other than publishing hypervisor details to guests. However, since 2008, the low end of his range has become hypervisor oriented as a sort of defacto standard.

This interface proposal https://lwn.net/Articles/301888/ coauthored by VMware, KVM, and Xen developers is as close as we get to standardization. 0x40000000 is for hypervisor identification and 0x40000001 - 0x4000000F is designated as a range available for hypervisor-specific usage.

libvmi/events.h
```diff
@@ -33,7 +33,7 @@
 #ifndef LIBVMI_EVENTS_H
 #define LIBVMI_EVENTS_H
 
-#define VMI_EVENTS_VERSION 0x00000002
+#define VMI_EVENTS_VERSION 0x00000003
```
tklengyel: PR description already stated "Bumping events version as this is an ABI change."

PR #429, opened 2016-12-17
Fix windows-offset-tool scripts downloading PDB

mattshockl:
windows-offset-tool currently does not work, as it cannot download the PDB file from the microsoft website using the python urlbuilder (at least on Fedora).  This fix that, and reduces the complexity of the code used.


tklengyel:
This is ok to test.

MikePetullo:
I can confirm that Microsoft changed their website, and that this broke the existing download script. The patch from mattshockl worked for me.

PR #430, opened 2016-12-22
Fix MSR events on Xen 4.8

tklengyel:
Starting with Xen 4.8 you have to subscribe to specific MSRs to receive events. For now we shortcut using MSR_ALL to set common MSRs. However, to fully support arbitrary MSR interception we will likely need to define all MSRs in libvmi.h and add the corresponding msr-indexes.


tklengyel:
I'll move this work to the gptw PR so we only have to bump the events ABI one time.

PR #433, opened 2017-01-12
Fix Windows RVA to Symbol lookup

srujan:
Name ordinal should be _multiplied_ with function address size to get the corresponding Function RVA rather than adding.


tklengyel:
This is OK to test.

tklengyel:
Thanks!

PR #434, opened 2017-01-15
Get and set multiple registers on Xen

### LINKED ISSUE #133: Get and set mulitple registers in one call
The current implementation of the get and set register functions work only with a single register at a time, which is quite wasteful when multiple registers are needed to be either get or set. If the hypervisor supports accessing all vCPU registers in one shot (like the HVM context on Xen) adding this feature is going to be better performance wise.

tklengyel:
Fixes #133

### DIFF
--- a/libvmi/libvmi.h
+++ b/libvmi/libvmi.h
@@ -158,7 +158,6 @@ typedef enum page_size {
 } page_size_t;
 
 typedef uint64_t reg_t;
-typedef reg_t registers_t;
 
 /**
  * The following definitions are used where the API defines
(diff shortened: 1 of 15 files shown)


libvmi/libvmi.h
```diff
@@ -1856,8 +1900,23 @@ unsigned int vmi_get_num_vcpus (
...
-    reg_t *value,
-    registers_t reg,
+    uint64_t *value,
+    reg_t reg,
+    unsigned long vcpu);
+
+/**
+ * Gets the current value of a VCPU registers.  This currently only
+ * supports x86 registers.  When LibVMI is accessing a raw
+ * memory file or KVM, this function will fail.
```
smaresca: Curious: what is needed for arm?

Also, s/Gets the current value of a VCPU registers/Gets the current value of all VCPU registers/

tklengyel:
Retest this please.

libvmi/libvmi.h
tklengyel: Typo fixed. For ARM we would need an arm_regs struct defined for registers similar to the x86 regs we use here.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

PR #435, opened 2017-01-16
Init rework

### LINKED ISSUE #287: Unify vmi_init functions
Would like to unify the vmi_init series of functions to just vmi_init() so we don't have so many methods of initializing the library. We are already passing the flags and info we need in the vmi_init_custom to make the decisions about the init and having this many init functions is going to confuse people. Especially newer users.

tklengyel:
Fixes #287 

This PR sanitizes the various init functions so that we have a more intuitive API. Here we deprecate the init_custom functions and rework vmi_init and vmi_init_complete as follows.

- vmi_init will allocate the vmi_instance and then only initialize the hypervisor driver interface of LibVMI (ie. only physical memory access and vCPU register functions). This requires the user to specify the hypervisor mode. VMI_AUTO mode is deprecated, instead a dedicated function is exposed - vmi_get_access_mode - the user can call if automatical hypervisor identification is needed. The inputs to this function are made more flexible but allowing the user to pass various domain-identification inputs, such as domain name or domain id. This approach will allow us to add other types of IDs later if necessary (ie. UUID) without having to change the API. A new field is added that can further carry information that may be required for the hypervisor driver to init (as anticipated for the VirtualBox driver).

 - vmi_init_paging remains largely as it was, initializing the v2p translation architecture specific functions of LibVMI. A new flags input is added to be able to configure paging specific features (ie. transition-pages) that previously relied on the OS interface being initialized. Calling this function multiple times will re-set the interface automatically.

 - vmi_init_os will initialize the OS specific components of LibVMI (ie. vmi_*_ksym) and this is the only init function that requires further configuration to be provided.

- vmi_init_complete will call all previous functions automatically - going from vmi_get_access_mode to vmi_init_os

### DIFF
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -44,20 +44,24 @@
 #ifndef ENABLE_CONFIGFILE
 static inline status_t
 read_config_file(vmi_instance_t UNUSED(vmi),
-                 FILE* UNUSED(config_file))
+                 FILE* UNUSED(config_file),
+                 GHashTable** UNUSED(_config))
 {
     return VMI_FAILURE;
 }
 
 static inline status_t
 read_config_string(vmi_instance_t UNUSED(vmi),
-                   const char* UNUSED(config))
+                   const char* UNUSED(config)
+                   GHashTable** UNUSUED(_config))
 {
     return VMI_FAILURE;
 }
 
 static inline status_t
-read_config_file_entry(vmi_instance_t UNUSED(vmi))
+read_config_file_entry(vmi_instance_t UNUSED(vmi),
+                       GHashTable** UNUSED(config),
+                       vmi_init_error_t* UNUSED(error))
 {
     return VMI_FAILURE;
 }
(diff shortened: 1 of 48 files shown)


tklengyel:
Retest this please.

dalevy:
*disclaimer I skimmed this PR*

Any thoughts on 

status = {VMI_SUCCESS,VMI_FAILURE,VMI_FAILURE_FILE_NOT_FOUND, VMI_FAILURE_UNKNOWN_VM, etc, etc ...} I reckon making return type enumerations more explicit would not be feasible as it would break backwards compatibility so maybe...additional functions:


//warning, pseudo code follows...

status_t value;
value = vmi_init_verbose( params...);

if ( value == (VMI_FAILURE | VMI_FAILURE_UNKNOWN_VM)){
  //do this when the vm is not there
}
else if (value == (VMI_FAILURE | VMI_FAILURE_FILE_NOT_FOUND){
  //do this when the rekall profile or libvmi config file is not found
}


//or in a more cumbersome api programmer friendly way
typedef struct vmi_init_error {
    status_t value;
    char *usefulMessage;

   ..with future implementation options 
};

vmi_init_error err;
err = vmi_init_verbose( params...);

The struct return value and definition of the api is probably overkill but one can dream I guess :)

tklengyel:
@dalevy Instead of having a fully separate function to do this, I would much rather just use an errno style feedback on vmi_init_complete for those users that need more information about an initialization failure. Also, with the functions as broken down in this PR the user can easily determine what part of the initialization failed (ie. hypervisor or vm not found, paging wasn't identified, OS wasn't found, etc.) if you are not using vmi_init_complete.

tklengyel:
@dalevy I've updated the PR so that both `vmi_init` and `vmi_init_complete` now has an extra optional input for `vmi_init_error_t`. In case the initialization fails and this input pointer is specified, it will contain the specific error that occurred. Can you take a look and check whether this will address your concern?

dalevy:
@tklengyel Thanks for making the changes. Is it possible to "bubble" a config file parse/existence failure up to vmi_init_error_t if config file/rekall profile is missing? Other than that, this change addresses my concerns.

tklengyel:
@dalevy Yeap, can add any number of errors to the vmi_init_error enum.

tklengyel:
@dalevy I've added two more error conditions: `VMI_INIT_ERROR_NO_CONFIG` and `VMI_INIT_ERROR_NO_CONFIG_ENTRY`. The first to specify if there was no valid config file found, the second if there was no entry found in the config for the target VM.

dalevy:
@tklengyel Great! These changes satisfy my use cases.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

smaresca:
@tklengyel Could you add a null check for tmp_name here https://github.com/tklengyel/libvmi/blob/3293cd73849943be7f6cd9410dd2157fcf44bf96/libvmi/core.c#L369  ?

scan-build report showed a potential issue there

tklengyel:
@smaresca: I've added a couple more NULL checks too in that code area.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

libvmi/core.c
```diff
@@ -367,9 +369,11 @@ set_id_and_name(
     if (VMI_SUCCESS == driver_get_name_from_id(vmi, id, &tmp_name)) {
         dbprint(VMI_DEBUG_CORE, "--got name from id (%"PRIu64" --> %s)\n", id, tmp_name);
         vmi->image_type = strndup(tmp_name, 100);
```
smaresca: The issue here that I mentioned had to do with the passing of tmp_name into strndup (a null-intolerant function) without checking after it was set in driver_get_name_from_id

tklengyel: driver_get_name_from_id returns VMI_FAILURE if it failed, so if we are in here, we know that tmp_name is not NULL.

tklengyel:
Retest this please.

PR #438, opened 2017-01-24
Search rekall profile for symbols in $FUNCTIONS by default

mattshockl:
Allow for LibVMI to also search the $functions array along with $constants array by default.  Makes using ksym2v much easier to use, especially for many function symbols in windows.


tklengyel:
This is ok to test.

tklengyel:
Looks good to me! Thanks.

PR #441, opened 2017-01-26
Add support for VCPU-safe register modification w/ HVMs

mattshockl:
Allows for HVMs to modify certain supported registers without worrying about race conditions between different VCPUs.  Couldn't completely rely on xen_set_vcpureg_pv without losing support of many control registers (e.g. MSRs) supported by HVMs

### DIFF
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -2455,15 +2455,50 @@ xen_set_vcpureg(
 #if defined(ARM32) || defined(ARM64)
     return xen_set_vcpureg_arm(vmi, value, reg, vcpu);
 #elif defined(I386) || defined (X86_64)
-    if (!xen_get_instance(vmi)->hvm) {
-        if (8 == xen_get_instance(vmi)->addr_width) {
-            return xen_set_vcpureg_pv64(vmi, value, reg, vcpu);
-        } else {
-            return xen_set_vcpureg_pv32(vmi, value, reg, vcpu);
+    if (xen_get_instance(vmi)->hvm) {
+        switch (reg) {
+        case RAX:
+        case RBX:
+        case RCX:
+        case RDX:
+        case RBP:
+        case RSI:
+        case RDI:
+        case RSP:
+        case R8:
+        case R9:
+        case R10:
+        case R11:
+        case R12:
+        case R13:
+        case R14:
+        case R15:
+        case RIP:
+        case RFLAGS:
+        case CR0:
+        case CR2:
+        case CR4:
+        case DR0:
+        case DR1:
+        case DR2:
+        case DR3:
+        case DR6:
+        case DR7:
+        case FS_BASE:
+        case GS_BASE:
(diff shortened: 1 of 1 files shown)


I-am-Batman:
Wow this worked great for me too! Thank you mattshockl.

libvmi/driver/xen/xen.c
```diff
@@ -2455,15 +2455,51 @@ xen_set_vcpureg(
...
+        case R11:
+        case R12:
+        case R13:
+        case R14:
+        case R15:
+        case RIP:
+        case RFLAGS:
+        case CR0:
+        case CR2:
+        case CR3:
```
tklengyel: Could you verify that setting CR3 is indeed going to work through this route? The pv setting functions perform some sanitization on the cr3 value passed from the user and it may not be actually needed for HVM guests. Ie. just try to set the CR3 to the same value that was received.

tklengyel:
This is OK to test.

libvmi/driver/xen/xen.c
mattshockl: So it seems `xen_pfn_to_cr3_x86_64(value >> 12)` just left-shifts it back 12 bits, essentially zeroing out the bottom 12 bits. I ran the following code about 1000 times with no issues, so unless there is some need for the bottom 12 bits of CR3, it shouldn't affect anything

```
reg_t tmp;
vmi_get_vcpureg(vmi, &tmp, CR3, event->vcpu_id);
vmi_set_vcpureg(vmi, tmp, CR3, event->vcpu_id);
```

tklengyel: Unfortunately if the guest is using PCID and we set CR3 through here we would have some unintended consequences (PCID uses the lower 12 bits of the CR3 as the TLB tag). So IMHO that cr3 sanitization should only take place if the guest is PV.

mattshockl: I just updated the code to remove CR3 support.   It seems like PV doesn't manipulate any of the other registers so everything should be equivalent to the VCPU-safe HVM methods

tklengyel:
Looks good to me! This is a nice improvement compared to having to get and set all vCPU contexts just for changing some of these registers. However, due to the fact that it doesn't cover all registers, the comment in libvmi.h about having to pause the entire domain before using this function I think is still appropriate as general guideline for users.

tklengyel:
Before merging, please squash the two commits in this PR to just one. Thanks!

mattshockl:
Just squashed (I think).  Should we amend the comment to say that some registers are VCPU-safe w/ HVMs?

tklengyel:
Thanks. And no, I don't think that's necessary as it will just be a source of error. It is also very heavily Xen specific.

tklengyel:
There are regressions introduced by this PR so it has been reverted for now. Setting RIP or RSP for 32bit HVM guests has been resulting in application crashes. Root cause needs to be better understood and need to determine if other registers are affected as well.

PR #444, opened 2017-01-30
Handle low-memory scenarios gracefully

tklengyel:
In low-memory scenarios crashing because of cache memory allocations is not acceptable and has to be handled more gracefully. This PR replaces all uses of safe_malloc in cache.c with g_malloc0 and adds appropriate tests to see whether memory was indeed allocated.

Also, alter vmi_v2p_cacheflush to add an option to flush by DTB. This is useful in multi-vCPU VM setups where flushing the whole cache is not always necessary.

### DIFF
--- a/libvmi/cache.c
+++ b/libvmi/cache.c
@@ -79,8 +79,9 @@ void key_128_init(vmi_instance_t vmi, key_128_t key, uint64_t low, uint64_t high
 
 key_128_t key_128_build (vmi_instance_t vmi, uint64_t low, uint64_t high)
 {
-    key_128_t key = (key_128_t) safe_malloc(sizeof(struct key_128));
-    key_128_init(vmi, key, low, high);
+    key_128_t key = (key_128_t) g_malloc0(sizeof(struct key_128));
+    if ( key )
+        key_128_init(vmi, key, low, high);
     return key;
 }
 
@@ -90,7 +91,6 @@ key_128_t key_128_build (vmi_instance_t vmi, uint64_t low, uint64_t high)
 struct pid_cache_entry {
     vmi_pid_t pid;
     addr_t dtb;
-    time_t last_used;
 };
 typedef struct pid_cache_entry *pid_cache_entry_t;
 
@@ -114,10 +114,13 @@ static pid_cache_entry_t pid_cache_entry_create(
     addr_t dtb)
 {
     pid_cache_entry_t entry =
-        (pid_cache_entry_t) safe_malloc(sizeof(struct pid_cache_entry));
+        (pid_cache_entry_t) g_malloc0(sizeof(struct pid_cache_entry));
+
+    if ( !entry )
+        return NULL;
+
     entry->pid = pid;
     entry->dtb = dtb;
-    entry->last_used = time(NULL);
     return entry;
 }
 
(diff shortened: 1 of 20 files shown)


tklengyel:
Retest this please.

tklengyel:
Retest this please.

libvmi/cache.c
```diff
@@ -162,12 +164,23 @@ pid_cache_set(
...
+        return;
 
     *key = pid;
     pid_cache_entry_t entry = pid_cache_entry_create(pid, dtb);
 
-    g_hash_table_insert(vmi->pid_cache, key, entry);
+    if ( !entry )
+        return;
+
+    if ( !g_hash_table_insert(vmi->pid_cache, key, entry) )
```
smaresca: The return value of g_hash_table_insert was unchecked deliberately here:  GLIB 2.38 and earlier returned void rather than gboolean 
https://developer.gnome.org/glib/2.38/glib-Hash-Tables.html#g-hash-table-insert

Supporting slightly older platforms is important, e.g. RHEL 6, so we need compat checks for these (or return to old unchecked behavior. tossup..result ends up being the same for a failure with current logic in most locations)

Lines 585 and 603 will generate a errors like "error: invalid use of void expression" and "error: void value not ignored as it ought to be" in their current form on older platforms as well.

libvmi/cache.c
```diff
@@ -292,22 +306,62 @@ sym_cache_set(
...
-        g_hash_table_insert(vmi->sym_cache, key, symbol_table);
-    } else {
-        free(key);
+        if ( !symbol_table )
+        {
+            g_free(key);
+            return;
+        }
+
+        if ( !g_hash_table_insert(vmi->sym_cache, key, symbol_table) )
```
smaresca: here too

libvmi/cache.c
```diff
@@ -292,22 +306,62 @@ sym_cache_set(
...
+
+    sym_dup = g_strndup(sym, 100);
+    if ( !sym_dup )
+    {
+        g_free(key);
+        g_free(entry);
+        return;
+    }
+
+    if ( !g_hash_table_insert(symbol_table, sym_dup, entry) )
```
smaresca: here too

libvmi/cache.c
```diff
@@ -524,14 +564,54 @@ v2p_cache_set(
...
+
+        v = g_hash_table_new_full(g_int64_hash, g_int64_equal, g_free, g_free);
+        if ( !v )
+            return;
+
+        addr_t *_dtb = g_memdup(&dtb, sizeof(addr_t));
+        if ( !_dtb )
+            return;
+
+        if ( !g_hash_table_insert(vmi->v2p_cache, _dtb, v) )
```
smaresca: here too

libvmi/cache.c
```diff
@@ -524,14 +564,54 @@ v2p_cache_set(
...
+    if ( !_va )
+        return;
+
+    addr_t *_pa = g_malloc0(sizeof(addr_t));
+    if ( !_pa )
+        return;
+
+    *_pa = pa & ~VMI_BIT_MASK(0,11);
+
+    if ( g_hash_table_insert(v, _va, _pa) )
```
smaresca: here too

libvmi/cache.c
tklengyel: Damn glib.. I'll look into this.

tklengyel: Fixed.

smaresca:
g_hash_table_insert_compat looks good, will watch for when this has passed the build gate

tklengyel:
Retest this please.

tklengyel:
@smaresca: good to go (just had to set the PKG_CONFIG_PATH appropriately in the ARM cross-compile chroot for the configure script to work as intended).

PR #447, opened 2017-02-01
Ensure cache coherence on ARM after writing to memory

tklengyel:


smaresca:
lgtm 

Caveat: the +1 is from review of the run+build tests, as my local arm node is out of commission at the moment

tklengyel:
That should be fine, the PR test-gate does both a cross-compile build test, as well as a compile + run on an ARM node (odroid-xu).

PR #450, opened 2017-02-05
Add support for establishing a Xen event channel from outside LibVMI

MikePetullo:
Signed-off-by: W. Michael Petullo <mike@flyn.org>

### DIFF
--- a/libvmi/libvmi.h
+++ b/libvmi/libvmi.h
@@ -55,6 +55,8 @@ extern "C" {
 
 #define VMI_INIT_SHM (1u << 3) /**< initialize SHM mode */
 
+#define VMI_INIT_XEN_EVTCHN (1u << 4) /**< use provided Xen file descriptor */
+
 typedef enum vmi_mode {
 
     VMI_XEN, /**< libvmi is monitoring a Xen VM */
@@ -651,7 +653,9 @@ typedef struct vmi_instance *vmi_instance_t;
  *                       to initialize further LibVMI features, such as events.
  * @param[in] init_data In case initialization requires additional information
  *                      for a given hypervisor, it can be provided via this
- *                      input.
+ *                      input. A subsequent call to vmi_destroy will release
+ *                      any handles provided here, and so the calling application
+ *                      cannot continue to use them after calling vmi_destroy.
  * @param[out] error Optional. If not NULL and the function returns VMI_FAILURE,
  *                   this will specify the stage at which initialization failed.
  * @return VMI_SUCCESS or VMI_FAILURE
@@ -684,7 +688,9 @@ status_t vmi_init(
  * @param[in] init_flags Additional flags to initialize
  * @param[in] init_data In case initialization requires additional information
  *                      for a given hypervisor, it can be provided via this
- *                      input.
+ *                      input. A subsequent call to vmi_destroy will release
+ *                      any handles provided here, and so the calling application
+ *                      cannot continue to use them after calling vmi_destroy.
  * @param[in] config_mode The type of OS configuration that is provided.
  * @param[in] config Configuration is passed directly to LibVMI (ie. in a string
  *                   or in a GHashTable) or NULL of global config file is used.
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -490,7 +490,7 @@ status_t vmi_init(
     vmi_mode_t mode,
     void* domain,
     uint32_t init_flags,
-    void* UNUSED(init_data),
+    void *init_data,
     vmi_init_error_t *error)
 {
     if ( VMI_FAILURE == driver_sanity_check(mode) )
@@ -515,7 +515,7 @@ status_t vmi_init(
     _vmi->init_flags = init_flags;
 
     /* driver-specific initilization */
-    if (VMI_FAILURE == driver_init(_vmi)) {
+    if (VMI_FAILURE == driver_init(_vmi, init_flags, init_data)) {
         if ( error )
             *error = VMI_INIT_ERROR_DRIVER;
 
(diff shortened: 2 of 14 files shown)


MikePetullo:
This is the first attempt at making it possible to integrate libvmi into a select/poll event loop. This implementation covers Xen 4.6/4.7 and 4.8, but I have not yet tested 4.8. I have not attempted to cover KVM because I do not have experience with KVM.

For example (the new function, used below, is vmi_event_get_fd):

static gboolean
gt_handle_vmi_event(GIOChannel *chan, GIOCondition condition, gpointer user_data)
{
        status_t status;
        GtLoop *loop = user_data;

        switch (condition) {
        case G_IO_IN:
                status = vmi_events_listen(loop->vmi, 0);
                if (status != VMI_SUCCESS) {
                        fprintf(stderr, "error waiting for events\n");
                        gt_running = FALSE;
                }
                break;
        default:
                fprintf(stderr, "error reading fd\n");
                break;
        }

        if (!gt_running) {
                g_main_loop_quit(loop->g_main_loop);
        }

        return gt_running;
}

int
main(int argc, char *argv[])
{
        ...
        vmi_fd = vmi_event_get_fd(vmi);
        channel_vmi = g_io_channel_unix_new(vmi_fd);
        gt_loop_add_watch(channel_vmi,
                                        G_IO_IN | G_IO_ERR | G_IO_HUP | G_IO_NVAL,
                                        gt_handle_vmi_event,
                                        loop);
        vmi_resume_vm(loop->vmi);
        g_main_loop_run(loop->g_main_loop);
        ...
}

tklengyel:
Exposing the file descriptor through LibVMI is a hypervisor specific feature-creep into the LibVMI API. As such I'm not a fan of this approach.

I would suggest that if your application wants to access the file-descriptor to have it setup the events channels using the Xen API, then use the new init system in PR #435 (https://github.com/libvmi/libvmi/pull/435/files#diff-0af4637539ddf2fd76051b0ac977e009L620) to pass in that information to LibVMI. The new init system has been specifically setup to address cases where a user would want to pass extra information for the hypervisor driver, like handles and descriptors it wouldn't be able or wouldn't want to replicate.

smaresca:
I had meant to chime in as well. I agree with Tamas..it's a sort of API
leakage we hope to avoid. With that said, I would be open to other
approaches; I had hoped to explore an alternative myself, but have not had
the time to do so. I was considering something using a func ptr callback
that could be invoked upon fd state change.

Ultimately, the exposure of a file descriptor forces a library user to
specialize code which is otherwise hypervisor agnostic, pulling in headers
for things like IOCTLs and errno values needed to interact with the file
descriptor.

A use case needing access to the file descriptor is, to me, suggestive that
something could be added to the xen driver instead.

Steve

tklengyel:
@smaresca: adding an extra callback interface for fd changes is also very hypervisor specific, even if we abstract away the headers/errnos. That API would be unlikely to be used for other hypervisors. So I'm not really a fan of that approach either. IMHO having the user pass in the fd for the driver during init is a much cleaner solution, using an API that was specifically made for that type of thing.

MikePetullo:
I think @tklengyel's approach would be fine. Just so I understand, this particular capability is not yet in the init_rework branch; is this correct? If so, would you be willing to add this functionality to the branch? If not, could you provide a simple example of what this would look like. (In the latter case, I must be overlooking something in the branch's code.)

tklengyel:
(That was me before just replied via the first email that hit my inbox.)

MikePetullo:
Okay, so the struct you describe might have an "fd" field among others. Perhaps if the field was not equal to -1, then the Xen driver would not initialize Xen itself, but would instead use the file descriptor fd. We would be very happy with this, and I agree it keeps the abstraction provided by libvmi at the right level. The application could take responsibility for peering down into the Xen level before passing things along to libvmi, but libvmi would not be responsible for itself exporting any of the Xen details back out.

We are hoping to release a prototype of the system that would use this in April. Might we be able to make progress on this before then? Or, should we plan to ship a locally-modified libvmi for the time being? I do not intend to apply any pressure; I just want to be able to plan our activities leading up to the prototype.

tklengyel:
Absolutely. The init rework PR will get merged shortly so you can treat it as semi-stable API that we will be migrating to. So you can start using it and basing your work on it. Also, feel free to open a PR for your work if you would want early feedback even before the init rework PR is merged.

MikePetullo:
I just pushed a prototype which explores my understanding of @tklengyel's design. I added a new, example VMI_INIT_XEN_EVTCHN flag and a vmi_init_data_t structure. A program can set fields in the vmi_init_data_t structure and pass it to vmi_init() along with flags that indicate which fields are set.

I was able to use this to permit guestrace access to the Xen event channel file descriptor.

libvmi/libvmi.h
```diff
@@ -661,7 +667,7 @@ status_t vmi_init(
     vmi_mode_t mode,
     void* domain,
     uint32_t init_flags,
-    void *init_data,
+    vmi_init_data_t *init_data,
```
tklengyel: Looks to me like it would be fine to keep this as a `void*` in the API. My reasoning with having this as a generic pointer was that the drivers can cast it to whatever type they need it as, and as long as its not a complex structure that needs to be passed in, there is no need for a struct definition in the public header.

tklengyel:
This is OK to test.

MikePetullo:
I removed the vmi_init_data_t structure.

tklengyel:
Looks good to me! One last thing, once LibVMI closes it will close this fd as well that was passed in. I think the public header should point this out so there are no surprises to anyone who later use this feature.

MikePetullo:
Just pushed a new version which adds some documentation and provides a more descriptive commit log.

tklengyel:
Great, thanks!

PR #452, opened 2017-02-09
Coverity fixes

tklengyel:


libvmi/driver/kvm/kvm.c
```diff
@@ -76,33 +76,46 @@ exec_qmp_cmd(
     char *query)
 {
     FILE *p;
-    char *output = safe_malloc(20000);
+    char *output = g_malloc0(20000);
+    if ( !output )
+        return NULL;
```
smaresca: goto error throughout here please? we're up to 6 returns otherwise..

tklengyel: I don't think we have a hard-rule that requires it either way. Since all the rest of the code here returns NULL instead of goto exit, this is IMHO also fine.

smaresca: certainly true, use of goto blah in the rest of libvmi is only a loose convention.

I'll dust off some of the work we did on code conventions and make a note in an issue so we don't pollute this PR too much.

PR #448, opened 2017-02-02
KASLR aware Linux address to symbol lookup

bentau:
The address to symbol lookup currently does not work when kasl is used!

PR #453, opened 2017-02-13
kaslr aware address to symbol lookup

scorpionzezz:
original idea: https://github.com/libvmi/libvmi/pull/448

### DIFF
--- a/libvmi/libvmi.h
+++ b/libvmi/libvmi.h
@@ -781,7 +781,6 @@ addr_t vmi_translate_sym2v(
 /**
  * Performs the translation from an RVA to a symbol
  * On Windows this function walks the PE export table.
- * The Linux translation works currently only for kernel symbols (pid=0 & base_vaddr=0).
  * Only the first matching symbol of System.map is returned.
  * ELF Headers are not supported.
  *
@@ -794,6 +793,21 @@ const char* vmi_translate_v2sym(
     vmi_instance_t vmi,
     const access_context_t *ctx,
     addr_t rva);
+
+/**
+ * Performs the translation from a VA to a symbol for Linux with KASLR offset
+ * Windows is not supported at this moment
+ * Only the first matching symbol of System.map is returned.
+ *
+ * @param[in] vmi LibVMI instance
+ * @param[in] ctx Access context
+ * @param[in] va VA to translate
+ * @return Symbol, or NULL on error
+ */
+const char* vmi_translate_v2ksym(
+    vmi_instance_t vmi,
+    const access_context_t *ctx,
+    addr_t va);
 
 /**
  * Given a pid, this function returns the virtual address of the
--- a/libvmi/os/linux/core.c
+++ b/libvmi/os/linux/core.c
@@ -359,7 +359,8 @@ status_t linux_init(vmi_instance_t vmi) {
     os_interface->os_pgd_to_pid = linux_pgd_to_pid;
     os_interface->os_ksym2v = linux_symbol_to_address;
     os_interface->os_usym2rva = NULL;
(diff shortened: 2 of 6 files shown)


libvmi/os/linux/core.c
```diff
@@ -360,6 +360,7 @@ status_t linux_init(vmi_instance_t vmi) {
     os_interface->os_ksym2v = linux_symbol_to_address;
     os_interface->os_usym2rva = NULL;
     os_interface->os_v2sym = linux_system_map_address_to_symbol;
```
tklengyel: This should be set to NULL now.

libvmi/os/linux/symbols.c
```diff
@@ -210,6 +210,17 @@ char* linux_system_map_address_to_symbol(
     return NULL;
 }
 
+char* linux_system_map_address_to_symbol_kaslr(
```
tklengyel: Do we need a separate function for this? I think the kaslr offset subtraction could just be moved into linux_system_map_address_to_symbol directly.

libvmi/libvmi.h
```diff
@@ -795,6 +795,11 @@ const char* vmi_translate_v2sym(
     const access_context_t *ctx,
     addr_t rva);
 
+const char* vmi_translate_v2ksym(
```
tklengyel: Missing function description comment.

libvmi/libvmi.h
```diff
@@ -795,6 +795,11 @@ const char* vmi_translate_v2sym(
     const access_context_t *ctx,
     addr_t rva);
```
tklengyel: v2sym description should be updated saying Linux is not supported.

scorpionzezz:
I changed it, I hope I get what you mean :)

libvmi/libvmi.h
```diff
@@ -781,7 +781,7 @@ addr_t vmi_translate_sym2v(
 /**
  * Performs the translation from an RVA to a symbol
  * On Windows this function walks the PE export table.
- * The Linux translation works currently only for kernel symbols (pid=0 & base_vaddr=0).
+ * Not supported in Linux
  * Only the first matching symbol of System.map is returned.
```
tklengyel: This line is not required either.

libvmi/libvmi.h
```diff
@@ -796,6 +796,21 @@ const char* vmi_translate_v2sym(
     addr_t rva);
 
 /**
+ * Performs the translation from an RVA to a symbol for Linux with KASLR offset
```
tklengyel: So this is not an RVA. This is an actual virtual address, so it should be called as such. We should also say this is not supported for Windows at the moment.

scorpionzezz: so maybe we should change it into VA instead or RVA ? :)

tklengyel: Yes, both in the comment and the variable name too.

scorpionzezz:
changed from RVA to VA :)

tklengyel:
This is ok to test.

libvmi/libvmi.h
```diff
@@ -796,6 +795,22 @@ const char* vmi_translate_v2sym(
     addr_t rva);
 
 /**
+ * Performs the translation from a VA to a symbol for Linux with KASLR offset
+ * Windows is not supported at this moment
+ * Only the first matching symbol of System.map is returned.
+ * ELF Headers are not supported.
```
tklengyel: The comment about ELF headers is not relevant here.

tklengyel:
Retest this please.

tklengyel:
Looks good to me. Needs another +1 for the API addition.

@smaresca @awsaba @bdpayne

PR #454, opened 2017-02-14
Fix pyvmi compilation

tklengyel:
`registers_t` is no longer the same as reg_t


tklengyel:
This is no longer enough to fix pyvmi compilation. @smaresca, what's your status on the Python3 rework? Should this be looked at in the meanwhile?

PR #455, opened 2017-02-15
Revert "Add support for VCPU-safe register modification w/ HVMs"

tklengyel:
Reverts libvmi/libvmi#441

Unfortunately there are some side-effects of using the PV register setting functions for HVM guests. The root cause of it is yet unknown but I have verified that setting either RIP or RSP through this method causes application crashes in Windows VMs.

Setting other registers through this method may also have issues, so until we have a better understanding of what causes this regression I'm reverting this patch.

@mattshockl


tklengyel:
Interestingly the issue only occurs for 32-bit HVM guests. So that might be a good clue to start investigating on.

PR #457, opened 2017-02-26
add memory access patch for QEMU v2.8.0

Wenzel:
This patch updates the memory access QMP command for QEMU v2.8.0.
- `qmp-commands.hx` has been replaced by `docs/qmp-commands.txt` since this [commit](https://github.com/qemu/qemu/commit/bd6092e407a5d682fbfddcbc510038b84bc1a1aa)
- tested with latest libvmi using the `process-list` example

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -66,6 +66,11 @@ YES_IS_DEFINED
 dnl -----------------------------------------------
 dnl Check package options
 dnl -----------------------------------------------
+ AC_CHECK_HEADER([qemu/libvmi_request.h],
+                 [AC_DEFINE([HAVE_LIBVMI_REQUEST], 1,
+                    [Define to 1 if you have <stdio.h>.])],
+                 [])
+
 AC_ARG_ENABLE([xen],
       [AS_HELP_STRING([--disable-xen],
          [Support memory introspection with live Xen domains (default is yes)])],
--- a/configure.ac
+++ b/configure.ac
@@ -66,11 +66,6 @@ YES_IS_DEFINED
 dnl -----------------------------------------------
 dnl Check package options
 dnl -----------------------------------------------
- AC_CHECK_HEADER([qemu/libvmi_request.h],
-                 [AC_DEFINE([HAVE_LIBVMI_REQUEST], 1,
-                    [Define to 1 if you have <stdio.h>.])],
-                 [])
-
 AC_ARG_ENABLE([xen],
       [AS_HELP_STRING([--disable-xen],
          [Support memory introspection with live Xen domains (default is yes)])],
@@ -222,6 +217,10 @@ AC_SUBST([GLIB_LIBS])
     AC_CHECK_HEADER(libvirt/libvirt.h, [],
         [AC_MSG_ERROR([No libvirt headers found. Install missing package or re-run with --disable-kvm])])
     AC_DEFINE([ENABLE_KVM], [1], [Define to 1 to enable KVM support.])
+    AC_CHECK_HEADER([qemu/libvmi_request.h],
+        [AC_DEFINE([HAVE_LIBVMI_REQUEST], 1,
+           [Define to 1 if you have <qemu/libvmi_request.h>.])],
+        [])
 [fi]
 
 [if test "$enable_shm_snapshot" = "yes"]
--- a/libvmi/driver/kvm/kvm.c
+++ b/libvmi/driver/kvm/kvm.c
@@ -51,12 +51,18 @@
 
 #define QMP_CMD_LENGTH 256
 
+#ifdef HAVE_LIBVMI_REQUEST
+# include <qemu/libvmi_request.h>
+#else
+
 // request struct matches a definition in qemu source code
 struct request {
     uint8_t type;   // 0 quit, 1 read, 2 write, ... rest reserved
     uint64_t address;   // address to read from OR write to
     uint64_t length;    // number of bytes to read OR write
 };
+
+#endif
 
 enum segment_type {
   SEGMENT_SELECTOR,
--- a/libvmi/driver/kvm/kvm.c
+++ b/libvmi/driver/kvm/kvm.c
@@ -42,6 +42,7 @@
 #include <glib/gstdio.h>
 #include <libvirt/libvirt.h>
 #include <libvirt/virterror.h>
+#include <json-c/json.h>
 
 #include "private.h"
 #include "driver/driver_wrapper.h"
@@ -137,6 +138,19 @@ exec_info_registers(
         "'{\"execute\": \"human-monitor-command\", \"arguments\": {\"command-line\": \"info registers\"}}'";
     return exec_qmp_cmd(kvm, query);
 }
+
+static struct json_object *
+exec_info_version(
+    kvm_instance_t *kvm)
+{
+    char *query =
+        "'{\"execute\": \"query-version\"}'";
+    char *output = exec_qmp_cmd(kvm, query);
+    struct json_object *jobj = json_tokener_parse(output);
+    free(output);
+    return jobj;
+}
+
 
 static char *
 exec_memory_access(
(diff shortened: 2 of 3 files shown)


tools/qemu-kvm-patch/qemu-2.8-memaccess.patch
```diff
@@ -0,0 +1,362 @@
...
++# Returns: Nothing on success
++#
++# Since: 2.4.0.1
++#
++# Notes: Derived from previously existing patches. When command
++# succeeds connect to the open socket. Write a binary structure to
++# the socket as:
++#
++# struct request {
++#     uint8_t type;   // 0 quit, 1 read, 2 write, ... rest reserved
```
tklengyel: So since this is an ABI I think it would be much better if we changed this to a uint64_t as well as to avoid various compilers deciding on different padding after it. I realize we have been using this in all other patches too but at least going forward we should be safer.

Wenzel: good idea. i just pushed the fix, but what about using `__attribute__((packed)) ` ?
Does all compilers use the same alignement strategy when there is this attribute specified ?

tklengyel: `__attribute__((packed))` is a GCC specific C extension, so other compilers may not honor it

Wenzel: At least, it works with Clang also.
But let's go for replacing with `uint64_t`

tklengyel:
```
/root/qemu-2.8.0/memory-access.c: In function ‘connection_handler’:
/root/qemu-2.8.0/memory-access.c:135:13: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘uint64_t’ [-Wformat=]
             fprintf(stderr, "Qemu pmemaccess: ignoring unknown command (%d)\n", req.type);
             ^
  CC      aarch64-softmmu/qtest.o
  CC      alpha-softmmu/memory-access.o
/root/qemu-2.8.0/memory-access.c: In function ‘connection_handler’:
/root/qemu-2.8.0/memory-access.c:135:13: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘uint64_t’ [-Wformat=]
             fprintf(stderr, "Qemu pmemaccess: ignoring unknown command (%d)\n", req.type);
             ^
  CC      alpha-softmmu/qtest.o

**Comment by tklengyel** [comment]
This is ok to test.

**Comment by tklengyel** [comment]
The `struct request` definition in LibVMI driver/kvm/kvm.c is now out-of-sync with what is being defined in this patch. IMHO it would be best if the structure wouldn't be defined in an exported header that LibVMI can include and only define it locally if no other definition was found.

**Comment by Wenzel** [comment]
So `LibVMI` should install that header ?
Or the modified `QEMU` should do it ?

**Comment by tklengyel** [comment]
The modified QEMU should export that header.

File: configure.ac [hunk 2]
```diff
@@ -66,6 +66,11 @@ YES_IS_DEFINED
 dnl -----------------------------------------------
 dnl Check package options
 dnl -----------------------------------------------
+ AC_CHECK_HEADER([qemu/libvmi_request.h],
```
tklengyel: This check should be moved below to the `[if test "$enable_kvm" = "yes"]` block.

Wenzel:
I'm more a fan of `CMake` vs `autotools`, so i copy pasted this `AC_CHECK_HEADER` from stackoverflow :)

If there is a more elegant way of writing it, and better, if you want to display a hint about `libvmi_request.h` in the `configure` summary, go ahead !

tklengyel:
The check itself is fine, it's position is what is still not good. There is a block of checks that only get performed if KVM is enabled. This check should be in the same block.

configure.ac
```diff
@@ -217,6 +217,10 @@ AC_SUBST([GLIB_LIBS])
     AC_CHECK_HEADER(libvirt/libvirt.h, [],
         [AC_MSG_ERROR([No libvirt headers found. Install missing package or re-run with --disable-kvm])])
     AC_DEFINE([ENABLE_KVM], [1], [Define to 1 to enable KVM support.])
+    AC_CHECK_HEADER([qemu/libvmi_request.h],
```
tklengyel: Almost there, just has to move it before the AC_DEFINE line =)

tklengyel:
Retest this please.

Wenzel:
i just rebased it into one commit.

Wenzel:
Can you explain what caused the test fail for `PR_LibVMI_KVM_Test_Ubuntu1204-64` ?

tklengyel:
Retest this please.

tklengyel:
So I've updated the KVM test machine to use QEMU 2.8 with this patch. It spins at 100% and then times out.

```                                                          
   803 libvirt+  20   0 2274452 580436  21720 R  99.5 28.8   5:03.03 qemu-system-x86

**Comment by tklengyel** [comment]
Retest this please.

**Comment by tklengyel** [comment]
Retest this please.

**Comment by tklengyel** [comment]
OK, so the problem had been that QEMU 2.8 is installed in `/opt`, thus ./configure wasn't finding `qemu/libvmi_request.h` properly. LibVMI then was silently compiled with the legacy struct definition, which of course hung the VM. To avoid this issue it might be worthwhile to explore adding an extra check (compile- or run-time) to verify that if we are running on QEMU 2.8+ that this struct was defined from the header.

**Comment by Wenzel** [comment]
How do you propose to implement that check ?
Maybe adding the hint in the configure summary would be the most simple thing to do here.

**Comment by tklengyel** [comment]
Doing it in configure is not necessarily the right place because you could compile LibVMI on one machine but run it on another. Hence a runtime check would be more robust. Though I'm not sure how that would be done.

**Comment by Wenzel** [comment]
What about adding a version number at the end of the `request` struct ?

Therefore we can check at runtime when we try to initalize the `KVM` driver.
On a standard QEMU `2.8`, the packet sent on the socket will be too small, and the `read(kvm_get_instance(vmi)->socket_fd, buf, length + 1);` will return an error.

On a QEMU `2.8+`, we can compare the version number against a defined constant in `libvmi_request.h`

**Comment by tklengyel** [comment]
Adding versioning to the struct is a good idea but relying on the out-of-bound read to fail is not a nice way to signal a version-mismatch. There really is no existing API to get the qemu version at runtime?

**Comment by Wenzel** [comment]
Ah, sorry, i didn't really understood when you mentionned QEMU `2.8+`.
So yes, there is the montior command `info` with the subcommand `version`
~~~
(qemu) info version
2.6.2 (qemu-2.6.2-7.fc24)
~~~

I can append some extra info there, and we could parse it when initalizing the driver.

**Comment by Wenzel** [comment]
So, the QEMU version is defined in the `VERSION` file, at the root of the repo :
`VERSION`
~~~
2.8.0
~~~

I tried to append `-vmi` here, however, it is not parsed as a string internally, but as 3 integers(https://github.com/qemu/qemu/blob/master/qmp.c#L53).

So as i didn't wanted to change a lot of things about version parsing inside QEMU, i just modified the `hmp_version` [function to append `vmi` for now](https://github.com/KVM-VMI/qemu/blob/qemu-v2.8-pmemaccess/hmp.c#L73)
~~~C
void hmp_info_version(Monitor *mon, const QDict *qdict)
{
    VersionInfo *info;

    info = qmp_query_version(NULL);

    monitor_printf(mon, "%" PRId64 ".%" PRId64 ".%" PRId64 "%s %s\n",
                   info->qemu->major, info->qemu->minor, info->qemu->micro,
                   info->package,
                   "vmi");

    qapi_free_VersionInfo(info);
}
~~~

On the libvmi KVM driver side, i added this runtime check in `kvm_init_vmi`
~~~C
#ifdef HAVE_LIBVMI_REQUEST
    char *version = exec_info_version(kvm);
    dbprint(VMI_DEBUG_KVM, "Checking QEMU version string : %s", version);
    // find substring "vmi"
    if (NULL == strstr(version, "vmi"))
    {
        dbprint(VMI_DEBUG_KVM, "Failed find 'vmi' in QEMU version\n");
        return VMI_FAILURE;
    }
#endif
~~~

I tested with `./vmi-process-list win7x64` in `examples`, it works.

Here is the debug output:
~~~
--qmp: virsh -c qemu:///system qemu-monitor-command nitro_win7x64 '{"execute": "human-monitor-command", "arguments": {"command-line": "info version"}}'
Checking QEMU version string : {"return":"2.8.0 vmi\r\n","id":"libvirt-40"}
~~~

Tell me what you think.
- should i move the runtime check somewhere else ?
- should i add more debug statements ?

File: libvmi/driver/kvm/kvm.c [hunk 4]
```diff
@@ -1364,6 +1380,17 @@ kvm_init_vmi(
     }
     vmi->num_vcpus = info.nrVirtCpu;
 
+#ifdef HAVE_LIBVMI_REQUEST
```
tklengyel: The whole point of checking the QEMU version is to determine if not having the request structure defined by a header is a problem or not. So right now you have it backwards. If the pre-processor check HAVE_LIBVMI_REQUEST succeeds we know we are 2.8+ and have the libvmi header. What we need is to check the runtime version and if it is 2.8+ and we don't have HAVE_LIBVMI_REQUEST then print an error message and exit.

Wenzel:
I reversed the runtime check, but for me this looks like the same at the end =)

libvmi/driver/kvm/kvm.c
```diff
@@ -1364,6 +1380,17 @@ kvm_init_vmi(
     }
     vmi->num_vcpus = info.nrVirtCpu;
 
+#ifndef HAVE_LIBVMI_REQUEST
+    char *version = exec_info_version(kvm);
+    dbprint(VMI_DEBUG_KVM, "--Checking QEMU version string : %s", version);
+    // find substring "vmi"
```
tklengyel: So there is no point appending "vmi" to the version string. The version should be checked whether it is 2.8 or later and if so, error out.

Wenzel: Ok, so i will remove `vmi` to the QEMU version string.
And to check that the QEMU version is `2.8.0` or above, we need to parse it.

As i would like to avoid manually parsing a string in C, because it's error prone and painful, i would rather use the QMP command `query-version` and load the json output:
~~~JSON
{"return":{"qemu":{"micro":0,"minor":8,"major":2},"package":""},"id":"libvirt-35"}
~~~
which will allow me to have the `major`, `minor` and `micro` int version number directly.
This would add a dependency on a JSON library.

Do we already use a JSON parser in libvmi ?
What do you think of this solution ?

tklengyel: That works for me. There is a json-c dependency for Rekall profiles already, so adding that as a KVM dependency too would be fine.

Wenzel: I may have noticed an issue in where we are doing the check.

the KVM driver has 2 memory accesses:
- the slow one, using QEMU commands
- the fast one, using the socket and the modified QEMU

Here, we are always failing if we detect that QEMU is `v2.8.0` or later.
So we deny the possibility of using the slow memory access.

Is it what you want, and focus only on the modified QEMU ?
Did i misunderstood something ?

Wenzel:
Any update ? :)

configure.ac
tklengyel: Ping.

tklengyel:
Retest this please.

Wenzel:
Hi @tklengyel ,

Just to know the status of this PR, are you waiting for me to make some changes, or is it OK for your review ?

Thanks :)

Wenzel:
The test `PR_LibVMI_KVM_Test_Winxpsp2` fails because it cannot find the C JSON library

~~~
libvmi/include -MT driver/kvm/libvmi_la-kvm.lo -MD -MP -MF driver/kvm/.deps/libvmi_la-kvm.Tpo -c driver/kvm/kvm.c  -fPIC -DPIC -o driver/kvm/.libs/libvmi_la-kvm.o
driver/kvm/kvm.c:45:25: fatal error: json-c/json.h: No such file or directory
 #include <json-c/json.h>
                         ^
compilation terminated.
~~~

tklengyel:
Retest this please.

tklengyel:
@Wenzel Looks good now! Thanks =)

tklengyel:
Retest this please.

Issue #460, opened 2017-03-13
kvm vmi-process-list can‘t work for win7-64 well, it shows:*** buffer overflow detected ***:


Badangel:
It works well for linux, but when I use it for win7.  I have got win7 offset use this:
   g++ -o getGUID getGUID.cpp
vmi-dump-memory windows7 win7ult.dd      
./getGUID win7ult.dd |python ./downloadPDB.py |python ./dumpPDB.py -o debugSymbols.txt
./createConfig.py -f debugSymbols.txt 
edit libvmi.conf:
win7ult {
    ostype = "Windows";
    win_tasks = 0xb8;
    win_pdbase = 0x18;
    win_pname = 0x16c;
    win_pid = 0xb4;
}

It shows me error like this:

```
root@dyf:/home/dyf/Downloads/libvmi-master/examples# vmi-process-list win7ult
*** buffer overflow detected ***: vmi-process-list terminated
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x77725)[0x7fa4729d7725]
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x5c)[0x7fa472a7889c]
/lib/x86_64-linux-gnu/libc.so.6(+0x1168a0)[0x7fa472a768a0]
/lib/x86_64-linux-gnu/libc.so.6(__strcat_chk+0x5d)[0x7fa472a75bad]
/usr/local/lib/libvmi-0.11.so(+0x60ef)[0x7fa472d2f0ef]
/usr/local/lib/libvmi-0.11.so(+0x6bd1)[0x7fa472d2fbd1]
/usr/local/lib/libvmi-0.11.so(+0x96c8)[0x7fa472d326c8]
/usr/local/lib/libvmi-0.11.so(+0x9e71)[0x7fa472d32e71]
/usr/local/lib/libvmi-0.11.so(+0x7e10)[0x7fa472d30e10]
/usr/local/lib/libvmi-0.11.so(+0xe7aa)[0x7fa472d377aa]
vmi-process-list[0x400ba6]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7fa472980830]
vmi-process-list[0x400e89]
======= Memory map: ========
00400000-00402000 r-xp 00000000 08:02 58467416                           /usr/local/bin/vmi-process-list
00601000-00602000 r--p 00001000 08:02 58467416                           /usr/local/bin/vmi-process-list
[...]

PR #461, opened 2017-03-14
Fix overflow in KVM segment register parsing

masthoon:
Should fix issue #460 

Null byte off-by-one in KVM driver.


tklengyel:
This is OK to test.

PR #462, opened 2017-03-15
Fix vmifs double-declaring vmi_instance_t

tklengyel:


mfusaro:
+1 this fixes the issue I was having.

PR #467, opened 2017-04-04
Don't include hvm/hvm_op.h in Xen events

tklengyel:
This header had been included for the HVMOP_TRAP_* definitions. The header is being renamed in Xen 4.9 so instead of doing configure checks to see which header is available we just define the required constants internally.


tklengyel:
Retest this please.

tklengyel:
Retest this please.

Issue #464, opened 2017-03-24
Possible inconsistent memory access using libvmi cache


Wenzel:
Hi,

I'm using `libvmi` to decode `syscall` parameters while intercepting them in `KVM`.

For example, `NtOpenKey`, on Windows :
[NtOpenKey on MSDN](https://msdn.microsoft.com/en-us/library/windows/hardware/ff567014(v=vs.85).aspx)
~~~C
NTSTATUS ZwOpenKey(
  _Out_ PHANDLE            KeyHandle,
  _In_  ACCESS_MASK        DesiredAccess,
  _In_  POBJECT_ATTRIBUTES ObjectAttributes
);
~~~

I'm decoding the `ObjectAttributes` paramters, since it contains interesting value :
[OBJECT_ATTRIBUTES on MSDN](https://msdn.microsoft.com/en-us/library/windows/hardware/ff557749(v=vs.85).aspx)
~~~C
typedef struct _OBJECT_ATTRIBUTES {
  ULONG           Length;
  HANDLE          RootDirectory;
  PUNICODE_STRING ObjectName;
  ULONG           Attributes;
  PVOID           SecurityDescriptor;
  PVOID           SecurityQualityOfService;
}  OBJECT_ATTRIBUTES, *POBJECT_ATTRIBUTES;
~~~

I'm using the `vmi_read_va` function to read each field value.

**However, i noticed that sometimes the memory content returned is inconsistent and makes no sense.**

if you take a look at `NtDef.h`, where this structure is declared, there is a macro which initializes this field :

~~~C
#define InitializeObjectAttributes(p,n,a,r,s) { \
  (p)->Length = sizeof(OBJECT_ATTRIBUTES); \
  (p)->RootDirectory = (r); \
  (p)->Attributes = (a); \
  (p)->ObjectName = (n); \
  (p)->SecurityDescriptor = (s); \
  (p)->SecurityQualityOfService = NULL; \
}
~~~

Therefore, the `Length` field here is fixed and must be `0x30`.

So, most of the times, it works well :
~~~
Building ObjectAttributes from 0x188f1e0
Field Length, I, at 0x188f1e0 + 0x0
Value: 0x30
Field Handle, P, at 0x188f1e0 + 0x8
Value: 0x0
Field PUnicodeString, P, at 0x188f1e0 + 0x10
Value: 0x188f1c0
Building UnicodeString from 0x188f1c0
Field Length, H, at 0x188f1c0 + 0x0
Value: 0x62
[...]

PR #469, opened 2017-04-06
add new API to flush the internal page cache

Wenzel:
A few weeks ago, i opened an issue on libvmi : #464 

The solution was to flush the internal page cache, which is not accessible from any exposed API.

This PR exposes an API to flush the cache when the user thinks the memory layout has changed.

It worked for me, but maybe it will need more polishing from your side.
give me your feedback :)

### DIFF
--- a/libvmi/cache.c
+++ b/libvmi/cache.c
@@ -37,6 +37,7 @@
 
 #include "private.h"
 #include "glib_compat.h"
+#include "driver/memory_cache.h"
 
 // This function borrowed from cityhash-1.0.3
 uint64_t hash128to64(
@@ -663,6 +664,13 @@ v2p_cache_flush(
     dbprint(VMI_DEBUG_V2PCACHE, "--V2P cache flushed\n");
 }
 
+void
+page_cache_flush(
+    vmi_instance_t vmi)
+{
+    memory_cache_flush(vmi);
+}
+
 // Below are wrapper functions for external API access to the cache
 void
 vmi_pidcache_add(
@@ -732,4 +740,11 @@ vmi_v2pcache_flush(
     addr_t dtb)
 {
     return v2p_cache_flush(vmi, dtb);
+}
+
+void
+vmi_pagecache_flush(
+    vmi_instance_t vmi)
+{
+    return page_cache_flush(vmi);
 }
--- a/libvmi/cache.h
+++ b/libvmi/cache.h
@@ -68,6 +68,8 @@ void v2p_cache_flush(vmi_instance_t vmi, addr_t dtb);
 status_t v2p_cache_get(vmi_instance_t vmi, addr_t va, addr_t dtb, addr_t *pa);
 status_t v2p_cache_del(vmi_instance_t vmi, addr_t va, addr_t dtb);
 
+void page_cache_flush(vmi_instance_t vmi);
+
 #else
 
 #define pid_cache_init(...)     NOOP
--- a/libvmi/driver/memory_cache.c
+++ b/libvmi/driver/memory_cache.c
@@ -78,7 +78,7 @@ static void
 clean_cache(
     vmi_instance_t vmi)
 {
-    while (g_queue_get_length(vmi->memory_cache_lru) > vmi->memory_cache_size_max / 2) {
+    while (g_queue_get_length(vmi->memory_cache_lru)) {
         gint64 *paddr = g_queue_pop_tail(vmi->memory_cache_lru);
 
         g_hash_table_remove(vmi->memory_cache, paddr);
@@ -261,6 +261,13 @@ memory_cache_destroy(
     release_data_callback = NULL;
 }
 
+void
+memory_cache_flush(
+        vmi_instance_t vmi)
+{
+    clean_cache(vmi);
+}
+
 #else
 void
 memory_cache_init(
@@ -314,4 +321,16 @@ memory_cache_destroy(
     get_data_callback = NULL;
     release_data_callback = NULL;
 }
+
+void
+memory_cache_flush(
+        vmi_instance_t vmi)
+{
+    if(vmi->last_used_page_key && vmi->last_used_page) {
+        release_data_callback(vmi->last_used_page, vmi->page_size);
+    }
+    vmi->last_used_page_key = 0;
+    vmi->last_used_page = NULL;
+}
+
 #endif
--- a/libvmi/driver/memory_cache.h
+++ b/libvmi/driver/memory_cache.h
@@ -49,4 +49,7 @@ void memory_cache_remove(
 void memory_cache_destroy(
     vmi_instance_t vmi);
 
+void memory_cache_flush(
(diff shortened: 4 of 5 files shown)


awsaba:
this is ok to test

smaresca:
@Wenzel the failed due to:
> ../libvmi/check_libvmi-cache.o: In function `page_cache_flush':
> /home/jenkins/workspace/PR_LibVMI_KVM_Test_Winxpsp2/tests/../libvmi/cache.c:670: undefined reference to `memory_cache_flush'
> /home/jenkins/workspace/PR_LibVMI_KVM_Test_Winxpsp2/tests/../libvmi/cache.c:670: undefined reference to `memory_cache_flush'
> collect2: error: ld returned 1 exit status
> Makefile:590: recipe for target 'check_libvmi' failed

libvmi/cache.c
```diff
@@ -663,6 +664,13 @@ v2p_cache_flush(
     dbprint(VMI_DEBUG_V2PCACHE, "--V2P cache flushed\n");
 }
 
+void
+page_cache_flush(
```
tklengyel: What's the point of having this extra function?

libvmi/cache.c
```diff
@@ -733,3 +741,10 @@ vmi_v2pcache_flush(
 {
     return v2p_cache_flush(vmi, dtb);
 }
+
+void
+vmi_pagecache_flush(
+    vmi_instance_t vmi)
+{
+    return page_cache_flush(vmi);
```
tklengyel: You could just as well call `memory_cache_flush` here.

tklengyel:
Evidently the KVM driver is lacking in comparison to Xen as what happens is that LibVMI reads the whole page into an internal buffer and caches it. Obviously this buffer will go out-of-sync sooner or later if the VM is running. So unless the VM is completely paused for the entire duration of the LibVMI instance, we can't guarantee that the memory stored in the cache is correct.

So this API addition doesn't really solve the problem and will not be evident that the user has to do the following:
1. pause vm
2. flush page cache
3. read/write
4. resume vm

If you don't pause the entire VM then the memory could go out-of-sync at any point. Furthermore, this is obviously a brute way of ensuring sanity as we don't know if we really need to flush the page or not (it might not have changed after all). But since the cache would have to be flushed every time you pause, then having the cache itself doesn't really get you much. You could just as well disable it and read the entire page in for yourself in your application when you pause the VM.

A much much better way would be to do away with reading in and caching the entire page to begin with. If we could get the KVM driver to provide us with a handle to the page like Xen does then we wouldn't have to worry about the buffer going out-of-sync and accessing it would also be faster (ie. zero-copy).

Bottom line, the current KVM memory access patches were not designed with running VMs in mind and this patch only provides a workaround that will be far from obvious and/or optimal.

Wenzel:
This is how i'm currently using Libvmi with KVM:
1. run the vm
2. wait until the next event
3. event received, vm is paused
4. read/write the memory, decode data
5. continue the vm execution

I could have disabled the internal page cache with `./configure --disable-page-cache`, and it works ofc.

> But since the cache would have to be flushed every time you pause, then having the cache itself doesn't really get you much.

Since i'm doing a lot of read/write operations on the same page while processing the event, i need it to be cached by Libvmi to keep a fast memory access.

> You could just as well disable it and read the entire page in for yourself in your application when you pause the VM.

I prefer to keep these low level operations on the Libvmi level.

> If we could get the KVM driver to provide us with a handle to the page like Xen does then we wouldn't have to worry about the buffer going out-of-sync and accessing it would also be faster (ie. zero-copy).

What do you mean by Xen providing you a `handle` to the page ?
What operations can you do with it ?

Also, what about using EPT to monitor the pages that have been written during the vm execution ?
> When accessed and dirty flags for EPT are enabled, software can track writes to guest-physical addresses using a feature called `page-modification logging`

It could be a solution for the new design of the Libvmi KVM driver.

tklengyel:
@Wenzel: the handle the Xen driver gives you is akin to a file handler. Effectively Xen maps the memory page from the target VM into the memory space of the process that uses LibVMI. That way the same physical memory is visible in both without having to pause and copy the page to a buffer. This would be the ideal setup for KVM as well.

Doing tricks with EPT is possible, ie. notify LibVMI when one of the pages in the cache have gone dirty, but IMHO that adds complexity. It can also leave the VM in an unstable state in case the process crashes and LibVMI's cleanup isn't called. It also adds a lot of overhead of trapping out every time the page changes when we might not even need the page in the cache anymore.

awsaba:
I missed the part of the original issue that this was KVM specific, so now I'm kind of torn about whether this should make it into the external API.

RE: EPT tricks, the reread of the page could be limited to only happening on the first read/write after the violation, but as @tklengyel states, that is a lot of complexity that would probably not make it into libvmi when the alternatives seem to be:

- Fix up the QEMU memaccess patch
- Use Xen

noxdafox:
@Wenzel maybe worth creating an issue on [KVM-VMI/kvm](https://github.com/KVM-VMI/kvm) describing the memory mapping feature @tklengyel is mentioning?

awsaba:
@noxdafox I don't have links to the discussions handy, but attempts have been made to mainline something like [this patch](tools/qemu-kvm-patch/kvm-qemu-2.4.0.1-libvmi.patch) have been made before and haven't made it.

Wenzel:
Designing a new approach to allow KVM to give a handle on a memory page like Xen does would be the way to go here.
However, this could easily take months or more, depending on how KVM manipulates the memory under the hood, and how to get an access to it without crashing the guest. 

In the meantime, would it be possible to still have the memory cache flushed ?

Like you said above, there is no need to add a new API, since we can flush when the user wants to flush the `v2p` cache:

~~~C
void
vmi_v2pcache_flush(
    vmi_instance_t vmi,
    addr_t dtb)
{
    memory_cache_flush(vmi);
    return v2p_cache_flush(vmi, dtb);
}
~~~

What does it changes on the Xen side ?
How do you know that the internal page cache needs to be flushed for Xen ?

tklengyel:
On Xen the internal page-cache rarely needs to get flushed. The only consideration there is how many handles do you want to keep open at any given time as they don't go stale. So it's only a memory use issue as you don't want to have hundreds of page handles active all the time.

Flushing it as part of v2p flush doesn't really make much sense, I would prefer to have a standalone API for this. As I said, I'm not totally against this API addition, it is just a half-measure that doesn't really solve the problem.

tklengyel:
So a quick read of https://lwn.net/Articles/658511 tells me what we should do is have QEMU share its `vmfd` with LibVMI and then have the KVM kernel module support an ioctl for `KVM_GET_USER_MEMORY_REGION`, similarly to the existing `KVM_SET_USER_MEMORY_REGION`. That would allow us to get the memory mapped into the LibVMI process without having to resort to actually copying the memory via qmp.

tklengyel:
I think it is safe to close this PR for now, @Wenzel?

PR #495, opened 2017-07-06
Rework set listener access required internals

tklengyel:


tklengyel:
I'll be merging this patch in a week if there are no objections.

PR #510, opened 2017-07-21
Add support for new events from Xen 4.8 - Xen 4.10

tklengyel:


smaresca:
@tklengyel would like to see an update to notes/events.txt for these

tklengyel:
@smaresca: that document is woefully out-of-date so I would recommend doing an update to that in a separate PR

smaresca:
@tklengyel yes please. It's important that it get an update.

PR #514, opened 2017-07-27
Fix process-list example so as to not print out trash value

### LINKED ISSUE #513: process-list example print a trash value for last record
When I run process-list example in libvmi/examples, the output always has  a trash value at the last record like this : 
[ 1004] svchost.exe (struct addr:fffffa800318a5f0)
[ 1844] taskmgr.exe (struct addr:fffffa800217d060)
[52581608] ������������ (struct addr:fffff80002c44a08)

dongphuong2410:
This commit fixes #513 .
Currently, the process list printed out by running ./vmi-process-list  example always contains  wrong data at the last line.
It happens because the 'Next' pointer in last EPROCESS in the list is the address of the PsActiveProcessHead symbol, not the address of the list element in the first EPROCESS. => print out EPROCESS fields here makes that trash values.

### DIFF
--- a/examples/process-list.c
+++ b/examples/process-list.c
@@ -36,7 +36,7 @@
 int main (int argc, char **argv)
 {
     vmi_instance_t vmi;
-    addr_t list_head = 0, next_list_entry = 0;
+    addr_t list_head = 0, cur_list_entry = 0, next_list_entry = 0;
     addr_t current_process = 0;
     char *procname = NULL;
     vmi_pid_t pid = 0;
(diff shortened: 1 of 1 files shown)


cuiyudaniuHT:
We will try it later.

examples/process-list.c
```diff
@@ -128,10 +128,23 @@ int main (int argc, char **argv)
...
+        /* follow the next pointer */
+        status = vmi_read_addr_va(vmi, next_list_entry, 0, &next_list_entry);
+        if (status == VMI_FAILURE) {
+            printf("Failed to read next pointer in loop at %"PRIx64"\n", next_list_entry);
+            goto error_exit;
+        }
+        /* If the next pointer points to the head of list, this pointer is actually the
+         * address of PsActiveProcessHead symbol, not the address of an ActiveProcessLink in
+         * EPROCESS struct. We should escape from the loop here. */
+        if (next_list_entry == list_head) {
```
tklengyel: This is only needed if the VM is Windows. For Linux even the last entry in the list should get printed.

dongphuong2410:
A fixing for Changes requested has been committed.

tklengyel:
This is OK to test.

PR #516, opened 2017-07-28
Standardize API to return status_t at most places

### LINKED ISSUE #248: Allow `vmi_pagetable_lookup_extended` to return 0 as a valid physical address
`vmi_pagetable_lookup_extended` should return VMI_SUCCESS and `info->paddr = 0` if the virtual address really does resolve to physical address 0.  This is very unlikely, but possible.

This change will probably involve modifications to the arch `v2p` functions, since whether or not this condition is an error can be detected there and by the driver if 0 is not a valid physical address.

tklengyel:
Fixes #248

### DIFF
--- a/examples/dump-memory.c
+++ b/examples/dump-memory.c
@@ -80,7 +80,7 @@ main(
     while (address < size) {
 
         /* write memory to file */
-        if (PAGE_SIZE == vmi_read_pa(vmi, address, memory, PAGE_SIZE)) {
+        if (VMI_SUCCESS == vmi_read_pa(vmi, address, PAGE_SIZE, memory, NULL)) {
             /* memory mapped, just write to file */
             size_t written = fwrite(memory, 1, PAGE_SIZE, f);
 
--- a/examples/event-example.c
+++ b/examples/event-example.c
@@ -140,7 +140,8 @@ event_response_t syscall_lm_cb(vmi_instance_t vmi, vmi_event_t *event){
 
 event_response_t cr3_one_task_callback(vmi_instance_t vmi, vmi_event_t *event){
 
-    vmi_pid_t pid = vmi_dtb_to_pid(vmi, event->reg_event.value);
+    vmi_pid_t pid = -1;
+    vmi_dtb_to_pid(vmi, event->reg_event.value, &pid);
 
     printf("one_task callback\n");
     if(event->reg_event.value == cr3){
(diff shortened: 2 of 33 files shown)


tklengyel:
Retest this please.

tklengyel:
Retest this please.

tklengyel:
Retest this please.

smaresca:
Please check the new output parameters to ensure they're non-NULL pointers. Example: vmi_get_offset should test for a non-null 'offset' parameter and return VMI_FAILURE. vmi_translate_ksym2v, vmi_translate_sym2v, vmi_dtb_to_pid, has a similar need, etc

smaresca:
Functions get_va_pages_ia32e,  get_va_pages_nopae, and get_va_pages_pae have inconsistent behavior with respect to  page_info_t allocation failures during a loop iterations (return vs continue) I think get_va_pages_nopae will leak pt_page/pgd_page due to the in-loop return for those alloc failures as well.

Regarding continue vs return, I'm not sure which I prefer -- continue for partial results (perhaps with a missing page in the list) or return for truncated results at the first failure.

tklengyel:
The get_va_page is a separate issue that will not be addressed in this PR. Please open an issue for it to be tracked.

smaresca:
@tklengyel no, you are introducing the get_va_page_* behavior inconsistency in this PR. See amd64.c line 279, 334, intel.c line 433

tklengyel:
> I think get_va_pages_nopae will leak pt_page/pgd_page due to the in-loop return for those alloc failures as well.

There is no leak.

tklengyel:
> Please check the new output parameters to ensure they're non-NULL pointers.

There are many many exposed functions through the API that do not check for NULL. If you would like that everything is NULL checked, that should be part of a separate PR.

tklengyel:
> Regarding continue vs return, I'm not sure which I prefer

It doesn't really matter. What I wanted to avoid there is a segfault which gets accomplished by both a `return` and by a `continue`. If you are that low on memory and you continue you will likely return the same list as if you just returned at that point. And you will likely going to crash soon too because, well, you are out of memory. But at least it won't be a segfault in LibVMI.

smaresca:
> > Regarding continue vs return, I'm not sure which I prefer

> It doesn't really matter. 
Agree, either path was fine. I just wanted them to be the same across that function family. Thank you for updating that component.

Issue #302, opened 2015-12-09
Change references to vmi->size to correctly use max_physical_ram_size and max_allocated_ram as appropriate


awsaba:
A few references to vmi->size were left in the experimental snaphshot build.

Work needs to be done to really figure out how that code should be using max_physical_ram_size vs. max_allocated_ram.  My intuition is that if it didn't make a distinction before and just used `size`, it probably didn't work correctly.

awsaba:
@bahaihe The only way to determine the correct way to incorporate `max_physical_address` and `allocated_ram_size` will be to thoroughly understand the snapshot code.  I do not think the fix will be as simple as changing the references to `vmi->size` to the new fields.

As mentioned in the comment above, I don't think the original authors of the snapshot code took into account the unmapped physical pages within the guests physical address range.  The snapshot code should fail/return null for unmapped pages from a snapshot the same way that any of the normal `vmi_read()` functions would.

Some ranges to start with would be:
 * 0xa0000-0xafffff, which is the legacy memory hole
 * 000ff000000-000ffffffff on a VM with 4GB of RAM allocated

You can refer to [this gist](https://gist.github.com/awsaba/5c807a8d07582c0f5770) for more examples

PR #517, opened 2017-08-01
Remove broken Xen SHM code

tklengyel:
The SHM code for Xen has been broken for a while and is without a use-case as it provides very little value compared to the KVM SHM implementation.


tklengyel:
Fixes #302

PR #522, opened 2017-08-14
minor fixes to the enable-shm-snapshot build option

ftPeter:
SHM-snapshot specific code had gotten crusty, so I got it compiling again. Just updated some variable names and parameter lists.

### DIFF
--- a/libvmi/driver/kvm/kvm.c
+++ b/libvmi/driver/kvm/kvm.c
@@ -867,7 +867,6 @@ create_v2m_table(
         }
         else {
             // user process page table
-            dtb = vmi->kpgd;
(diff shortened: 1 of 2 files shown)


tklengyel:
This is OK to test.

libvmi/driver/kvm/kvm.c
```diff
@@ -867,8 +867,8 @@ create_v2m_table(
         }
         else {
             // user process page table
-            dtb = vmi_pid_to_dtb(vmi, pid);
-            if (!dtb) {
+            dtb = vmi->kpgd;
```
tklengyel: There is no point assigning `kpgd` here

ftPeter: Good catch, I'll review that and update the PR in the next day or so.

PR #530, opened 2017-09-18
add meaningful error messages for pmemaccess QEMU socket

Wenzel:
This patch just adds meaningful error messages when the pmemeaccess socket creation fails on the QEMU side.

Example:

~~~
$ virsh -c qemu:///system qemu-monitor-command nitro_win7x64 '{"execute": "pmemaccess", "arguments": {"path": "/etc/vmi.socket"}}'
{"id":"libvirt-20","error":{"class":"GenericError","desc":"Qemu pmemaccess: bind failed: Permission denied"}}
~~~

### DIFF
--- a/tools/qemu-kvm-patch/kvm-qemu-v2.8-libvmi.patch
+++ b/tools/qemu-kvm-patch/kvm-qemu-v2.8-libvmi.patch
@@ -104,10 +104,10 @@ index 0000000..e4fc87a
 +#endif /* LIBVMI_REQUEST_H */
 diff --git a/memory-access.c b/memory-access.c
 new file mode 100644
-index 0000000..d23b7cf
+index 0000000..09b7cf8
 --- /dev/null
 +++ b/memory-access.c
-@@ -0,0 +1,214 @@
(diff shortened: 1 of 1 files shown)


tools/qemu-kvm-patch/kvm-qemu-v2.8-libvmi.patch
```diff
@@ -352,6 +351,12 @@ index 0000000..21c7137
 +};
 +
 +#endif /* MEMORY_ACCESS_H */
+diff --git a/pixman b/pixman
+--- a/pixman
++++ b/pixman
+@@ -1 +1 @@
+-Subproject commit 87eea99e443b389c978cf37efc52788bf03a0ee0
++Subproject commit 87eea99e443b389c978cf37efc52788bf03a0ee0-dirty
```
tklengyel: This should not be part of the patch.

Wenzel: Sorry for the `pixman`, again :)

PR #533, opened 2017-09-27
add json-c-dev as build dep in debian packaging

Wenzel:
This PR adds a dependency on `json-c-dev` to the debian package.

We introduced this dependency to parse QEMU string version, but forgot to add it also in the control file.


Wenzel:
Debian package: https://packages.debian.org/fr/jessie/libjson-c-dev

debian/control
```diff
@@ -11,7 +11,8 @@ Build-Depends: debhelper (>= 8.0.0),
  bison,
  libglib2.0-dev,
  libvirt-dev,
```
tklengyel: If you are already editing this file might as well remove `libvirt-dev`, it's no longer required to be present for all cases, it gets automatically detected at runtime if it is.

configure.ac
```diff
@@ -218,6 +218,8 @@ AC_SUBST([GLIB_LIBS])
         [AC_DEFINE([HAVE_LIBVMI_REQUEST], 1,
            [Define to 1 if you have <qemu/libvmi_request.h>.])],
         [])
+    AC_CHECK_HEADERS([json-c/json.h], [],
```
tklengyel: `AC_CHECK_HEADER` would be enough

Wenzel:
This PR is ready for another review or merging :+1:

tklengyel:
Thanks, looks good!

Issue #511, opened 2017-07-24
It circles at '--MEMORY cache hit 0x201000' for win10 of process-list  


cuiyudaniuHT:
I tested process-list for win10 which runs out that the process circles at `--MEMORY cache hit 0x201000` and do not stop (libvmi is newly cloned, qemu 2.8 
[1.txt](https://github.com/libvmi/libvmi/files/1168613/1.txt)
with patch).

First, as the README said, I run win-guid which shows :
`    PDB GUID: dda812f4ac284269ab8073d8423801a41                                       
    Kernel filename: ntkrnlmp.pdb                                                     
    Multi-processor without PAE                                                       
--PEParse: magic is 0x20b                                                             
    Signature: 17744.                                                                 
    Machine: 34404.                                                                   
    # of sections: 26.                                                                
    # of symbols: 0.                                                                  
    Timestamp: 1499407595.                                                            
    Characteristics: 34.                                                              
    Optional header size: 240.                                                        
    Optional header type: 0x20b  `

Then, I run rekall commands getting the rekall-profile.json and write to profile.
` win10_x64 {                                                                     
      ostype = "Windows";                                                         
      rekall_profile = "/home/sysmap/win10_x64.json";                             
  } `

At last, I run `./vmi-process-list win10_x64` with debug. 
It always prints after a second :
`--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
--MEMORY cache hit 0x201000
[...]

PR #540, opened 2017-10-22
KVM: Parse registers as unsigned

namidairo:
Using strtoll meant that any value over LONG_MAX would be returned as LONG_MAX. Change these calls to the unsigned version so we don't end up mangling all the higher virtual addresses.


namidairo:
Fixes #511 (Since the IDT was always > 0x7fffffffffffffff)

tklengyel:

tklengyel:
Looks good to me, thanks!

PR #543, opened 2017-10-26
Add Travis CI style and compile tests and format code with astyle

tklengyel:
The first commit in this PR applies automatic style adjustments using astyle with the K&R style. The second commit adds the Travis CI config to automatically check new commits for style conformity and compile tests with gcc and clang.

There are no functional changes in this PR to the code, just stylistic.

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -1,0 +1,6 @@
+indent-switches
+suffix=none
+style=kr
+indent=spaces=4
+keep-one-line-blocks
+pad-header
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,88 @@
+dist: trusty
+language: c
+sudo: false
+
+matrix:
+    include:
+
+#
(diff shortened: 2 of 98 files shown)


awsaba:
:+1:

tklengyel:
Not sure why the Coverity scan is not getting kicked off properly, but it may just be because the PR is coming from a branch outside libvmi/libvmi.

rianquinn:
your setup looks good. I have had this problem in the past, have you tried running coverity scan's build tools manually to see if you can get a build uploaded? It might shed light on the issue.

tklengyel:
@rianquinn it's pretty odd as the same travis syntax worked fine for me on another project, while in here the config says `  "global_env": null,` when it should be `  "global_env": "COVERITY_SCAN_TOKEN=[secure]",`. I'll try a manual build and see if that makes a difference.

PR #544, opened 2017-11-25
Fix in get_kpgd_from_rekall_profile(). Adding RVA to physical address is not reliable, because virtual pages could be mapped into physical memory non-sequentially.

ReallyNothing:
This is just a minor fix in `get_kpgd_from_rekall_profile()` function. The problem appeared, if some virtual pages, that contained `ntoskrnl`, were mapped into physical memory non-sequentially. In this case we cannot just add RVA to physical address, because relative virtual address could be not equal to offset to data in physical memory.

### DIFF
--- a/libvmi/os/windows/core.c
+++ b/libvmi/os/windows/core.c
@@ -639,7 +639,12 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
         if ( VMI_FAILURE == ret )
             return ret;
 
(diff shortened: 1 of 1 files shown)


ReallyNothing:
If everything is good, I will squash these two commits into one before merge into master.

tklengyel:
This is OK to test.

tklengyel:
@ReallyNothing go ahead and squash the commits

ReallyNothing:
Commits have been squashed into a single one.
Please, let me know, if I have to do anything for this pull request.

tklengyel:
Thanks!

ReallyNothing:
Thank you!

PR #550, opened 2017-12-20
add poll to monitor qemu socket and prevent read errors

Wenzel:
This PR uses `poll` in the QEMU memory-access handler to watch the socket and exit in case of errors.


tklengyel:
@libvmi-jenkins Retest this please.

Issue #531, opened 2017-09-21
size of init_flags in vmi_init ?


Wenzel:
Hi,

i just noticed that we have 2 different declarations of the `init_flags`:
*vmi_init*: https://github.com/libvmi/libvmi/blob/master/libvmi/core.c#L492
~~~C
status_t vmi_init(
    vmi_instance_t *vmi,
    vmi_mode_t mode,
    void* domain,
    uint32_t init_flags,
    void *init_data,
    vmi_init_error_t *error)
~~~

*vmi_init_complete*: https://github.com/libvmi/libvmi/blob/master/libvmi/core.c#L731
~~~C
status_t
vmi_init_complete(
    vmi_instance_t *vmi,
    void *domain,
    uint64_t init_flags,
    void *init_data,
    vmi_config_t config_mode,
    void *config,
    vmi_init_error_t *error)
~~~

Is it a mistake ?

tklengyel:
Yes, it is.

Wenzel:
PR has been merged, thanks @tklengyel !

PR #555, opened 2018-01-03
fix vmi_init init_flags type to uint64_t

Wenzel:
This fixes issue https://github.com/libvmi/libvmi/issues/531


tklengyel:

PR #560, opened 2018-01-19
libvmi/driver: Fix wrapper for driver_set_privcall_event

sergej-proskurin:
This fix adjusts the wrapper driver_set_privcall_event so that it calls the
function that was registered in the vmi->driver.set_privcall_event_ptr function
pointer (instead of driver.set_debug_event_ptr).


tklengyel:

PR #561, opened 2018-01-22
Fixed getting of virtual address of Windows kernel. Now it does not r…

ReallyNothing:
The problem happened, because the way, how virtual address of `ntoskrnl` is determined, relied on the position of specific handler in IDT.
Instead of this, Windows kernel could be searched in backward direction, starting from some interrupt routine address. It is more slow, but should be reliable.

### DIFF
--- a/libvmi/os/windows/core.c
+++ b/libvmi/os/windows/core.c
@@ -662,6 +662,280 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
 }
 
 static status_t
+get_interrupt_routine_offset_32(vmi_instance_t vmi, addr_t idt_descriptor_address, addr_t* result)
+{
+    /*
+     * http://wiki.osdev.org/Interrupt_Descriptor_Table#Structure_IA-32
+     * */
+#pragma pack(push)
+#pragma pack(1)
+    struct
+    {
+        uint16_t offset1;   // offset bits 0..15
+        uint16_t selector;  // a code segment selector in GDT or LDT
+        uint8_t zero;       // unused, set to 0
+        uint8_t type_attr;  // type and attributes, see below
+        uint16_t offset2;   // offset bits 16..31
+    } idt_descriptor;
+#pragma pack(pop)
+
+    size_t bytes_readen = 0;
+    if (VMI_FAILURE == vmi_read_va(vmi, idt_descriptor_address, 0, sizeof(idt_descriptor), &idt_descriptor, &bytes_readen) || 
+        bytes_readen != sizeof(idt_descriptor))
+    {
+        return VMI_FAILURE;
+    }
+
+    uint32_t offset = ((uint32_t)idt_descriptor.offset2 << 16) | idt_descriptor.offset1;
+
+    *result = offset;
+    return VMI_SUCCESS;
+}
+
+static status_t
+get_interrupt_routine_offset_64(vmi_instance_t vmi, addr_t idt_descriptor_address, addr_t* result)
+{
+    /*
+     * http://wiki.osdev.org/Interrupt_Descriptor_Table#Structure_AMD64
+     * */
+#pragma pack(push)
+#pragma pack(1)
+    struct 
+    {
+        uint16_t offset1;   // offset bits 0..15
+        uint16_t selector;  // a code segment selector in GDT or LDT
+        uint8_t ist;        // bits 0..2 holds Interrupt Stack Table offset, rest of bits zero.
+        uint8_t type_attr;  // type and attributes
+        uint16_t offset2;   // offset bits 16..31
+        uint32_t offset3;   // offset bits 32..63
+        uint32_t zero;      // reserved
+    } idt_descriptor;
+#pragma pack(pop)
+
+    size_t bytes_readen = 0;
+    if (VMI_FAILURE == vmi_read_va(vmi, idt_descriptor_address, 0, sizeof(idt_descriptor), &idt_descriptor, &bytes_readen) || 
+        bytes_readen != sizeof(idt_descriptor))
+    {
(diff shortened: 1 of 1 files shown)


smaresca:
This is ok to test

tklengyel:
There are some style issues: https://travis-ci.org/libvmi/libvmi/jobs/331973358 and https://travis-ci.org/libvmi/libvmi/jobs/331973357

I'll also try to setup a win10 guest for the test rig so we have some real testing of this.

libvmi/os/windows/core.c
```diff
@@ -661,6 +661,280 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
...
+            continue;
+        }
+
+        // it is a PE-image. Let's get the module name
+        ctx.addr = start_address_va;
+        if (VMI_SUCCESS == peparse_get_export_table(vmi, &ctx, &export_tbl, &et_rva, &et_size))
+        {
+            et_rva = start_address_va + export_tbl.name;
+
+            if (kernel_image_name_max_size != vmi_read_va(vmi, et_rva, 0, kernel_image_name_max_size, buffer, &bytes_readen) ||
```
tklengyel: This is incorrect, `vmi_read_va` returns a `status_t`.

libvmi/os/windows/core.c
```diff
@@ -661,6 +661,280 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
...
+     *    // ...
+     *    // following structure varies depending on page mode
+     * }
+     * 
+     * http://wiki.osdev.org/Interrupt_Descriptor_Table
+     * */
+    uint16_t selector = 0;
+
+    size_t bytes_readen = 0;
+    if (VMI_FAILURE == vmi_read_va(vmi, idt_descriptor_address + sizeof(uint16_t), 0, sizeof(selector), &selector, &bytes_readen) ||
```
tklengyel: Just use `vmi_read_16_va` here..

libvmi/os/windows/core.c
```diff
@@ -661,6 +661,280 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
...
+     *    // following structure varies depending on page mode
+     * }
+     * 
+     * http://wiki.osdev.org/Interrupt_Descriptor_Table
+     * */
+    uint16_t selector = 0;
+
+    size_t bytes_readen = 0;
+    if (VMI_FAILURE == vmi_read_va(vmi, idt_descriptor_address + sizeof(uint16_t), 0, sizeof(selector), &selector, &bytes_readen) || 
+        bytes_readen != sizeof(selector))
```
tklengyel: ...and then you don't need this.

libvmi/os/windows/core.c
```diff
@@ -661,6 +661,280 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
...
+     * that represents memory segment.
+     * We need to determine starting address
+     * of memory segment.
+     * 
+     * http://wiki.osdev.org/GDT
+     * */
+    uint8_t buffer[8] = {0};
+
+    bytes_readen = 0;
+    if (VMI_FAILURE == vmi_read_va(vmi, gdtr, 0, sizeof(buffer), buffer, &bytes_readen) ||
```
tklengyel: Use `vmi_read_64_va` and just cast `(uint64_t *)buffer`

libvmi/os/windows/core.c
```diff
@@ -661,6 +661,280 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
...
+        .translate_mechanism = VMI_TM_PROCESS_PID,
+        .addr = 0,
+        .pid = 0
+    };
+
+    // virtual address of PE-image is always aligned to 4K boundaries
+    start_address_va &= ~mask_4k;
+
+    size_t bytes_readen = 0;
+    while (start_address_va > VMI_PS_4KB)       // do not check virtual address 0x1000, because there is no kernel there
```
tklengyel: Going anywhere under `0xffff800000000000` is guaranteed to not have a kernel on 64-bit machines. On 32-bit going under `0x80000000` is also safe to assume to be the case (albeit that boundary being configurable).

tklengyel:
So while this approach would work, the only thing really that changed in WIndows is that the first address in the IDTR is now named `KiDivideErrorFaultShadow` instead of `KiDivideErrorFault`. It would be a lot cleaner and faster to just look up that address in the Rekall profile in case initializing with the RVA of `KiDivideErrorFault` fails.

libvmi/os/windows/core.c
ReallyNothing: Sorry, my mistake. I was "porting" it from the older version, where this function returned number of bytes, written into the buffer.

libvmi/os/windows/core.c
```diff
@@ -661,6 +661,256 @@ get_kpgd_from_rekall_profile(vmi_instance_t vmi)
...
+    /* Each entry has a complex structure,
+     * that represents memory segment.
+     * We need to determine starting address
+     * of memory segment.
+     *
+     * http://wiki.osdev.org/GDT
+     * */
+    uint8_t buffer[8] = {0};
+
+    bytes_readen = 0;
```
tklengyel: This is now unused.

tklengyel:
@libvmi-jenkins Retest this please.

tklengyel:
@libvmi-jenkins Retest this please.

tklengyel:
In light of PR #564 I don't think this is necessary any more.

ReallyNothing:
Yes, the approach from [#564](https://github.com/libvmi/libvmi/pull/564) is more effective. I need to cancel this pull request.

PR #571, opened 2018-01-30
[linux-offset-finder] Add mm_types.h header

mmomeu:
The struct mm_struct not found by the compiler. This commit includes
<linux/mm_types.h> which exposes struct mm_struct. The struct mm_struct
was previously exported through <linux/sched.h>, however, this has been
changed in Linux version 4.11.

Signed-off-by: Marius Momeu <momeu@sec.in.tum.de>


tklengyel:
Thanks!

PR #587, opened 2018-03-27
fix request header alignment for kvm patch kvm-qemu-v2.8

azizulhakim:


tklengyel:
I'm not sure about this patch. `type` was deliberately typed as `uint64_t` so ABI holes won't be an issue.

azizulhakim:
In kvm driver, `type` is `uint8_t` and structure memory padding puts nonzero values when padding it to `uin64_t`. So in kvm patch, value of `type` is some garbage value instead of [0-2].

tklengyel:
I see what you are saying, but then the solution is to fix the driver, not the patch.

azizulhakim:
I agree. But in that case, other kvm patch also needs to be fixed which may cause problems with people already using those patches. All other kvm patch has `type` field as `uint8_t`. Let me know what you think.

tklengyel:
I would say standardizing on `uint64_t` across all patches is the right way forward. So if old patches still have `uint8_t` for the type, those should be upgraded as well.

azizulhakim:
Okay. But is there any specific reason of choosing `uint64_t` over `uint8_t` as the `type` value is never going to be that large?

tklengyel:
Yes, to ensure ABI compatibility. When `uint8_t` is used it leaves it up to the compiler to define how much padding there should be. Compilers do this to make variables cache-aligned for performance. However, compilers can choose to do whatever they want. So since we can't ensure that LibVMI is compiled with the same compiler KVM is, the two sides when compiled could get misaligned. If we specify type to be `uint64_t`, there is no hole and there is no padding the compilers can disagree on.

azizulhakim:
Got it. I will update the pull request.

tklengyel:

PR #600, opened 2018-04-19
Fix KASLR offset bug for x86_64

richarduu:
When the kernel version is 4.13.0+, the previous kaslr lookup function
fails somehow.

Determining KASLR offset in x86_64 can be done in a simpler way following
the official kernel address layout documentation:
    https://www.kernel.org/doc/Documentation/x86/x86_64/mm.txt

What we need to do is just keep attemping to get the k2p mapping of the first
page mapped in kernel text mapping. We can get the offset once we get the first
kernel text page mapped. Because in the range of 0xffffffff80000000 ~
0xffffffffa0000000 there is only kernel text, and the first page found mapped
beginning from 0xffffffff81000000 is the start of the kernel text mapping, thus
we can get the kaslr offset.

Note that here it should be 0xffffffff81000000 instead of 0xffffffff80000000.

Also, kernel text is by default using 2M huge page so we are adding 0x200000
instead of 0x1000.

Tested on x86_64 Ubuntu 16.04 with kernel v4.4.0 and v4.15.0.

### DIFF
--- a/libvmi/os/linux/core.c
+++ b/libvmi/os/linux/core.c
@@ -253,6 +253,24 @@ static status_t init_task_kaslr_test(vmi_instance_t vmi, addr_t page_vaddr)
     return ret;
 }
 
+status_t get_kaslr_offset_ia32e(vmi_instance_t vmi)
+{
+    addr_t va, pa;
+    addr_t kernel_text_start = 0xffffffff81000000;
+    addr_t kernel_text_end = kernel_text_start + (1024*1024*1024);
+
+    linux_instance_t linux_instance = vmi->os_data;
+    for (va = kernel_text_start; va < kernel_text_end; va += 0x200000) {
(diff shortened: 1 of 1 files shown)


richarduu:
One check failed because of a `{` placed at the same line with the function name... 
Should I cancel it and create another pull request, or just ignore it?

```
-status_t get_kaslr_offset_ia32e(vmi_instance_t vmi){
+status_t get_kaslr_offset_ia32e(vmi_instance_t vmi)
+{
```

smaresca:
@richarduu no need to create another pull request, just make the modification, commit, squash/fixup down to one commit + push

richarduu:
Thanks @smaresca , it is squashed, and passed.

tklengyel:

tklengyel:
Looks good, thanks!

Issue #597, opened 2018-04-17
Status of Debian packaging


Wenzel:
I would like to open this issue to track the status of the Debian packaging.

- Does it work ?
- Is the quality enough that we could consider an upload to the official Debian repositories ?

Thanks.

Wenzel:
> Does it work ?

Running `debuild -us -uc`, the package is generated
~~~
-rw-r--r--  1 tarrma tarrma  86K avril 17 11:59 libvmi0_0.11_amd64.deb
-rw-r--r--  1 tarrma tarrma  85K avril 17 11:59 libvmi_0.11_amd64.build
-rw-r--r--  1 tarrma tarrma 1,5K avril 17 11:59 libvmi_0.11_amd64.changes
-rw-r--r--  1 tarrma tarrma  717 avril 17 11:59 libvmi_0.11.dsc
-rw-r--r--  1 tarrma tarrma 889K avril 17 11:59 libvmi_0.11.tar.xz
-rw-r--r--  1 tarrma tarrma 123K avril 17 11:59 libvmi-dev_0.11_amd64.deb
~~~

I have a ton of `lintian` warnings/errors though:
~~~
Now running lintian...
E: libvmi changes: bad-distribution-in-changes-file unstable
E: libvmi source: source-contains-unsafe-symlink INSTALL
E: libvmi source: source-contains-unsafe-symlink compile
E: libvmi source: source-contains-unsafe-symlink config.guess
E: libvmi source: source-contains-unsafe-symlink config.sub
E: libvmi source: source-contains-unsafe-symlink depcomp
E: libvmi source: source-contains-unsafe-symlink install-sh
E: libvmi source: source-contains-unsafe-symlink missing
E: libvmi source: source-contains-unsafe-symlink test-driver
E: libvmi source: source-contains-unsafe-symlink ylwrap
W: libvmi source: debhelper-but-no-misc-depends libvmi-dev
E: libvmi source: build-depends-on-build-essential build-depends
W: libvmi source: file-without-copyright-information .astylerc
W: libvmi source: file-without-copyright-information .idea/inspectionProfiles/Project_Default.xml
W: libvmi source: file-without-copyright-information .idea/inspectionProfiles/profiles_settings.xml
W: libvmi source: file-without-copyright-information .idea/libvmi.iml
W: libvmi source: file-without-copyright-information .idea/misc.xml
W: libvmi source: file-without-copyright-information .idea/modules.xml
[...]

PR #601, opened 2018-04-19
Bump version to 0.12

tklengyel:


Wenzel:
I would suggest to bump also the `debian/changelog`.
Btw, what is the `debian/libvmi0.symbols` file ?
the version number `0.11` is everwhere in this one.

tklengyel:
The debian packaging is not really well maintained, `libvmi0.symbols` looks autogenerated but I'm not sure by what.

Wenzel:
I opened an issue for that:
https://github.com/libvmi/libvmi/issues/597

I wanted to track what was needed to get it uploaded to debian mirrors.

tklengyel:
@Wenzel it was git moved, it is no longer in the folder

PR #603, opened 2018-04-22
Pass init_flags and init_data to driver_test() functions

smaresca:
Minor modification of driver_init_mode to pass init_flags and init_data parameters down to the driver test function, e.g., xen_test(). 

Preparatory work for a) a new driver that needs to perform sanity checking during init, b) enhanced tests for existing drivers.


tklengyel:
@smaresca LGTM, will merge once I've cut the 0.12 release as this doesn't affect the public API

PR #605, opened 2018-04-23
Bump version to 0.13

tklengyel:


Wenzel:
But why ? :thinking: 
We just bumped to 0.12

tklengyel:
Even-numbers are for releases, odd numbers are for dev

PR #620, opened 2018-05-17
meltdown kernel patch fix (Linux)

scorpionzezz:
Hi, 

as discussed here: https://groups.google.com/forum/#!topic/vmitools/w6t2NgYYaug
We do not have Windows VM so we could not know whether this thing also happen in windows.
Affected kernel version:
4.14.12
9.9.75
4.4.110


tklengyel:

PR #629, opened 2018-05-30
This is a fix for systems without meltdown patches

bentau:
After masking the last bits of pgd of the CR3 register
it can occur that it's different to the PGD of the task struct.
Not sure if this will fix all problems.


tklengyel:

PR #635, opened 2018-06-05
add VMI_EVENT_RESPONSE_EMULATE_NOWRITE in xen driver so that xen will…

scorpionzezz:
… process the flag


tklengyel:
Good catch!

PR #637, opened 2018-06-11
ARM: improved support for privcall traps

ufo95:
The current implementation does not return event details to the event callback.

This patch extends process_privcall_event to return more informations of the event.
This adds better error handling to xen_set_privcall_event_48.
xen_events_destroy_48 now also disables privcall events.


ufo95:
Closed because of the failing travis test.

ufo95:
okay sorry for the chaos. I have now created a new pull request at: #638

PR #638, opened 2018-06-11
ARM: improved support for privcall traps

ufo95:
The current implementation does not return event details to the event callback.

This patch extends process_privcall_event to return more informations of the event.
This adds better error handling to xen_set_privcall_event_48.
xen_events_destroy_48 now also disables privcall events.

### DIFF
--- a/libvmi/driver/xen/xen_events_48.c
+++ b/libvmi/driver/xen/xen_events_48.c
@@ -150,7 +150,11 @@ void process_response ( event_response_t response, vmi_event_t* event, vm_event_
                         }
                         break;
                     case VMI_EVENT_RESPONSE_SET_REGISTERS:
+#if defined(I386) || defined(X86_64)
                         memcpy(&rsp->data.regs.x86, event->x86_regs, sizeof(struct regs_x86));
+#elif defined(ARM32) || defined(ARM64)
+                        memcpy(&rsp->data.regs.arm, event->arm_regs, sizeof(struct regs_arm));
+#endif
                         break;
                 };
 
@@ -332,7 +336,11 @@ status_t process_register(vmi_instance_t vmi,
 
     event->slat_id = (req->flags & VM_EVENT_FLAG_ALTERNATE_P2M) ? req->altp2m_idx : 0;
     event->vcpu_id = req->vcpu_id;
+#if defined(I386) || defined(X86_64)
     event->x86_regs = (x86_registers_t *)&req->data.regs.x86;
+#elif defined(ARM32) || defined(ARM64)
+    event->arm_regs = (arm_registers_t *)&req->data.regs.arm;
+#endif
 
     vmi->event_callback = 1;
     process_response ( event->callback(vmi, event), event, rsp );
@@ -455,7 +463,11 @@ static status_t process_guest_requested_event(vmi_instance_t vmi,
         return VMI_FAILURE;
     }
 
+#if defined(I386) || defined(X86_64)
     event->x86_regs = (x86_registers_t *)&req->data.regs.x86;
+#elif defined(ARM32) || defined(ARM64)
+    event->arm_regs = (arm_registers_t *)&req->data.regs.arm;
+#endif
     event->slat_id = (req->flags & VM_EVENT_FLAG_ALTERNATE_P2M) ? req->altp2m_idx : 0;
     event->vcpu_id = req->vcpu_id;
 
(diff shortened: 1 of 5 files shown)


tklengyel:

libvmi/driver/xen/xen_events_48.c
```diff
@@ -1335,11 +1387,15 @@ void xen_events_destroy_48(vmi_instance_t vmi)
...
     (void)xen->libxcw.xc_monitor_write_ctrlreg(xch, dom, VM_EVENT_X86_CR0, false, false, false);
     (void)xen->libxcw.xc_monitor_write_ctrlreg(xch, dom, VM_EVENT_X86_CR3, false, false, false);
     (void)xen->libxcw.xc_monitor_write_ctrlreg(xch, dom, VM_EVENT_X86_CR4, false, false, false);
     (void)xen->libxcw.xc_monitor_write_ctrlreg(xch, dom, VM_EVENT_X86_XCR0, false, false, false);
     (void)xen->libxcw.xc_monitor_software_breakpoint(xch, dom, false);
+#elif defined(ARM32) || defined(ARM64)
+    (void)xen->libxcw.xc_monitor_privileged_call(xch, dom, false);
+#endif
     (void)xen_set_guest_requested_event_48(vmi, 0);
     (void)xen_set_cpuid_event_48(vmi, 0);
```
tklengyel: The cpuid and debug events are also x86 specific.

libvmi/events.h
```diff
@@ -337,6 +337,15 @@ typedef struct {
...
 } interrupt_event_t;
 
+typedef struct {
+    int8_t reinject;
+
+    /* OUT */
+    addr_t gla;         /**< (Global Linear Address) == PC of the trapped instruction */
+    addr_t gfn;         /**< (Guest Frame Number) == 'physical' page where trap occurred */
+    addr_t offset;      /**< Offset in bytes (relative to GFN) */
+} privcall_event_t;
```
tklengyel: This isn't used anywhere

libvmi/events.h
```diff
@@ -597,6 +606,14 @@ struct vmi_event {
             (_event)->callback = _callback; \
         } while(0)
 
+#define SETUP_PRIVCALL_EVENT(_event, _reinject, _callback) \
+        do { \
+            (_event)->version = VMI_EVENTS_VERSION; \
+            (_event)->type = VMI_EVENT_PRIVILEGED_CALL; \
+            (_event)->interrupt_event.reinject = _reinject; \
```
tklengyel: Privcalls are not interrupts so there is no `reinject` behavior

sergej-proskurin: Concerning re-injecting privcalls: while we don't need to reinject SMCs at the moment to place SMCs as traps at certain locations that are unlikely to contain SMCs. This might change if we start placing SMCs at arbitrary positions in the guest's code. For instance, SMC calls are used to communicate with TZ. If we talk about using a single-stepping mechanism using SMCs, we might stumble across situations where we would need to re-inject SMCs.

tklengyel: I'm not sure how that reinjection mechanism is envisioned to work.. Currently there is no support for that in Xen so I rather not have in the LibVMI ABI either.

sergej-proskurin: That's fair enough, thank you.

libvmi/driver/xen/xen_events_48.c
```diff
@@ -562,12 +574,30 @@ status_t process_privcall_event(vmi_instance_t vmi,
...
     event->slat_id = (req->flags & VM_EVENT_FLAG_ALTERNATE_P2M) ? req->altp2m_idx : 0;
     event->vcpu_id = req->vcpu_id;
+    event->interrupt_event.offset = req->data.regs.arm.pc & VMI_BIT_MASK(0,11);
+    event->interrupt_event.gla = req->data.regs.arm.pc;
+
+    if ( VMI_FAILURE == vmi_translate_kv2p(vmi, req->data.regs.arm.pc, &event->interrupt_event.gfn) ) {
+        errprint("%s: cannot translate pc to physical address\n", __FUNCTION__);
+        return VMI_FAILURE;
+    }
+    event->interrupt_event.gfn = event->interrupt_event.gfn >> 12;
```
tklengyel: Instead of stuffing these infos into the `interrupt_event` structure, please define a new structure in `libvmi/events.h` to hold them.

sergej-proskurin: I absolutely agree. Ulrich, as already mentioned before, privcalls are not interrupts. As such, both structures must be differentiated in the code, as well.

libvmi/driver/xen/xen_events_48.c
ufo95: You're right I will fix it

libvmi/events.h
ufo95: I will remove this line. It is not necessary now and only leads to confusion.

libvmi/driver/xen/xen_events_48.c
tklengyel: Thanks and remember you can just push to this branch ;)

PR #647, opened 2018-08-01
notify based on vcpu count

scorpionzezz:
Based on my performance test, start from 7 vCPUs, it will be faster to notify xen after each event (in the loop) not after all events being processed

screenshot from 2018-08-01 08-29-41

First column represent the vCPUs count. Second, third (before this patch) and fifth (after the patch) column represents the average number of int3 callback being called on after 1 minute run-time. I executed each case for 5 minutes.


tklengyel:
@scorpionzezz please provide more information what those numbers represent in the picture. It would be better if you post that data in the PR itself instead of linking to a picture somrwhere else.

scorpionzezz:
done :)

PR #648, opened 2018-08-01
notify based on vcpus for older xen (continuation #647)

scorpionzezz:


tklengyel:

PR #658, opened 2018-09-04
add more altp2m api

scorpionzezz:
hi, this might not be the final one but please let me know whether it is okay (in general).

### DIFF
--- a/libvmi/driver/xen/altp2m.c
+++ b/libvmi/driver/xen/altp2m.c
@@ -24,6 +24,36 @@
 #include "driver/xen/xen.h"
 #include "driver/xen/xen_private.h"
 
+status_t xen_altp2m_init (vmi_instance_t vmi, uint64_t *init_memsize)
+{
+    int rc;
+    xen_instance_t *xen = xen_get_instance(vmi);
+    xc_interface * xch = xen_get_xchandle(vmi);
+    domid_t domain_id = xen_get_domainid(vmi);
+
+    rc = xen->libxcw.xc_domain_setmaxmem(xch, domain_id, ~0);
+    if (rc < 0)
+        return VMI_FAILURE;
+
+    xc_dominfo_t info = { 0 };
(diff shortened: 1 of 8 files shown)


tklengyel:
I'm afraid these functions don't belong in `libvmi/slat.h` as they don't have anything to do with altp2m. These are functions used to add extra pages to the VM that can then be later used as altp2m shadow copies, but not necessarily. I'm not entirely sure they belong in LibVMI, even if they are better named.

libvmi/driver/xen/altp2m.c
```diff
@@ -24,6 +24,37 @@
...
+    return VMI_SUCCESS;
+}
+
+status_t xen_altp2m_deinit (vmi_instance_t vmi, uint64_t init_memsize)
+{
+    xen_instance_t *xen = xen_get_instance(vmi);
+    xc_interface * xch = xen_get_xchandle(vmi);
+    domid_t domain_id = xen_get_domainid(vmi);
+
+    xen->libxcw.xc_domain_setmaxmem(xch, domain_id, init_memsize);
```
tklengyel: This is a hack copied from DRAKVUF that I would not like to repeat in LibVMI. The only way to reverse the domain to its initial state would be get the return value of `XENMEM_maximum_reservation` hypercall before it got changed, and it is not currently exposed through `libxc`.

tklengyel: Never mind, some more grepping shows that `xc_domain_getinfo` returns that same initial value `XENMEM_maximum_reservation` does. I guess `XENMEM_maximum_reservation` is just kept around for compatibility reasons in case some toolstack still uses it.

scorpionzezz:
That's make sense. Any better suggestion to getting those functions (create new page and so on) since libvmi does not expose the xen wrapper and the interfaces.

tklengyel:
@scorpionzezz Not sure.. As I said, I don't know whether this really belongs in LibVMI as adding/removing pages to the VM is not really introspection related.

scorpionzezz:
True :+1:  I'm closing the PR

PR #666, opened 2018-09-18
Fix registering event for specific MSR

tklengyel:
For specific MSR events an additional lookup is needed on `vmi->msr_events`. There is also some clean-up and renaming done to make it more sane.


tklengyel:
@smaresca since I'm changing a public header here I will need your +1

PR #669, opened 2018-09-27
debian: add dependency on libvirt-dev to libvmi0 package

Wenzel:
`libvirt-dev` package provides the symlink:
`/usr/lib/x86_64-linux-gnu/libvirt.so`

which is required for the `KVM` driver initialization (`dlopen("libvirt.so")`:
https://github.com/libvmi/libvmi/blob/master/libvmi/driver/kvm/libvirt_wrapper.c#L44

### DIFF
--- a/libvmi/driver/kvm/libvirt_wrapper.c
+++ b/libvmi/driver/kvm/libvirt_wrapper.c
@@ -46,8 +46,13 @@ status_t create_libvirt_wrapper(kvm_instance_t *kvm)
     wrapper->handle = dlopen ("libvirt.so", RTLD_NOW | RTLD_GLOBAL);
 
     if ( !wrapper->handle ) {
-        dbprint(VMI_DEBUG_KVM, "--failed to open a handle to libvirt\n");
-        return VMI_FAILURE;
+        // fallback to libvirt.so.0
+        wrapper->handle = dlopen ("libvirt.so.0", RTLD_NOW | RTLD_GLOBAL);
+
+        if ( !wrapper->handle ) {
+            dbprint(VMI_DEBUG_KVM, "--failed to open a handle to libvirt\n");
+            return VMI_FAILURE;
+        }
     }
 
     wrapper->virConnectOpenAuth = dlsym(wrapper->handle, "virConnectOpenAuth");


tklengyel:
I'm not sure about this one, non -dev version of LibVMI should not depend on any -dev package.

Wenzel:
I know.
I'm not a big fan of this, but we have no way of knowing the full library name from the libvirt package.

Or do we ?

tklengyel:
Not sure what you mean: https://packages.debian.org/stretch/amd64/libvirt0/filelist has `/usr/lib/libvirt.so.0`

Wenzel:
I just had a case on a Debian Stretch installation where the libvirt wrapper could not be initialized because `libvirt.so` was not present, even if `/usr/lib/x86_64-linux-gnu/libvirt.so.0` was there..

Wenzel:
Let me try to reproduce this in a Docker container.

tklengyel:
The obvious fix then is to try to also open `libvirt.so.0` if `libvirt.so` is not there.

Wenzel:
> The obvious fix then is to try to also open libvirt.so.0 if libvirt.so is not there.

Opening  `libvirt.so.0` in the first place is the right fix in my opinion.

`-dev` packages are supposed to bring the symlink to `libxxx.so` and their `libxxx.a` static library.
We don't want to rely on `libvirt.so` and `-dev` packages being installed.

Wenzel:
Maybe you could also add a `dbprint` in case the libvirt wrapper could not be opened ?
I had `VMI_DEBUG_CORE | VMI_DEBUG_KVM` enabled and i couldn't see any log messages :/

Wenzel:
I will wait for the PR #670 to be merged, and then i can test my Dockerfile.

Wenzel:
Thanks @smaresca , i have my proof:
~~~Dockerfile
FROM debian:stable

RUN apt-get update && apt-get install -y build-essential git libtool \
        libvirt-dev autoconf pkg-config flex bison libglib2.0-dev libxen-dev libjson-c-dev \
        python3-pip libffi-dev ipython3

# install libvmi
RUN cd /root && git clone https://github.com/libvmi/libvmi
RUN cd /root/libvmi && \
        sed -i 's/\/\///g' libvmi/debug.h && \
        ./autogen.sh && ./configure --prefix=/usr && make && make install

# install libvmi python
RUN pip3 install libvmi

# remove libvirt-dev
RUN apt-get remove --purge -y libvirt-dev
~~~

:tada: 
failed

Wenzel:
I fixed the libvirt library name to `libvirt.so.0`
new `Dockerfile` to test
~~~Dockerfile
FROM debian:stable

RUN apt-get update && apt-get install -y build-essential git libtool \
        libvirt-dev autoconf pkg-config flex bison libglib2.0-dev libxen-dev libjson-c-dev \
        python3-pip libffi-dev ipython3

# install libvmi
RUN cd /root && git clone https://github.com/KVM-VMI/libvmi -b libvmi_depend_libvirt-dev
RUN cd /root/libvmi && \
        sed -i 's/\/\///g' libvmi/debug.h && \
        ./autogen.sh && ./configure --prefix=/usr && make && make install

# install libvmi python
RUN pip3 install libvmi

# remove libvirt-dev
RUN apt-get remove --purge -y libvirt-dev
~~~
result
libvmi

:tada:

Wenzel:
Interesting, there is no package which provides `libxenctrl.so.x`:
~~~
$ apt-file search -x 'libxenctrl.so*'
libxen-dev: /usr/lib/x86_64-linux-gnu/libxenctrl.so
~~~

You have to install `libxen-dev` to use Libvmi's Xen driver in production.

tklengyel:
It's just apparently named `libxenctrl-<version>.so` https://packages.debian.org/stretch/amd64/libxen-4.8/filelist

Wenzel:
Do you want to find a solution for the Xen driver also ?
hardcoding `libxenctrl-4.8.so` doesn't seem like a long-term solution :/

tklengyel:
Just have to add the format of the filename as a fallback after https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/libxc_wrapper.c#L148

Wenzel:
And it's not portable on other distribution :(
~~~
$ dnf provides '*/libxenctrl*'
xen-libs-4.10.0-8.fc28.x86_64 : Libraries for Xen tools
Repo        : fedora
Matched from:
Filename    : /usr/lib64/libxenctrl.so.4.10
Matched from:
Filename    : /usr/lib64/libxenctrl.so.4.10.0
~~~

tklengyel:
We already look for the file in the format `libxenctrl.so.<version>`

Wenzel:
rebased.
Should we merge ?

tklengyel:
I think this should still be used only as a fallback if `libvirt.so` is not found.

PR #671, opened 2018-10-09
Look for libxenctrl.so as packaged by Debian

tklengyel:


libvmi/driver/xen/libxc_wrapper.c
```diff
@@ -144,19 +144,26 @@ status_t create_libxc_wrapper(xen_instance_t *xen)
...
-        if ( !wrapper->handle ) {
-            fprintf(stderr, "Failed to find a suitable libxenctrl.so at any of the standard paths!\n");
-            return VMI_FAILURE;
-        }
+        wrapper->handle = dlopen (alternate, RTLD_NOW | RTLD_GLOBAL);
+        g_free(alternate);
+    }
+
+    if ( !wrapper->handle ) {
+        fprintf(stderr, "Failed to find a suitable libxenctrl.so at any of the standard paths!\n");
```
Wenzel: could you replace this by a `dbprint(`VMI_DEBUG_XEN` .....)` ?

tklengyel: For critical errors we are not using dbprint.

Wenzel: I did the same in my last PR, that's why i thought it was relevant:
https://github.com/libvmi/libvmi/blob/master/libvmi/driver/kvm/libvirt_wrapper.c#L49

tklengyel: The difference is that here we already know we are running under the Xen hypervisor as the version hypercall already succeeded. Not finding xenctrl afterwards is a critical error that warrants an error printout. In the libvirt wrapper that's not the case and printing an error would not be appropriate as it may just be that we are running under another hypervisor.

Wenzel: If LibVMI is deployed with Xen and KVM drivers, on KVM, this error message is always displayed :(

tklengyel: Hmm... It looks like I never I never actually ended up implementing getting the version via a hypercall and currently the major/minor versions are not set until the libxc wrapper is setup properly. Whoops, time to fix that :)

smaresca: Using /sys/hypervisor/version, CPUID leaves, or something else?

tklengyel: Well, this is annoying. Xen reports it's version number via CPUID only when executed in kernel mode. And it only accepts hypercalls coming from kernel mode...

smaresca:
yup. use /sys/hypervisor/version . it is being added for kvm now, too

tklengyel:
@smaresca Ah yea, I saw that too today. That works.

PR #672, opened 2018-10-16
Move Windows magic values from public header

tklengyel:
The KDBG size magic values have no use outside initialization, so remove them from the public header.


tklengyel:
@smaresca Touching the public header here so I'll need a +1

smaresca:
@tklengyel sorry, thought I gave the thumbs-up already. 

LGTM, +1

PR #675, opened 2018-11-03
Xen 4.12 support + events cleanup and coalescing

tklengyel:
This is a huge restructuring of the internal code handling Xen events. Due to how the Xen rings define custom C structures specific to the ABI versions, currently there is a massive code-duplication happening which makes the code hard to understand and maintain. Furthermore, bugfixes often have to be ported to each interface, which is error-prone.

In this patch we make the observation that the event control and event processing logic is in fact ABI independent, only the memory representation of the event is ABI specific. Thus, in this patch we convert each incoming event from the Xen ABI to an internal `vm_event_compat_t` structure that can be used for all Xen ABIs. Only this compat structure is utilized for event processing logic. Thus, the only code that remains ABI-specific are the mapping code to this compat structure and the functions that interact with the Xen ring directly.

Also note, support for Xen events ABI prior to Xen 4.6 is dropped in this patch.


smaresca:
@tklengyel I have this on my todo list for review, though I know it's in progress. Aiming for this afternoon

tklengyel:
@smaresca I'm finished tweaking it now, feel free to hit merge if you are good with it.

tklengyel:
Interestingly if we forgo calling `xc_evtchn_pending` before unmasking the event, unmasking doesn't take effect and the `poll` keeps returning, causing the CPU to spin 100%. Oddly there is no error reported when calling unmasking as such. This seems to be a bug in the Xen toolstack since `xc_evtchn_pending` should only be needed to tell us the port that fired in case there are multiple used. But here we know that port already. Odd.

tklengyel:
ping @smaresca

tklengyel:

smaresca:
@tklengyel ping received. Life reared its head and derailed me.  Saw the 4.12 push though, which gave me a swift kick

tklengyel:
Also, I've tested this PR with Xen 4.12-unstable and DRAKVUF and works as expected.

Issue #677, opened 2018-11-18
Cannot set RIP with vmi_set_vcpuregs()


Wenzel:
Hi !

I'm playing with the registers API at the moment, and I'm facing an issue when setting the `rip` register:
~~~Python
#!/usr/bin/env python3
from libvmi import Libvmi, X86Reg

register = X86Reg.RIP

vmi = Libvmi('winxp')
vmi.pause_vm()
# get regs
regs = vmi.get_vcpuregs(0)
print(hex(regs[register]))
# set RIP = 1
regs[register] = 1
vmi.set_vcpuregs(regs, 0)
# check
regs = vmi.get_vcpuregs(0)
print(hex(regs[register]))
vmi.resume_vm()
~~~

I added a few prints in `libvmi` C library:
~~~C
status_t
vmi_get_vcpuregs(
    vmi_instance_t vmi,
    registers_t *regs,
    unsigned long vcpu)
{
#ifdef ENABLE_SAFETY_CHECKS
    if (!vmi || !regs)
        return VMI_FAILURE;
#endif

    status_t status;
    status = driver_get_vcpuregs(vmi, regs, vcpu);
    printf("libvmi get RIP: %x\n", regs->x86.rip);
    return status;
}

status_t
vmi_set_vcpuregs(
    vmi_instance_t vmi,
    registers_t *regs,
    unsigned long vcpu)
{
#ifdef ENABLE_SAFETY_CHECKS
    if (!vmi || !regs)
        return VMI_FAILURE;
#endif

    printf("libvmi set RIP: %x\n", regs->x86.rip);
    return driver_set_vcpuregs(vmi, regs, vcpu);
}
~~~

and the result is the following:
~~~
libvmi get RIP: f67ab162
0xf67ab162
libvmi set RIP: 1
libvmi get RIP: f67ab162
0xf67ab162
~~~

The strange thing is that it works for other registers like `rax`:
~~~
libvmi get RAX: 61207e1f
0x61207e1f
libvmi set RAX: 1
libvmi get RAX: 1
0x1
~~~

What's happening here ? :thinking:

pjcolp:
Looking at xen_set_vcpuregs_hvm(), I see that there is no code to actually write rip. In xen_get_vcpuregs_hvm(), we have this:
 ```
    ...
    regs->x86.r14 = hvm_cpu->r14;
    regs->x86.r15 = hvm_cpu->r15;
    regs->x86.rip = hvm_cpu->rip;
[...]

PR #679, opened 2018-11-21
xen: Fix setting RIP with vmi_set_vcpuregs()

pjcolp:
Commit 992619cd0ed2 ("Get and set multiple registers on Xen") added the
funtion xen_set_vcpuregs_hvm() which sets all vcpu registers at the
same time. Unfortunately, it missed setting the RIP register.

Add setting cpu->rip in function xen_set_vcpuregs_hvm().


tklengyel:

Wenzel:
This should fix #677 .

PR #687, opened 2018-12-19
Add example tool for gathering kernel offsets for Windows

skvl:


examples/win-offsets.c
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#define _GNU_SOURCE
+
+#include <libvmi/libvmi.h>
+#include <libvmi/peparse.h>
+#include <libvmi/os/windows/windows.h>
```
tklengyel: This is a private header of LibVMI. You should not include it in an example.

skvl: I need it to print internal offsets. Then I could use it to create _libvmi.conf_.

tklengyel: The internal offsets are created _from_ the configuration that is passed _in_ to LibVMI. If you want a tool that automatically creates libvmi.conf take a look at `tools/config-editor`, `tools/linux-offset-finder` and `tools/windows-offset-finder`

skvl: Unfortunately I didn't find a way to get _ntoskrnl_ and _ntoskrnl_va_. Both are used in path of `windows_init()` if I understand the code. So the alternative is to add getters for it in `libvmi/accessors.c`

tklengyel: Sure, you can add new accessor functions to libvmi.h, there are is no explicit `vmi_get_kernel_base` function at the moment. I personally found getting a kernel symbol enough as based on that you can calculate the kernel base yourself. But having dedicated functions for this would probably.be more straight forward.

skvl: I have decided to extend `windows_get_offset()` to return all essential fields. Is it good enough?

tklengyel:

examples/win-offsets.c
```diff
@@ -0,0 +1,171 @@
...
+    g_hash_table_insert(config, g_strdup("os_type"), g_strdup("Windows"));
+    g_hash_table_insert(config, g_strdup("rekall_profile"), g_strdup(rekall_profile));
+
+    if (VMI_PM_UNKNOWN == vmi_init_paging(vmi, VMI_PM_INITFLAG_TRANSITION_PAGES) ) {
+        printf("Failed to init LibVMI paging.\n");
+        g_hash_table_destroy(config);
+        return 1;
+    }
+
+    vmi_init_os(vmi, VMI_CONFIG_GHASHTABLE, config, NULL);
```
tklengyel: You need to check the return value here and verify that it Windows.

examples/win-offsets.c
```diff
@@ -0,0 +1,175 @@
...
+    if (VMI_PM_UNKNOWN == vmi_init_paging(vmi, VMI_PM_INITFLAG_TRANSITION_PAGES) ) {
+        printf("Failed to init LibVMI paging.\n");
+        g_hash_table_destroy(config);
+        return 1;
+    }
+
+    os_t os = vmi_init_os(vmi, VMI_CONFIG_GHASHTABLE, config, NULL);
+    if (VMI_OS_WINDOWS != os) {
+        printf("Failed to init LibVMI library.\n");
+        return 1;
```
tklengyel: You are leaking memory here for the ghastable and the for the vmi_instance.

examples/win-offsets.c
```diff
@@ -0,0 +1,175 @@
...
+        printf("Failed to create GHashTable!\n");
+        return 1;
+    }
+
+    g_hash_table_insert(config, g_strdup("os_type"), g_strdup("Windows"));
+    g_hash_table_insert(config, g_strdup("rekall_profile"), g_strdup(rekall_profile));
+
+    if (VMI_PM_UNKNOWN == vmi_init_paging(vmi, VMI_PM_INITFLAG_TRANSITION_PAGES) ) {
+        printf("Failed to init LibVMI paging.\n");
+        g_hash_table_destroy(config);
```
tklengyel: You are leaking memory here for the vmi_instance.

skvl: Fixed!

examples/win-offsets.c
skvl: Fixed!

PR #688, opened 2018-12-24
Fix error in initializing vmi->kpgd from CR3 for Windows

skvl:
Because of the error the check `if (VMI_SUCCESS == real_kpgd_found)` fails every time. So other methods (`get_kpgd_from_rekall_profile`, `get_kpgd_method0`, `get_kpgd_method1`, `get_kpgd_method2`) has to be called to get it.

### DIFF
--- a/libvmi/os/windows/core.c
+++ b/libvmi/os/windows/core.c
@@ -1017,6 +1017,8 @@ windows_init(vmi_instance_t vmi, GHashTable *config)
         } else {
             real_kpgd_found = VMI_SUCCESS;
         }
+    } else {
+        real_kpgd_found = VMI_SUCCESS;
     }
 
     if (VMI_FAILURE == init_core(vmi))


skvl:
Then what is the purpose of the code:
```
1040     if (VMI_FAILURE == driver_get_vcpureg(vmi, &vmi->kpgd, CR3, 0)) {                                         
1041         if (VMI_FAILURE == get_kpgd_method2(vmi)) {                                                           
1042             errprint("Could not get kpgd, will not be able to determine page mode\n");                        
1043             goto done;                                                                                        
1044         } else {                                                                                              
1045             real_kpgd_found = VMI_SUCCESS;                                                                    
1046         }                                                                                                     
1047     }
```

tklengyel:
Well, if `get_kpgd_method2` returns true, then the real kpgd is found. So we set the variable accordingly.

tklengyel:
Also, keep in mind that some of that funky logic is there when introspecting files, not live memory. In those cases getting vcpu registers is not possible, there is only some heuristics in place to find a possible canditate for cr3.

tklengyel:
The commend above the block you copied is also pretty clear:
```
/* At this point we still don't have a directory table base,
     * so first we try to get it via the driver (fastest way).
     * If the driver gets us a dtb, it will be used _only_ during the init phase,
     * and will be replaced by the real kpgd later. */
```

PR #690, opened 2018-12-25
Move Windows version detection from init_from_rekall_profile_real

skvl:
This allows some optimizations while using _libvmi.conf_.


tklengyel:

PR #695, opened 2018-12-28
Only forgo polling when external event channel is used 

### LINKED ISSUE #694: Failed to read port from event channel
Hi everyone

still working on **pyvmidbg**, and I got a new error that I never had before with Xen.
Usually, after I stopped listening to VMI events, I'm [clearing the queue](https://github.com/libvmi/libvmi/blob/master/libvmi/events.h#L753) by issuing a call to `vmi_events_listen()` with `timeout = 0`:
~~~Python
    def action_singlestep(self):
        cb_data = {
            'interrupted': False
        }

        def cb_on_sstep(vmi, event):
            self.log.debug('singlestepping')
            vmi.pause_vm()
            cb_data['interrupted'] = True
        # unregister sstep_recoil
        self.ctx.vmi.clear_event(self.ss_event_recoil)

        num_vcpus = self.ctx.vmi.get_num_vcpus()
        ss_event = SingleStepEvent(range(num_vcpus), cb_on_sstep)
        self.ctx.vmi.register_event(ss_event)

        self.ctx.vmi.resume_vm()
        while not cb_data['interrupted']:
            self.ctx.vmi.listen(1000)

  -->   self.ctx.vmi.listen(0)
        self.ctx.vmi.clear_event(ss_event)

        # reregister sstep_recoil
        self.ctx.vmi.register_event(self.ss_event_recoil)
~~~

This time, this call doesn't return for no clear reason.
When I `CTRL-C`, i get a Python backtrace and the following error:
`VMI_ERROR: Failed to read port from event channel`
which corresponds to:
https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L1820

Any idea what's happening ? :thinking: 

xen_version            : 4.11.1 (distro package)
Fedora 29

Thanks !

tklengyel:
Fixes #694

### DIFF
--- a/libvmi/driver/xen/xen_events.c
+++ b/libvmi/driver/xen/xen_events.c
@@ -1815,12 +1815,12 @@ status_t unmask_event(xen_instance_t *xen, xen_events_t *xe)
 {
     int rc, port = xen->libxcw.xc_evtchn_pending(xe->xce_handle);
 
-#ifdef ENABLE_SAFETY_CHECKS
     if ( -1 == port ) {
-        errprint("Failed to read port from event channel\n");
+        dbprint(VMI_DEBUG_XEN, "No event channel port is pending.\n");
         return VMI_FAILURE;
     }
 
+#ifdef ENABLE_SAFETY_CHECKS
     if ( port != xe->port ) {
         errprint("Event received for invalid port %i, Expected port is %i\n",
                  port, xe->port);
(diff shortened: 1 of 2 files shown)


tklengyel:
Not calling `poll` before trying to check if there is a Xen event channel port pending results in the syscall hanging (at least as of Linux 4.9). The only time we need to forgo the internal poll is when the event channel handle is already polled on in an external application. Adjust internal logic accordingly.

PR #699, opened 2018-12-31
Remove the vmi_step_event API

tklengyel:
The reason we are removing this API is due to it promoting a potentially flawed use of VMI events. With multi-vCPU systems using `vmi_step_event` could cause the event being stepped to be missed on other executing vCPUs. The API was introduced when only single-vCPU systems were supported, where this scenario was not present. It is not currently possible to make this API safe for multi-vCPU systems. Recent static analysis with clang also revealed subtle bugs in the code.


tklengyel:
@smaresca: except the code is buggy and prevents the cmake PR from being merged without being fixed

thatreguy:
Why is this being deprecated? Could you please add in the description?

Wenzel:
Anything blocking you to merge this PR ?
Should we have a debate about _removing_ vs _deprecating_ ?

tklengyel:
@Wenzel our current policy is that PRs touching the API require two maintainers' +1, or in case its a PR from a maintainer, another +1 from a maintainer. We might revisit this policy if it becomes problematic and move to lazy consensus on all PRs.

PR #707, opened 2019-01-11
Read rekall profile file once

disaykin:


tklengyel:

tklengyel:
@disaykin This breaks compilation on machines where there is no json-c installed. Using Rekall profiles are optional so you need to put in the proper #ifdefs to make it work.

disaykin:
@tklengyel Done.

PR #708, opened 2019-01-11
use Ubuntu Xenial as base distro for travis tests

Wenzel:
Ubuntu Trusty's EOF will be in April 2019
https://wiki.ubuntu.com/Releases


Wenzel:
Actually `clang-7.0` is being used
`scan-build: Using '/usr/local/clang-7.0.0/bin/clang-7' for static analysis`
And it has found 2 bugs, so Travis won't go green here.

tklengyel:
One of the bugs will get eliminated by #699 but I'm not sure which one is the other bug it found.

.travis.yml
```diff
@@ -162,9 +162,9 @@ matrix:
         apt:
             sources:
                 - ubuntu-toolchain-r-test
-                - llvm-toolchain-trusty-5.0
+                - llvm-toolchain-xenial-6.0
```
tklengyel: switch to `llvm-toolchain-xenial-7`

.travis.yml
```diff
@@ -162,9 +162,9 @@ matrix:
...
             packages:
-                - clang-5.0
+                - clang-6.0
```
tklengyel: clang-7.0

tklengyel: Actually, if we have `llvm-toolchain-xenial-7` we don't even need to specify it as a package here

.travis.yml
```diff
@@ -174,15 +174,15 @@ matrix:
                 - libxen-dev
                 - libfuse-dev
       before_install:
-        - export PATH=$PATH:/usr/local/clang-5.0.0/bin
-        - export CC=clang-5.0
-        - export CXX=clang++-5.0
+        - export PATH=$PATH:/usr/local/clang-6.0.0/bin
+        - export CC=clang-6.0
+        - export CXX=clang++-6.0
```
tklengyel: These could all be removed

tklengyel:
We could also remove the `before_install` section of the astyle check

tklengyel:
Looks good to me now but won't be able to merge until the bugs making it trip are fixed

PR #709, opened 2019-01-14
add hypervisor support section in README

Wenzel:
Add a section in `README.rst` to quickly describe the hypervisors we are currently supporting and the features available.


README.rst
```diff
@@ -41,6 +41,20 @@ documentation online at http://www.libvmi.com.
 .. image:: https://scan.coverity.com/projects/14159/badge.svg
     :target: https://scan.coverity.com/projects/libvmi-libvmi
 
+Hypervisor support
+------------------
```
tklengyel: This is a good idea but we should probably have some more in-depth description of what the columns actually mean.  Some of the events listed here are also architecture specific and only available on arm or x86.

smaresca: +1

bentau: would be also pretty cool if you could add some details about the features that are supported by AMD processors

Wenzel: Could you be more specific ?

bentau: notes/events.txt says: "Currently only the Xen Hypervisor provides these features, and some of these are specifically only available on Intel CPU". Would be nice to know if this is still up to date or if some features still only work on Intel.

Wenzel: I updated the table to reflect that CPUID and MSR were x86 specific events.
I never worked on AMD architecture, so I'm sure about the descriptor, control register and debug register.
cc @bentau

Wenzel: update on this ?
@bentau , @tklengyel

bentau: i will try to contribute for AMD specific behavior soon

FelixBer: Hi, @Wenzel @bentau I am currently working on AMD issues, and I see 4 problems
1. single stepping not supported. AMD lacks the monitor-trap-flag, and xen doesnt implement any other way. i have a patch for xen, but currently I cannot test it.
2.  anything related to VMI_EVENT_MEMORY (e.g. syscall tracing example) results in a VMI_ERROR: xc_hvm_set_mem_access failed with code: -1. Despite my best efforts I cant seem to track down the underlying issue.
3. register monitoring works, but not with MSR_ALL - here I get hangs on either one or even both VMs, the monitor and the one being monitored! Again i am not sure why.
4. in the breakpoint example, re-injecting the bp doesnt work, but this is probably related to singlestepping not being available.

The rest actually seems to work fine, from working with the libvmi examples.

PR #710, opened 2019-01-15
Fix possible inconsistency while initializing OS

skvl:


examples/win-offsets.c
```diff
@@ -80,6 +80,12 @@ int main(int argc, char **argv)
         return 1;
     }
 
+    /* pause the vm for consistent memory access */
+    if (vmi_pause_vm(vmi) != VMI_SUCCESS) {
+        printf("Failed to pause VM\n");
+        return 1;
```
tklengyel: Returning here will leak the memory allocated for the vmi_instance_t. While it's not that big a deal here since the whole process is going away, for the examples we try to showcase the cleanest use of the library, which would involve cleaning up its memory after its done.

tklengyel:

PR #712, opened 2019-01-17
Save kpgd offset in libvmi.conf for use with VM snapshots

skvl:
Thus we search for `kpgd` only once while building _libvmi.conf_. The use it many times with every instance of VM created from snapshot.

### DIFF
--- a/examples/win-offsets.c
+++ b/examples/win-offsets.c
@@ -118,6 +118,7 @@ int main(int argc, char **argv)
     addr_t sysproc = 0;
     addr_t kpcr = 0;
     addr_t kdbg = 0;
+    addr_t kpgd = 0;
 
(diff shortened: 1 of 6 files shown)


tklengyel:
Please add reading in the `kpgd` to Linux and FreeBSD as well. It could be useful there too.

skvl:
@tklengyel I have added _kpgd_ support in _libvmi.conf_ for FreeBSD and Linux. Though there is no much use there because one just uses _CR3_ value. So there is no search which is different to Windows.

tklengyel:

tklengyel:
@skvl To the contrary, it could be useful to manually specify the pagetable to be used and not have LibVMI rely on an arbitrary CR3

PR #674, opened 2018-10-21
Switch to cmake

Wenzel:
I couldn't sleep and I'm already back from my future next week.
This should help fix #673 

A quick `CMake` tutorial:
~~~
mkdir build
cd build
cmake ..
make
~~~

:tada: 
capture d ecran de 2018-10-21 04-09-39

`CMake` pro tip:
use `ccmake ..` once you have alredy configured the build with `cmake ..`, this will call an interactive interface in your terminal to configure the build options:
capture d ecran de 2018-10-21 04-03-17

# TODO

## Options
- [x] `ENABLE_XEN`
- [x] `ENABLE_KVM`
- [x] `ENABLE_FILE`
- [x] `ENABLE_SHM_SNAPSHOT`
- [x] `ENABLE_WINDOWS`
- [x] `ENABLE_LINUX`
- [x] `ENABLE_FREEBSD`
- [x] `ENABLE_XENSTORE`
- [x] `ENABLE_VMIFS`
- [x] `ENABLE_CONFIGFILE`
- [x] `ENABLE_ADDRESS_CACHE`
- [x] `ENABLE_PAGE_CACHE`
- [x] `ENABLE_STATIC`
- [x] `MAX_PAGE_CACHE_SIZE`
- [x] `REKALL_PROFILES`
- [x] `ENABLE_SAFETY_CHECKS`
- [x] `BUILD_EXAMPLES`
- [x] `debug` -> `CMAKE_BUILD_TYPE`

## Check for libraries
- [x] check for `libvmi_request.h`
- [x] check for architecture `I386` `X86_64` `ARM32` `ARM64` 
- [x] find `GLib`
- [x] find `json-c` (if rekall_profile or kvm driver)
- [x] find `check` (if tests)
- [x] find `libvirt` (if kvm)
- [x] find `dlopen` (if xen or kvm drivers)
- [x] check xen deps
  - [x] `xenctrl.h xen/io/ring.h`
  - [x] `hvmmem_access_t`
  - [x] check for `xenstore.h xs.h` (if xenstore)

# Misc
- [x] handle `flex` and `bison` to generate the config file parser, and add the generated sources to the build
- [x] manage static and dynamic build of `libvmi`
- [x] make install
- [x] use `CMake` as a test suite ?
- [x] remove autotools
  - [x] remove autotools patterns in `.gitignore`
- [x] update `debian/control` dependencies
- [x] update README and documentation
- [x] update `.travis.yml` 

The `autotools` build system will be removed at the end.

[...]

PR #713, opened 2019-01-21
fix clang static analyzer warning in step_and_reg_events

Wenzel:
This PR fixes a warning generated by clang static analyzer on the `step_and_reg_events` function.

Related to #674, #708 , #699.
Warning:
screenshot_20190121_191840

Basically the `singlestep_event` memory is freed by `vmi_clear_event` call at line `274`, and the analyzer assumes that the `while` condition is still `true` after that.
So it enters into a test against a previously freed memory, therefore the warning.

I assume that I just have to put a `break` there, but i would be glad to have your comments on this function, that I don't know the internals.

Is this a good enough fix for you @tklengyel ?

### DIFF
--- a/libvmi/events.c
+++ b/libvmi/events.c
@@ -257,7 +257,7 @@ event_response_t step_and_reg_events(vmi_instance_t vmi, vmi_event_t *singlestep
         step_and_reg_event_wrapper_t *wrap =
             (step_and_reg_event_wrapper_t *) reg_list->data;
 
-        if (wrap->vcpu_id == singlestep_event->vcpu_id) {
+        if (singlestep_event && wrap->vcpu_id == singlestep_event->vcpu_id) {
             wrap->steps--;
         }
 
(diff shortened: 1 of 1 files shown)


tklengyel:
I think that break is only warranted under the same condition when `vmi_clear_event` is issued.

Wenzel:
> I think that break is only warranted under the same condition when vmi_clear_event is issued.

Do you wanted me to move the `break` after `vmi_clear_event()` call (which i did) ? :thinking:

tklengyel:
Yea... but now looking at it I don't know if breaking that loop will actually cause memory to leak in case the list still has `wrap` structures...

Wenzel:
What about this instead ?
I'm nullifying the pointer (event though it's a parameter), removing the break, so the loop can continue and free more wrap structures if needed.

tklengyel:
I think this will do.

Issue #726, opened 2019-02-14
Compiling on arm64 with cmake


scorpionzezz:
Hi,

I tried to compile libvmi with cmake on arm64, and I got following error:

> -- Could NOT find LibvmiRequest (missing:  LibvmiRequest_INCLUDE_DIR) 
CMake Error at cmake/modules/DetectArchitecture.cmake:15 (message):
  Undetect architecture aarch64 to be configured in config.h.in
Call Stack (most recent call first):
  CMakeLists.txt:26 (include)
-- Configuring incomplete, errors occurred!
See also "/mnt/test/libvmi/build/CMakeFiles/CMakeOutput.log".
See also "/mnt/test/libvmi/build/CMakeFiles/CMakeError.log".

Cheers!

Wenzel:
Hi @scorpionzezz ,
i wrote the `DetectArchitecture` cmake module based on internet research, i had no ARM processor to test it :)
As you can see, the error quite trivial to fix:
https://github.com/libvmi/libvmi/blob/master/cmake/modules/DetectArchitecture.cmake#L15

Can you make a PR ?
Thanks

scorpionzezz:
Hi @Wenzel 

I tried to change arm64 to "aarch64" and the "cmake .." works fine. But, then when I run "make"

>/mnt/test/libvmi/libvmi/driver/bareflank/hypercall.S: Assembler messages:
/mnt/test/libvmi/libvmi/driver/bareflank/hypercall.S:32: Error: unknown mnemonic `push' -- `push %rbx#rbx'
/mnt/test/libvmi/libvmi/driver/bareflank/hypercall.S:33: Error: unknown mnemonic `movq' -- `movq %rdx,%r8'
/mnt/test/libvmi/libvmi/driver/bareflank/hypercall.S:34: Error: unknown mnemonic `movq' -- `movq $1073746'
/mnt/test/libvmi/libvmi/driver/bareflank/hypercall.S:35: Error: unknown mnemonic `cpuid' -- `cpuid'
/mnt/test/libvmi/libvmi/driver/bareflank/hypercall.S:36: Error: unknown mnemonic `movq' -- `movq %rbx,(%r'
/mnt/test/libvmi/libvmi/driver/bareflank/hypercall.S:37: Error: unknown mnemonic `movq' -- `movq %rcx,(%r'
[...]

PR #727, opened 2019-02-14
add aarch64 on Cmake architecture

scorpionzezz:
based on #726  :+1:


cmake/modules/DetectArchitecture.cmake
```diff
@@ -11,6 +11,8 @@ elseif (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "arm32")
     set(ARM32 ON)
 elseif (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "arm64")
```
Wenzel: I think you can replace the `arm3264` strings by `aarch32/64`.

scorpionzezz: hmmm true, looks like it refers to the same thing (based on: https://unix.stackexchange.com/questions/461179/what-is-the-difference-between-different-implemetation-of-arm64-aarch64-for-linu)

scorpionzezz: done :+1:

Wenzel: maybe @tklengyel knows more about ARM ?

scorpionzezz: true, I tried with my Hikey960 board and it works, maybe @tklengyel  could try with the other boards.

tklengyel:
I got the following error on my cubietruck:

```
CMake Error at cmake/modules/DetectArchitecture.cmake:15 (message):
  Undetect architecture armv7l to be configured in config.h.in
Call Stack (most recent call first):
  CMakeLists.txt:26 (include)


-- Configuring incomplete, errors occurred!

scorpionzezz:
Then I guess we need to add arm7l as ARM32 ?

tklengyel:
Looks like it

scorpionzezz:
updated

tklengyel:
I'm still getting the same error. You seem to mistyped `armv7l` missing the v from it.

scorpionzezz:
ups, sorry, my bad, fixed :)

tklengyel:
OK, it gets passed the cmake phase. Unfortunately either though it detects now that it's building on/for ARM, it still tries to build the Bareflank driver which of course fails. It would be nice of the cmake configuration automatically disabled that (as the configure script did).

PR #731, opened 2019-02-22
Disable build components

### LINKED ISSUE #730: CMake doesn't disable components automatically
The cmake configuration doesn't automatically disable compiling parts of LibVMI based on the architecture and/or the availability of packages. For example, on ARM it should not build the Bareflank driver. Also, flex and json-c should be optional packages and cmake should disable the parts of LibVMI that depend on them (flex -> libvmi.conf, json-c -> rekall profiles/kvm/bareflank).

Wenzel:
Fix #730 .

This PR changes:
- Bareflank doesn't compile if ARM32 or 64 is detected as architecture
- KVM/Bareflank/rekall requires json-c and will be disabled if not detected
- Configfile requires flex and bison and will be disabled otherwise


tklengyel:

tklengyel:
Would it be possible with cmake to print the features that will be enabled at the end like configure did?

Wenzel:
Yes, there is a module that:
https://cmake.org/cmake/help/v3.0/module/FeatureSummary.html

I have never used it before but it looks easy to integrate.

tklengyel:
If you could add it, that would be great. Perhaps in a separate PR (I'm just gonna merge this one).

PR #733, opened 2019-03-04
Add get_xsave_info

aisaila:
This patch adds a new function to extract the cpu xsave info.
    
### DIFF
--- a/libvmi/accessors.c
+++ b/libvmi/accessors.c
@@ -174,6 +174,19 @@ vmi_get_kernel_struct_offset(
     return vmi->os_interface->os_get_kernel_struct_offset(vmi, symbol, member, addr);
 }
 
+status_t vmi_get_xsave_info(
+    vmi_instance_t vmi,
+    unsigned long vcpu,
+    xsave_area_t *xsave_info)
+{
+#ifdef ENABLE_SAFETY_CHECKS
+    if (!vmi || !xsave_info)
+        return 0;
+#endif
+
+    return driver_get_xsave_info(vmi, vcpu, xsave_info);
+}
+
 uint64_t
 vmi_get_memsize(
     vmi_instance_t vmi)
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -774,6 +774,101 @@ xen_get_tsc_info(
     return VMI_SUCCESS;
 }
 
+void
+xen_get_xsave_info_46(
+    struct hvm_hw_cpu_xsave_46 *info,
+    xsave_area_t *xsave_info)
+{
+    xsave_info->xcomp_bv = 0;
+    memcpy(xsave_info->fpu_sse, info->save_area.fpu_sse.x, sizeof(xsave_info->fpu_sse));
+    xsave_info->xstate_bv = info->save_area.xsave_hdr.xstate_bv;
+}
+
+void
+xen_get_xsave_info_412(
+    struct hvm_hw_cpu_xsave_412 *info,
+    xsave_area_t *xsave_info)
+{
+    memcpy(xsave_info->fpu_sse, info->save_area.fpu_sse.x, sizeof(xsave_info->fpu_sse));
+    xsave_info->xcomp_bv = info->save_area.xsave_hdr.xcomp_bv;
+    xsave_info->xstate_bv = info->save_area.xsave_hdr.xstate_bv;
+}
+
+status_t
+xen_get_xsave_info(
+    vmi_instance_t vmi,
+    unsigned long vcpu,
+    xsave_area_t *xsave_info)
+{
+    xen_instance_t *xen = xen_get_instance(vmi);
+    uint32_t size = 0;
+    uint8_t *buf = NULL;
+    status_t ret = VMI_SUCCESS;
+    uint32_t off = 0;
+    void *info = NULL;
+    struct hvm_save_descriptor *desc = NULL;
+
+    /* calling with no arguments --> return is the size of buffer required
+     *  for storing the HVM context
+     */
+    size = xen->libxcw.xc_domain_hvm_getcontext(xen->xchandle,xen->domainid, 0, 0);
+
+    if (size <= 0) {
+        errprint("Failed to fetch HVM context buffer size.\n");
+        ret = VMI_FAILURE;
+        goto _bail;
+    }
+
+    buf = malloc(size);
+    if (buf == NULL) {
+        errprint("Failed to allocate HVM context buffer.\n");
+        ret = VMI_FAILURE;
+        goto _bail;
+    }
+
+    if (xen->libxcw.xc_domain_hvm_getcontext(xen->xchandle,
+            xen->domainid, buf, size) < 0) {
+        errprint("Failed to fetch HVM context buffer.\n");
+        ret = VMI_FAILURE;
+        goto _bail;
+    }
+
+    off = 0;
+    while (off < size) {
+        desc = (struct hvm_save_descriptor *)(buf + off);
+
+        off += sizeof (struct hvm_save_descriptor);
+
+        if (desc->typecode == CPU_XSAVE_CODE && desc->instance == vcpu) {
+            info = (struct hvm_hw_cpu_xsave *)(buf + off);
+            break;
+        }
+        off += desc->length;
+    }
+
+    if (info == NULL) {
+        errprint("Failed to locate HVM xsave context.\n");
+        ret = VMI_FAILURE;
+        goto _bail;
+    }
+
+    switch (xen->minor_version) {
+        case 6 ... 11:
+            xen_get_xsave_info_46( (struct hvm_hw_cpu_xsave_46 *)info, xsave_info);
+            break;
+        default:
+        case 12:
+            xen_get_xsave_info_412( (struct hvm_hw_cpu_xsave_412 *)info, xsave_info);
+            break;
+    };
+
+_bail:
+
+    free(buf);
+
+    return ret;
+}
+
 #if defined(I386) || defined(X86_64)
 
 static status_t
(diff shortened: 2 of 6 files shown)


libvmi/accessors.c
```diff
@@ -189,6 +202,20 @@ vmi_get_memsize(
     return vmi->allocated_ram_size;
 }
 
+void *
+vmi_get_memory_pfn(
```
tklengyel: I don't think this is a good addition. LibVMI thus far deliberately avoided passing the raw pointer out so that it can clean up properly on shutdown. LibVMI also implements a cache that optimizes accesses to mapped-in pages so that if LibVMI functions need access to the page they don't have to remap it - which this new API now completely bypasses.

So this is a -1 from me. If the user needs this type of access to guest memory, the suggestion is to use libxc directly.

aisaila: We want to use libvmi with our introspection engine and we really actually do need this type of access to guest memory. What would be a good way to resolve the "map this page and write in it" case?
Our idea is to only use libvmi alone and without libxc in the same time.

tklengyel: That's what the `vmi_write` is for: https://github.com/libvmi/libvmi/blob/master/libvmi/libvmi.h#L1514

aisaila: In our case it's more efficient to work directly with mem pages

tklengyel: LibVMI by design doesn't give you direct access to the mem pages. You can either work with the existing API or you have to use libxc.

aisaila: Ok, I will drop this commit and re push

libvmi/libvmi.h
```diff
@@ -572,6 +572,19 @@ typedef struct registers {
...
+    struct {
+        struct { char x[512]; } fpu_sse;
+
+        struct hvm_hw_cpu_xsave_hd {
+            uint64_t xstate_bv;         /* Updated by XRSTOR */
+            uint64_t xcomp_bv;          /* Updated by XRSTOR{C,S} */
+            uint64_t reserved[6];
+        } xsave_hdr;                    /* The 64-byte header */
+    } save_area;
+} xen_cpu_xsave_info_t;
```
tklengyel: This seems to be a 1:1 copy from a Xen header. I don't recommend copying these directly but rather define a LibVMI specific structure that only holds the fields you actually need and document what you need these fields for. FOr example, if you only need `xcr0` then instead of creating a whole new API you should just extend `xen_get_vcpureg_hvm` to return XCR0 if queried for.

aisaila: I need xcr0 and save_area from that xsave structure. The structure is not 1:1 with the xen one. I've dropped the 
 uint64_t xfeature_mask;        /* Ignored */
 uint64_t xcr0_accum;           /* Updated by XSETBV */

tklengyel: The point of LibVMI is to abstract away the hypervisor specific bits and make the API applicable across-hypervisors (ideally). You are breaking that design here: you introduce a structure that is includes stuff like `xen` and `hvm`. The structure also have fields that are unexplained and their usecase is undocumented with LibVMI. What's `fpu_sse` used for? Nothing? Drop it. Why does it have `reserved` fields in there when it's not exposed through LibVMI as an ABI? So  this is a badly designed API right now.

aisaila: We use the SSE regs so fpu_sse is needed. I can drop the reserved filed but for the rest, we need that data. Is this ok or is this patch not usable?

tklengyel: As I said, bleeding Xen specific stuff into the LibVMI public header is a no-go. If you can clean up the API, and especially if you could introduce an API that is more extendable going forward, that would be preferred. The LibVMI API should also be as architecture -neutral as possible. So give some thought on how to do that. Introducing new APIs for stuff that are pretty much just pass-through from other accessible libraries is IMHO pretty much pointless.

tklengyel: That sounds like a hacky solution to me

aisaila: Yes, I agree but then what would be a good way to approach this or is the xsave a no go?

tklengyel: Not in this current form, no. If you can come up it with a cleaner API then it would be considered.

razvan-cojocaru: Hello Tamas,

In libbdvmi, we get the XSAVE area with the help of two helper functions, `getXSAVESize()`:

https://github.com/bitdefender/libbdvmi/blob/master/src/xendriver.cpp#L710

and `getXSAVEArea()`:

https://github.com/bitdefender/libbdvmi/blob/master/src/xendriver.cpp#L748

We need this in order to be able to work with the SSE registers. The most abstract way of doing this, AFAICT, is to simply fill up a user-provided (`void *`) buffer with the XSAVE area contents. If you look at "13.4 XSAVE AREA" in the Intel SDM, that's quite a lot of information in there, so A) it's not possible to decide on what is relevant and what is not, for the user, in all that, and B) if you look at how Xen and KVM look at that data, for example, you'll see that they're using different (small) parts of it.

We're fine passing this buffer along to our introspection engine to deal with it as it pleases, so for us the `void *` strategy is a perfectly fine use case. So we're not sure what you see as a better option for it, but if you'll let us know we'll try to fit the work into your design preferences.

Thanks.

tklengyel: @razvan-cojocaru I don't have anything against XSAVE itself or having a way to get these required bits of information exposed through the LibVMI API. I already expressed my concerns with how this PR intends to add support to it to the public API - including names like `xen` and a structure that looks as-if it was an ABI when it's not and including garbage fields. If those get fixed, the PR will be considered. Adding a void* structure return here would just make the API even more incomprehensible so that route would be -1 from me.

razvan-cojocaru: Fair enough, re-reading @aisaila's patch I think I know what you mean. :)
We'll re-shape the patch.

aisaila:
I've pushed a new version with a revised xsave structure

libvmi/libvmi.h
```diff
@@ -572,6 +572,16 @@ typedef struct registers {
     };
 } registers_t;
 
+typedef struct xsave_area {
+    struct { char x[512]; } fpu_sse;
+
+    struct {
+        uint64_t xstate_bv;         /* Updated by XRSTOR */
+        uint64_t xcomp_bv;          /* Updated by XRSTOR{C,S} */
+        uint64_t reserved[6];
```
tklengyel: Why is this reserved field still here?

libvmi/libvmi.h
```diff
@@ -572,6 +572,16 @@ typedef struct registers {
     };
 } registers_t;
 
+typedef struct xsave_area {
+    struct { char x[512]; } fpu_sse;
```
tklengyel: Why is there the struct wrapper? Completely unnecessery.

libvmi/libvmi.h
```diff
@@ -572,6 +572,16 @@ typedef struct registers {
...
+
+    struct {
```
tklengyel: Why is this struct here? Also unneeded

aisaila:
Ok, I will leave the structure like this then:

typedef struct xsave_area {
    char fpu_sse[512];
    uint64_t xstate_bv;         /* Updated by XRSTOR */
    uint64_t xcomp_bv;          /* Updated by XRSTOR{C,S} */

} xsave_area_t;

aisaila:
I've pushed a new version with the requested changes

libvmi/driver/xen/xen.c
```diff
@@ -774,6 +774,73 @@ xen_get_tsc_info(
...
+        off += desc->length;
+    }
+
+    if (info == NULL) {
+        errprint("Failed to locate HVM xsave context.\n");
+        ret = VMI_FAILURE;
+        goto _bail;
+    }
+
+    memcpy(xsave_info, &info->save_area, sizeof(struct xsave_area));
```
tklengyel: This assumes that the structure layout of `xsave_info` is the same as it is of `info->save_area`. That is not necessarily the case. Please copy fields explicitly instead. Also, add a safety check that makes sure `xsave_info` is not NULL.

aisaila: There is a small problem with the structure in xen 4.6 in that the xcomp_bv field was in the reserverd portion and then in 4.12 it came to be a field. If I assign per field the I will loose xcomp_bv on 4.6.

If that is the way we should go then I will have to have different versions according to xen version.

tklengyel: > If that is the way we should go then I will have to have different versions according to xen version.

Please do that otherwise this code is very fragile and could easily break in the future and make it extremely hard to debug.

libvmi/accessors.c
```diff
@@ -174,6 +174,19 @@ vmi_get_kernel_struct_offset(
     return vmi->os_interface->os_get_kernel_struct_offset(vmi, symbol, member, addr);
 }
 
+status_t vmi_get_xsave_info(
+    vmi_instance_t vmi,
+    unsigned long vcpu,
+    xsave_area_t *xsave_info)
+{
+#ifdef ENABLE_SAFETY_CHECKS
+    if (!vmi)
```
tklengyel: Please also NULL-check `xsave_info` here

libvmi/libvmi.h
```diff
@@ -572,6 +572,13 @@ typedef struct registers {
     };
 } registers_t;
 
+typedef struct xsave_area {
+    char fpu_sse[512];
+    uint64_t xstate_bv;         /* Updated by XRSTOR */
+    uint64_t xcomp_bv;          /* Updated by XRSTOR{C,S} */
+
```
tklengyel: Extra line

aisaila:
Is there something else needed for this pr?

tklengyel:
@smaresca LGTM, hit merge when you get a chance

tklengyel:

PR #737, opened 2019-03-13
Add domain watcher

aisaila:
This patch adds a domain watcher that signals a new domain or if a domains was destroyed. The callback function returns a structure with domain id, uuid and if the domain was created or not.
    
After a domain is created it is inserted in a gTree so we can have
information about the destroy event.

### DIFF
--- a/examples/wait-for-domain-example.c
+++ b/examples/wait-for-domain-example.c
@@ -1,0 +1,85 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+ * memory in a target virtual machine or in a file containing a dump of
+ * a system's physical memory.  LibVMI is based on the XenAccess Library.
+ *
+ * Author: Alexandru Isaila (aisaila@bitdefender.com)
+ *
+ * This file is part of LibVMI.
+ *
+ * LibVMI is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.
+ *
+ * LibVMI is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdlib.h>
+#include <signal.h>
+#include <libvmi/libvmi.h>
+#include <libvmi/events.h>
+
+static int interrupted = 0;
+static void close_handler(int sig)
+{
+    interrupted = sig;
+}
+vmi_event_t watch_event = {0};
+
+event_response_t wait_for_domain(vmi_instance_t vmi, vmi_event_t *event)
+{
+
+    if ( event->watch_event.created )
+        printf("domain %d with uuid %s created\n", event->watch_event.domain, event->watch_event.uuid);
+    else
+        printf("domain %d with uuid %s deleted\n", event->watch_event.domain, event->watch_event.uuid);
+
+    return 1;
+}
+
+int main ()
+{
+    vmi_instance_t vmi;
+    struct sigaction act;
+    act.sa_handler = close_handler;
+    act.sa_flags = 0;
+    sigemptyset(&act.sa_mask);
+    sigaction(SIGHUP,  &act, NULL);
+    sigaction(SIGTERM, &act, NULL);
+    sigaction(SIGINT,  &act, NULL);
+    sigaction(SIGALRM, &act, NULL);
+
+    // Initialize the libvmi library.
+    if (VMI_FAILURE ==
+            vmi_init(&vmi, VMI_XEN, NULL, VMI_INIT_DOMAINWATCH, NULL, NULL)) {
+        printf("Failed to init LibVMI library.\n");
+        return 1;
+    }
+
+    printf("LibVMI init succeeded!\n");
+    watch_event.version = VMI_EVENTS_VERSION;
+    watch_event.type = VMI_EVENT_DOMAIN_WATCH;
+    watch_event.callback = wait_for_domain;
+
+    if (vmi_register_event(vmi, &watch_event) == VMI_FAILURE)
+        printf("Failed to register event\n");
+
+    printf("Wait for domains\n");
+
+    while (!interrupted) {
+        vmi_events_listen(vmi,500);
+    }
+
+    printf("Finished with test.\n");
+
+    // cleanup any memory associated with the libvmi instance
+    vmi_destroy(vmi);
+
+    return 0;
+}
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -544,6 +544,20 @@ status_t vmi_init(
     }
     dbprint(VMI_DEBUG_CORE, "--completed driver init.\n");
 
+    if (init_flags & VMI_INIT_DOMAINWATCH) {
+        if ( VMI_FAILURE == driver_domainwatch_init(_vmi, init_flags) ) {
+            if ( error )
+                *error = VMI_INIT_ERROR_DRIVER;
+            goto error_exit;
+        }
+
+        if ( init_flags == VMI_INIT_DOMAINWATCH ) {
+            /* we have all we need to wait for domains. Return if there is nothing else */
+            *vmi = _vmi;
+            return VMI_SUCCESS;
+        }
+    }
+
     /* resolve the id and name */
     if (VMI_FAILURE == set_id_and_name(_vmi, domain)) {
         if ( error )
--- a/libvmi/driver/driver_interface.h
+++ b/libvmi/driver/driver_interface.h
@@ -39,6 +39,9 @@ typedef struct driver_interface {
         vmi_instance_t,
         uint32_t init_flags,
         vmi_init_data_t *init_data);
+    status_t (*domainwatch_init_ptr) (
+        vmi_instance_t vmi,
+        uint32_t init_flags);
     void (*destroy_ptr) (
         vmi_instance_t);
     uint64_t (*get_id_from_name_ptr) (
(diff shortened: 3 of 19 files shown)


examples/wait-for-domain-example.c
```diff
@@ -0,0 +1,74 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+ * memory in a target virtual machine or in a file containing a dump of
+ * a system's physical memory.  LibVMI is based on the XenAccess Library.
+ *
+ * Author: Steven Maresca (steven.maresca@zentific.com)
```
tklengyel: I don't think you are Steve ;)

aisaila: Ah yes, I will change that. I was copied from another test

tklengyel:
This is a good start but I think it would make sense if this would actually be integrated with the `vmi_events_listen` API. That way a user of the API could just have one loop with a new event type that says effectively that the domain is now introduced. Of course there would be a need for a new `VMI_INIT_DOMAINWATCH` flag for `vmi_init` that would set things up enough for that work.

aisaila:
Ok, I will add this functionality to the vmi_events_listen

aisaila:
I've pushed a new version for the domain watch event

examples/wait-for-domain-example.c
```diff
@@ -0,0 +1,88 @@
...
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdlib.h>
+#include <signal.h>
+#include <libvmi/libvmi.h>
+#include <libvmi/libvmi.h>
```
tklengyel: Double includes

examples/wait-for-domain-example.c
```diff
@@ -0,0 +1,88 @@
...
+static int interrupted = 0;
+static void close_handler(int sig)
+{
+    interrupted = sig;
+}
+vmi_event_t interrupt_event = {0};
+
+event_response_t wait_for_domain(vmi_instance_t vmi, vmi_event_t *event)
+{
+    watch_domain_data_t *data = (watch_domain_data_t *)event->data;
```
tklengyel: event->data is user defined, you can't just use it like this. You have to create an event structure similar to the existing event types to convey extra information about the event

libvmi/core.c
```diff
@@ -544,6 +544,20 @@ status_t vmi_init(
...
+        if ( VMI_FAILURE == driver_light_init(_vmi, init_flags) ) {
+            if ( error )
+                *error = VMI_INIT_ERROR_DRIVER;
+            goto error_exit;
+        }
+
+        if ( init_flags == VMI_INIT_WAIT_FOR_DOMAIN ) {
+            /* we have all we need to wait for domains. Return if there is nothing else */
+            status = VMI_SUCCESS;
+            goto error_exit;
```
tklengyel: Going to error exit to return success looks wrong. You also can't just ignore all the other flags. Why are they necessarily incompatible? What if you are starting on an existing domain and just want to catch the event when it's deleted?

libvmi/driver/xen/xen_events.c
```diff
@@ -1878,8 +1993,11 @@ static
 status_t wait_for_event_or_timeout(vmi_instance_t vmi, unsigned long ms, bool *needs_unmasking)
 {
     xen_events_t *xe = xen_get_events(vmi);
-
-    switch ( poll(&xe->fd, 1, ms) ) {
+#ifdef HAVE_LIBXENSTORE
+    switch ( poll(xe->fd, 2, ms) ) {
```
tklengyel: Just because xenstore is enabled it doesn't mean domain watch is. Only poll on it if domain watch is explicitely enabled.

libvmi/libvmi.h
```diff
@@ -55,6 +56,8 @@ extern "C" {
 
 #define VMI_INIT_DOMAINUUID (1u << 3) /**< initialize using domain uuid */
 
+#define VMI_INIT_WAIT_FOR_DOMAIN (1u << 4) /**< initialize using a domain watcher */
```
tklengyel: Not sure this is a good name if we are looking for domain deletion events

libvmi/core.c
aisaila: I go to out only if VMI_INIT_WAIT_FOR_DOMAIN is set in the init_flags. If there are other flags then I go on

libvmi/libvmi.h
aisaila: Can you suggest a better name to go on?

libvmi/core.c
tklengyel: It still looks wrong to go to error_exit with a success code.

aisaila: Agreed but the error_exit checks the status. I can move the *vmi = _vmi and return success up there but I did not want to double the code.

libvmi/libvmi.h
tklengyel: How about `VMI_INIT_DOMAINWATCH`?

aisaila: Sure looks in check with all the other names

aisaila:
I've pushed a new version with the requested changes

libvmi/driver/xen/xen_events.c
```diff
@@ -1104,6 +1129,93 @@ status_t process_desc_access(vmi_instance_t vmi, vm_event_compat_t *vmec)
...
+                    unsigned int size = 0;
+
+                    snprintf(path, sizeof( path ), "/local/domain/%d/vm", dominfo.domid);
+                    tmp = xen->libxsw.xs_read(xen->xshandle, XBT_NULL, path, &size);
+                    if (tmp || (sizeof(tmp) > 4)) {
+                        uint32_t *domid = malloc(sizeof(uint32_t));
+                        char *uuid = strdup(tmp + 4);
+
+                        *domid = dominfo.domid;
+                        g_tree_insert(xen->domains, domid, uuid);
```
tklengyel: Could you explain why we need that tree to store the domain's in? Why not just pass the event straight through to the registered event handler?

tklengyel: Never mind, I see it's just to look up the uuid from instead of doing a read from Xenstore. I don't think that's necessary. Is that lookup going to be that much faster then doing the read from xenstore?

razvan-cojocaru: It's probably faster, but that's not the main reason. The problem is that on `@introduceDomain` we can just call `xs_is_domain_introduced()` and figure out which domains are new. But on `@releaseDomain`, the domain is gone, so iterating over all the available domains to find the one that has finished is pointless (only the active domains will show up).

tklengyel: Gotcha, makes sense.

tklengyel:
Clang spotted a bug in the PR on travis

aisaila:
It's all fine now

libvmi/driver/driver_interface.h
```diff
@@ -203,5 +209,9 @@ status_t driver_init_vmi(
     uint32_t init_flags,
     vmi_init_data_t *init_data);
 
+status_t driver_light_init(
```
tklengyel: Could this not be folded into `driver_init`? I'm not a fan of introducing yet another obscure driver init function. It's already quite opaque having two separate ones.

aisaila: Yes, I can move that code at the end of driver_init

aisaila: On a second look that would involve to init xen_events_t * xe in xen.c and not in xen_events.c. If that is ok then I will move it

tklengyel: Hm, no, that doesn't sound right. How about moving it into `xen_init_events`?

aisaila: There was a problem with that in the idea that xen_init_events() needs to have xen_init_vmi() called before(or it is called by) so that is why I opted for a different function because the code would have changed very much

tklengyel: Right. In that case I guess just pick a better name for this, instead of driver_light_init make it driver_domainwatch_init

aisaila: I've changed the name and pushed

libvmi/driver/xen/xen_events.c
```diff
@@ -1878,8 +1990,18 @@ static
 status_t wait_for_event_or_timeout(vmi_instance_t vmi, unsigned long ms, bool *needs_unmasking)
 {
     xen_events_t *xe = xen_get_events(vmi);
+#ifdef HAVE_LIBXENSTORE
+    int rc = 0;
+
+    if ( vmi->init_flags & VMI_INIT_DOMAINWATCH )
```
tklengyel: I would prefer if we didn't have this bitmask in here, but rather `xe` had a field that recorded the size of the `fd` array, and we passed that in to `poll`. This is somewhat in the fast-path during events so adding an extra bitmask and if to every call to this function is going to add unnecessary overhead.

aisaila: Is a new field mandatory? I think we can use a macro for COUNT_OF or something like 
switch ( poll(xe->fd, sizeof(xe->fd)/sizeof(xe->fd[0]), ms) )

tklengyel: Well, afaict the array is always of size two but you only use the first array index with events without domain watcher. So for what you propose you would have to grow the array size dinamically.

aisaila: I've added the fd_size field in the xe structure

tklengyel:

tklengyel:
@aisaila Please update to resolve the conflicts with the now merged other PR. Thanks!

aisaila:
I've rebased with the latest master

smaresca:
@tklengyel I'm taking one last read through now

tklengyel:
@smaresca If there are no objections I'm merging this tomorrow

smaresca:
@tklengyel no objections; I'll just merge now.  

Applicability to other hypervisors was foremost on my mind, but as built, it's good enough for most cases. The only item I can foresee is broadening of the bool in watch_domain_event_t if we ever need to represent more states.

PR #739, opened 2019-03-13
Xen altp2m_idx setting correct value

ufo95:
While playing with drakvuf @sergej-proskurin and I stumbled upon this issue.

This commit fixes an issue in process_request in the xen_driver. When altp2m is active vmec->altp2m_idx does not remain unchanged as would be expected, but instead is set to 128. This commits fixes that.


tklengyel:
Ah, yes, I've seen that popping up in the drakvuf logs too but never chased down the issue :) Thanks!

tklengyel:

Issue #734, opened 2019-03-04
CMake minimum version should be 3.1


Wenzel:
In our CMakeLists we are using `target_sources` which is only available in CMake `3.1`.

This has been catched by https://github.com/KVM-VMI/kvm-vmi/issues/32#issuecomment-469290638

@Bondey can you make this PR ?

Bondey:
sure, done here: https://github.com/libvmi/libvmi/pull/735

sry if it's not the correct way

PR #735, opened 2019-03-04
Set CMake minimum version to 3.1 on Readme

Bondey:

PR #741, opened 2019-03-15
Set CMake min version to 3.1

Bondey:
Needed to add support for the command "target_sources" of CMake


Wenzel:
Fix #734

Wenzel:
@tklengyel can we merge this ?
Also #735

tklengyel:

PR #742, opened 2019-03-20
Fix cache bugs

mleinhos:
The glib function `g_hash_table_insert()` does not return a success/failure code. Rather it returns whether the key was previously in the table. However, `cache.c` treats the return code as though it could indicate failure. This PR fixes that oversight.

For confirmation, see
https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-insert

### DIFF
--- a/libvmi/cache.c
+++ b/libvmi/cache.c
@@ -35,6 +35,7 @@
 #include <time.h>
 #include <string.h>
 
+#include "libvmi.h"
 #include "private.h"
 #include "glib_compat.h"
 
@@ -733,10 +734,17 @@ vmi_rvacache_add(
     addr_t rva,
     char *sym)
 {
+    addr_t pgd = 0;
+
     if (!vmi)
         return;
 
-    return rva_cache_set(vmi, base_addr, pid, rva, sym);
+    if (VMI_SUCCESS != vmi_pid_to_dtb(vmi, pid, &pgd)) {
+        dbprint(VMI_DEBUG_SYMCACHE, "--SYM cache failed to find base for PID %u\n", pid);
+        return;
+    }
+
+    return rva_cache_set(vmi, base_addr, pgd, rva, sym);
 }
 
 void
--- a/libvmi/cache.c
+++ b/libvmi/cache.c
@@ -187,7 +187,6 @@ pid_cache_set(
 cleanup:
     g_free(key);
     g_free(entry);
-
 }
 
 status_t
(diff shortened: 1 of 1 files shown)


libvmi/cache.c
```diff
@@ -80,7 +81,7 @@ void key_128_init(vmi_instance_t vmi, key_128_t key, uint64_t low, uint64_t high
 
 key_128_t key_128_build (vmi_instance_t vmi, uint64_t low, uint64_t high)
 {
-    key_128_t key = (key_128_t) g_malloc0(sizeof(struct key_128));
+    key_128_t key = (key_128_t) g_malloc(sizeof(struct key_128));
```
tklengyel: Why the switch from `g_malloc0`?

mleinhos: Speed - we know that we're writing to each struct member. If you don't like this I can change it back.

tklengyel: Fair enough

tklengyel:

mleinhos:
I'm unsure what to do about the clang warnings like:
```
/home/travis/build/libvmi/libvmi/libvmi/cache.c:357:12: warning: Potential leak of memory pointed to by 'key'
    (void) g_hash_table_insert_compat(symbol_table, sym_dup, entry);
```

The function cannot fail and glib assumes ownership over the key and value memory.

tklengyel:
Not sure what it refers to but something in this PR changes something that makes that error appear, so have to hunt it down before we can merge this.

mleinhos:
I'll look into it. I don't have the local build env to do the checks, so I may be hitting your infrastructure a couple times until I run this to ground.

tklengyel:
Pcid refers to a tlb tag so that wouldn't be a good name. Does the pid not get translated to dtb in between those functions?

mleinhos:
The PID is not translated. Maybe PCID is a bad name, but perhaps something like ASID (address space ID) would suffice. Really, it doesn't have to be the PID or the PDB, so long as the user is guaranteeing uniqueness on a per-address space basis.

tklengyel:
Looks to me that `vmi_rvacache_add` is buggy and should actually have translated the pid to a dtb before calling `rva_cache_set`. That would be the solution. Keeping the name as dtb I think is fine cause that explicitly says that the address space is associated with that table that start at the address the dtb points to.

tklengyel:
@libvmi-jenkins Retest this please

PR #747, opened 2019-03-26
cmake: disable testing by default

### LINKED ISSUE #745: Enable check build dependency only when testing is enabled
related to #744 

Testing option is enabled by default:
https://github.com/libvmi/libvmi/blob/master/CMakeLists.txt#L56

check is added here as a dependency:
https://github.com/libvmi/libvmi/blob/master/tests/CMakeLists.txt#L1

The fix should be to disable testing by default.
What do you think @tklengyel ?

Wenzel:
fixes #745

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -217,7 +217,7 @@ matrix:
                 - libfuse-dev
       script:
         - mkdir build && cd build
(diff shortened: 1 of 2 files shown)


tklengyel:
Why are you removing check from travis?

Wenzel:
Because check becomes an optional dependency and is not needed anymore to build libvmi.
We are only using make test in Jenkins.

Wenzel:
As you wish, i updated the PR

Wenzel:
@tklengyel all green

PR #754, opened 2019-04-05
Add possibility of debug level configuration via cmake option

disaykin:


tklengyel:

PR #761, opened 2019-04-30
Fix domainwatch init

tklengyel:
Spotted by coverity


smaresca:
ah, very good. hooray for coverity

PR #762, opened 2019-05-02
Add vmi_request_page_fault to libvmi

aisaila:
Support for injecting page faults

### DIFF
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -2505,6 +2505,39 @@ xen_set_vcpureg_arm(
 #endif
 
 status_t
+xen_request_page_fault(
+    vmi_instance_t vmi,
+    unsigned long vcpu,
+    uint64_t virtual_address,
+    uint32_t error_code)
+{
+#ifdef ENABLE_SAFETY_CHECKS
+#if defined(I386) || defined(X86_64)
+    uint64_t value;
+
+    if (xen_get_vcpureg_hvm(vmi, &value, SS_ARBYTES, vcpu) == VMI_FAILURE)
+        return VMI_FAILURE;
+
+    if (((value >> 5) & 3) != 3) {
+        errprint("%s not in user mode\n", __FUNCTION__);
+        return VMI_FAILURE;
+    }
+#endif
+#endif
+    xen_instance_t *xen = xen_get_instance(vmi);
+    int rc = xen->libxcw.xc_hvm_inject_trap(xen->xchandle, xen->domainid,
+                                            vcpu, X86_TRAP_page_fault,
+                                            X86_TRAP_hw_exc,
+                                            error_code, 0, virtual_address);
+    if (rc < 0) {
+        errprint("%s error %d injecting page fault exception\n", __FUNCTION__, rc);
+        return VMI_FAILURE;
+    }
+
+    return VMI_SUCCESS;
+}
+
+status_t
 xen_get_vcpumtrr(
     vmi_instance_t vmi,
     mtrr_regs_t *hwMtrr,
--- a/libvmi/accessors.c
+++ b/libvmi/accessors.c
@@ -230,6 +230,21 @@ vmi_get_num_vcpus(
 }
 
 status_t
+vmi_request_page_fault(
+    vmi_instance_t vmi,
+    unsigned long vcpu,
+    uint64_t virtual_address,
+    uint32_t error_code)
+{
+#ifdef ENABLE_SAFETY_CHECKS
+    if (!vmi)
+        return 0;
+#endif
+
+    return driver_request_page_fault(vmi, vcpu, virtual_address, error_code);
+}
+
+status_t
 vmi_get_tsc_info(
     vmi_instance_t vmi,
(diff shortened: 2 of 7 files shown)


libvmi/driver/xen/xen.c
```diff
@@ -788,6 +788,27 @@ xen_get_memsize(
...
+status_t
+xen_request_page_fault(
+    vmi_instance_t vmi,
+    unsigned long vcpu,
+    uint64_t virtual_address,
+    uint32_t error_code)
+{
+
+    xen_instance_t *xen = xen_get_instance(vmi);
+    int rc = xen->libxcw.xc_hvm_inject_trap(xen->xchandle, xen->domainid,
```
tklengyel: Please add a safety check (behind the ifdef) that grabs cs_sel and checks the CPL of the vCPU to verify it's safe to call this function.

aisaila: Do you mean ss_sel? It looks like then xen check is done with ss_sel 
  unsigned int vmx_get_cpl(void)
  {
      unsigned long attr;
  
      __vmread(GUEST_SS_AR_BYTES, &attr);
  
      return (attr >> 5) & 3;
  }
I guess that the guest_mode check is  if((ss_attr >> 5) == 3)
Is this what you want?

tklengyel: That's actually the `RPL` you reference. In this context we should check the `CPL`. See https://css.csail.mit.edu/6.858/2014/readings/i386/s06_03.htm

razvan-cojocaru: Are we discussing the use of CS vs. SS? If so, the Intel SDM says in "5.5 PRIVILEGE LEVELS":
```
Current privilege level (CPL) — The CPL is the privilege level of the currently executing program or task. It is stored in bits 0 and 1 of the CS and SS segment registers. Normally, the CPL is equal to the privilege level of the code segment from which instructions are being fetched. The processor changes the CPL when program control is transferred to a code segment with a different privilege level. The CPL is treated slightly differently when accessing conforming code segments. Conforming code segments can be accessed from any privilege level that is equal to or numerically greater (less privileged) than the DPL of the conforming code segment. Also, the CPL is not changed when the processor accesses a conforming code segment that has a different privilege level than the CPL.
```
https://software.intel.com/sites/default/files/managed/a4/60/325384-sdm-vol-3abcd.pdf
Also, Jan's been quite adamant about using SS:
https://lists.xenproject.org/archives/html/xen-devel/2014-07/msg00415.html
And the KVM developers have switched from CS to SS (if I'm reading their patch correctly):
https://lore.kernel.org/patchwork/patch/464719/

tklengyel: Yay for x86 being straight forward :) AFAIU cs_sel and the ss attr CPL bits should be equal for most cases, but in corner-cases it's safer to use ss attr bits. TIL

tklengyel:

Wenzel:
Thanks for this PR, I actually needed this API for one of my projects.
Quick question: What about kernel mode pagefault injection ?

I know there are cases where it's just not possible, example on Windows:
- you are not supposed to deal with a pagefault if your `IRQL >= 2`
- you are not supposed to trigger a pagefault in a nonpaged memory area

both of them will lead to a BSOD.

but nontheless, it still a useful API.

Have you considered this possibility and the corner cases to deal with ?
Restricting this API to usermode is meant to be a safeguard for the API user ?

aisaila:
I've added safety checks for the user mode in xen_request_page_fault()

tklengyel:
@smaresca new API is being added, please review and hit merge when you get a chance

smaresca:
I have been following this PR with interest - it is a very helpful contribution. Much appreciated @aisaila 

@tklengyel almost done, LGTM so far. Will merge when complete

libvmi/driver/xen/xen.c
```diff
@@ -2504,6 +2504,39 @@ xen_set_vcpureg_arm(
...
+    uint32_t error_code)
+{
+#ifdef ENABLE_SAFETY_CHECKS
+#if defined(I386) || defined(X86_64)
+    uint64_t value;
+
+    if (xen_get_vcpureg_hvm(vmi, &value, SS_ARBYTES, vcpu) == VMI_FAILURE)
+        return VMI_FAILURE;
+
+    if (((value >> 5) & 3) != 3) {
```
thatreguy: Could you please explain this ?

thatreguy: Sorry this is post merge.

smaresca: @thatreguy nothing wrong with commenting after a merge in my view - sometimes issues are noticed only after the fact, and we might as well place those comments with the related change.

Anyway, I'm generally a fan of macros, but during review I saw this as fairly succinct and clear in context due to the proximity of the register and the error message.  We're certainly open to contributions if you had something in mind to improve documentation/readability.

Wenzel:
@aisaila , @tklengyel I still my question regarding pagefault injection in the kernel.
Can I have your opinion on the matter ? :slightly_smiling_face: 

Thanks.

PR #763, opened 2019-05-10
Fix init for Windows 10

tklengyel:
The function `init_from_rekall_profile_real` has been very messy. There are 4 methods in it to find the KPCR for Windows, but any error in any of them results in bailing out from initializing instead of trying any of the other methods still available. I split out the 4 methods into separate functions so that each can be tried in succession.

### DIFF
--- a/libvmi/os/windows/core.c
+++ b/libvmi/os/windows/core.c
@@ -1,4 +1,3 @@
-
 /* The LibVMI Library is an introspection library that simplifies access to
  * memory in a target virtual machine or in a file containing a dump of
  * a system's physical memory.  LibVMI is based on the XenAccess Library.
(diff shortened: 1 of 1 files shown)


smaresca:
+1000

libvmi/os/windows/core.c
```diff
@@ -746,129 +746,161 @@ find_windows_version_from_rekall_profile(vmi_instance_t vmi)
...
+            dbprint(VMI_DEBUG_MISC, "** driver_get_vcpureg(..) failed.\n");
+            goto done;
+        }
+
+        if ( VMI_SUCCESS == kpcr_find1(vmi, windows, kpcr_reg) ) {}
+        else if ( VMI_SUCCESS == kpcr_find2(vmi, windows) ) {}
+        else if ( VMI_SUCCESS == kpcr_find3(vmi, windows) ) {}
+        else if ( VMI_SUCCESS == kpcr_find4(vmi, windows, kpcr_reg) ) {}
+        else goto done;
+
```
smaresca: the above is a bit odd, though I see why you did it

tklengyel: Yea, it's a bit odd from a C perspective but I find it quite readable `¯\_(ツ)_/¯`

tklengyel: The other option I was contemplating is adding a `kpcr_found` goto label, but I think this is nicer.

PR #767, opened 2019-05-17
Fix unused variable for release builds in domain watcher example

skvl:


tklengyel:

PR #768, opened 2019-05-17
make domain parameter const in vmi_init as well as vmi_init_complete

rageagainsthepc:
This change simplifies the usage of libvmi in conjunction with C++ `std::string`. Otherwise it would be necessary to cast `const char*` to `char*` when passing the domain name.

### DIFF
--- a/libvmi/core.c
+++ b/libvmi/core.c
@@ -291,7 +291,7 @@ set_image_type_for_file(
 static status_t
 set_id_and_name(
     vmi_instance_t vmi,
-    void *domain)
+    const void *domain)
 {
     const char *name = NULL;
     uint64_t id = 0;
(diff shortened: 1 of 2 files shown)


doeik:
It ain't much but it's honest work ;)

tklengyel:

tklengyel:
Thanks for the PR. For this to be merged you should fix some of the other functions as well where the now const variable is being passed without the const qualifier. For example, `vmi_get_access_mode` and `set_id_and_name`. There may be other functions as well, so please go check where that variable is being passed.

doeik:
I have updated the PR. Could not find any other functions than those you have mentioned before though.

tklengyel:
@smaresca: LGTM

rageagainsthepc:
Hi, I just wanted to check in and ask if you guys are still considering this PR? Any feedback would be appreciated.

tklengyel:
It is ready for merging, just waiting for @smaresca to hit merge or give a +1

rageagainsthepc:
Ok, thanks for the info

smaresca:
@rageagainsthepc sorry about that. Looks good to me as well, merging now.

PR #774, opened 2019-05-23
use CMAKE_CURRENT_SOURCE_DIR for module path

rageagainsthepc:
CMAKE_SOURCE_DIR does not work well if libvmi is added to another cmake file as a sub project


tklengyel:

PR #775, opened 2019-05-24
make libvmi.la generation optional

rageagainsthepc:
This change makes the generation of libvmi.la optional. Otherwise cmake fails if static library building is disabled.

### DIFF
--- a/libvmi/libtool_template.in
+++ b/libvmi/libtool_template.in
@@ -11,7 +11,7 @@ dlname='$<TARGET_SONAME_FILE_NAME:vmi_shared>'
 library_names='$<TARGET_FILE_NAME:vmi_shared> $<TARGET_SONAME_FILE_NAME:vmi_shared> libvmi.so'
 
 # The name of the static archive.
-old_library='$<TARGET_FILE_NAME:vmi_static>'
+old_library='$<$<BOOL:${ENABLE_STATIC}>:$<TARGET_FILE_NAME:vmi_static>>'
 
 # Linker flags that cannot go in dependency_libs.
 inherited_linker_flags=''


tklengyel:
Why does it fail if the static version isn't built?

rageagainsthepc:
CMake can't find the target `vmi_static`. This is the error output:
```
CMake Error at libvmi-github/libvmi/CMakeLists.txt:187 (file):
  Error evaluating generator expression:

    $<TARGET_FILE_NAME:vmi_static>

  No target "vmi_static"
```
because it is created only if `ENABLE_STATIC` is true:
```
if (ENABLE_STATIC)
    # create libvmi.a
    add_library (vmi_static STATIC $<TARGET_OBJECTS:vmi>)
    set_target_properties(vmi_static PROPERTIES OUTPUT_NAME "vmi")
endif ()
```
(same CMakeLists.txt in line 53)

tklengyel:
I don't think this is the correct solution. `libvmi.la` should be built even if only the dynamic or the static version of the library is built. /cc @Wenzel

rageagainsthepc:
You're right. Guess I was a little too hasty with this one. I'm gonna do some reading about how libtool works and then come back with a better solution (hopefully).

Wenzel:
The problem is that I'm using the `vmi_static` target directly in the libtool, without verification:
https://github.com/libvmi/libvmi/blob/master/libvmi/libtool_template.in#L14

We should use the conditioned variable or something like that to expand the content of `vmi_static` only it is defined.
@rageagainsthepc  can you look into this ?
Thanks.

rageagainsthepc:
This should do it.

PR #777, opened 2019-05-29
fix cmakedefines

rageagainsthepc:
There are quite a few cmake variables defined with `#cmakedefine01` which expands to either `VAR = 1` or `VAR = 0`. This means that these are always defined resulting in `#ifdef` macros always evaluating to true.

### DIFF
--- a/libvmi/cache.h
+++ b/libvmi/cache.h
@@ -38,7 +38,7 @@ gboolean key_128_equals(gconstpointer key1, gconstpointer key2);
 void key_128_init(vmi_instance_t vmi, key_128_t key, uint64_t low, uint64_t high);
 key_128_t key_128_build (vmi_instance_t vmi, uint64_t low, uint64_t high);
 
-#if ENABLE_ADDRESS_CACHE == 1
+#ifdef ENABLE_ADDRESS_CACHE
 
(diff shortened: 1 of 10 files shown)


rageagainsthepc:
I should mention that I chose the path with the least code changes. Just tell me if you prefer consistency instead ;)

tklengyel:
Yea, we should probably pick on or the other and stick with it to avoid this problem creeping back up. So +1 from me to go with consistency.

tklengyel:

rageagainsthepc:
Personally, I prefer `#cmakedefine` over `#cmakedefine01` but I'll leave the decision to you guys.

tklengyel:
`#cmakedefine` is fine, go with that

rageagainsthepc:
I have also removed some defines I couldn't find anywhere in the code.

Wenzel:
I can't remember why i picked cmakedefine01.

 Thanks for looking into it.

tklengyel:
Travis seems to be failing

rageagainsthepc:
Yes, but unfortunately I don't have time to look into it until monday. :/

rageagainsthepc:
Never mind. I think I've found it. Stupid mistake :D

PR #779, opened 2019-06-03
address implicit fallthrough warnings

rageagainsthepc:
Clang 8.0.0 complains about implicit fallthroughs. This should get the compiler off our backs.


tklengyel:

PR #781, opened 2019-06-03
address cast-function-type warnings

rageagainsthepc:
* `g_queue_foreach` expects a function that takes two parameters but `g_free` only takes one parameter
* `g_tree_new_full` expects a function that takes three parameters but `domains_compare` only provides two

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ include(GNUInstallDirs)
 #-----------------------------------------------------------------------------
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -g3 -O0")
-set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -Werror -O2")
+set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -Wno-cast-function-type -Werror -O2")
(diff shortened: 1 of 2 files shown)


tklengyel:
Casting to `GFunc` a function that uses viewer inputs is fine and adding an extra function just to explicitly call out that the second argument is ignored is not needed. It's just yet another indirection. Adding the third argument to `domains_compare` is fine though, I have no problem with that.

tklengyel:
Or is there an actual compiler warning about the `GFunc` cast now with some compiler? In that case I would be OK with getting this merged, but otherwise I don't see the point.

rageagainsthepc:
> Casting to `GFunc` a function that uses viewer inputs is fine and adding an extra function just to explicitly call out that the second argument is ignored is not needed.

I actually did not know that but it makes sense if you think about it. I also agree with you that unnecessary indirections are bad. However, as you've suspected, there exists a compiler warning for it in Clang 8.0.0. Maybe we could figure out another way to get rid of it?

rageagainsthepc:
We could use `-Wno-error=cast-function-type`. That way we would still see the warnings but compilation is not aborted for release builds.

rageagainsthepc:
I went with disabling the cast-function-type warning for release builds. This will emit `cc1: warning: unrecognized command line option ‘-Wno-cast-function-type’` for older compilers though but I think that's endurable.

rageagainsthepc:
Hi, what about this PR? I know that this might not be the nicest solution but it would be cool to be able to compile release builds with newer compilers. Got any suggestions for that?

tklengyel:

Issue #780, opened 2019-06-03
Possible conversion loss with key_128_hash function


rageagainsthepc:
https://github.com/libvmi/libvmi/blob/684ebcf24a942d5fa50466e84d1787d2e5e0fa5f/libvmi/cache.c#L280

`key_128_hash` returns a 64bit integer but `g_hash_table_new_full` expects a hash function that returns a 32bit integer. I'm wondering whether it is intended that the higher 32bit of the hash are discarded. If so I could provide a patch that suppresses the warning (`-Wcast-function-type`). Otherwise I would have to pass because I have little to no experience with hash functions.

tklengyel:
No, I don't think that truncating the hash was intended. It's anyone's guess whether that truncation is a problem - thus far it didn't really seem to have been - but to be safe I guess we should just further hash the 64-bit value with `g_int64_hash` to make it conform to the glib API more strictly.

rageagainsthepc:
I went ahead and created a pull request that implements your suggestion.

PR #782, opened 2019-06-03
make key_128_hash function return a 32 bit integer

rageagainsthepc:
This change addresses issue #780.


tklengyel:

PR #786, opened 2019-07-09
Implement vmi_mmap_guest feature - direct guest memory access

icedevml:
Implements `vmi_mmap_guest` function which allows to read given memory region directly from guest's memory. Works only with Xen driver (internally using `xc_map_foreign_pages`).

In case of broad memory ranges, it may happen that some virtual address ranges would be not mapped to any physical page. In such a case, a fake `zero_page_gpfn` is used as a placeholder, so you read "all-zeros" in such a case, thus still keeping the original alignment of memory.

### DIFF
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -584,7 +584,6 @@ xen_init_vmi(
 {
     status_t ret = VMI_FAILURE;
     xen_instance_t *xen = xen_get_instance(vmi);
-    unsigned char *memory = NULL;
     int rc;
 
     /* setup the info struct */
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -584,6 +584,7 @@ xen_init_vmi(
 {
     status_t ret = VMI_FAILURE;
     xen_instance_t *xen = xen_get_instance(vmi);
+    unsigned char *memory = NULL;
     int rc;
 
     /* setup the info struct */
@@ -634,6 +635,25 @@ xen_init_vmi(
     if ( vmi->vm_type >= PV32 && (xen->max_gpfn << XC_PAGE_SHIFT) < (xen->info.max_memkb * 1024)) {
         xen->max_gpfn = (xen->info.max_memkb * 1024) >> XC_PAGE_SHIFT;
     }
+
+    vmi->zero_page_gpfn = ++(xen->max_gpfn);
+
+    rc = xen->libxcw.xc_domain_populate_physmap_exact(xen->xchandle, xen->domainid, 1, 0, 0, &vmi->zero_page_gpfn);
+    if (rc < 0) {
+        errprint("Failed to populate physmap for additional page\n");
+        ret = VMI_FAILURE;
+        goto _bail;
+    }
+
+    memory = xen_get_memory_pfn(vmi, vmi->zero_page_gpfn, PROT_WRITE);
+    if (NULL == memory) {
+        errprint("Failed to xen_get_memory_pfn for additional page\n");
+        ret = VMI_FAILURE;
+        goto _bail;
+    }
+
+    memset(memory, 0, vmi->page_size);
+    xen_release_memory(vmi, memory, vmi->page_size);
 
     ret = xen_setup_live_mode(vmi);
 
@@ -2634,6 +2654,17 @@ xen_read_page(
     addr_t paddr = page << vmi->page_shift;
 
     return memory_cache_insert(vmi, paddr);
+}
+
+void *
+xen_mmap_guest(
+    vmi_instance_t vmi,
+    unsigned long *pfns,
+    unsigned int size)
+{
+    xen_instance_t *xen = xen_get_instance(vmi);
+    void *memory = xen->libxcw.xc_map_foreign_pages(xen->xchandle, xen->domainid, PROT_READ, pfns, size);
+    return memory;
 }
 
 status_t
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -636,25 +636,6 @@ xen_init_vmi(
         xen->max_gpfn = (xen->info.max_memkb * 1024) >> XC_PAGE_SHIFT;
     }
 
-    vmi->zero_page_gpfn = ++(xen->max_gpfn);
-
-    rc = xen->libxcw.xc_domain_populate_physmap_exact(xen->xchandle, xen->domainid, 1, 0, 0, &vmi->zero_page_gpfn);
-    if (rc < 0) {
-        errprint("Failed to populate physmap for additional page\n");
-        ret = VMI_FAILURE;
-        goto _bail;
-    }
-
-    memory = xen_get_memory_pfn(vmi, vmi->zero_page_gpfn, PROT_WRITE);
-    if (NULL == memory) {
-        errprint("Failed to xen_get_memory_pfn for additional page\n");
-        ret = VMI_FAILURE;
-        goto _bail;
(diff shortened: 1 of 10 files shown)


tklengyel:

tklengyel:
For the most part this looks good.

The only point I have concerns about is the mapping in the zero page to the guest to use as a "pad" when the requested virtual address in not valid in the target context. I think that fundamentally makes it hard for a user to differentiate when the mapping worked and the page is actually just filled with zeroes vs the virtual address is invalid and the magic zero page is mapped in. Also, since this page is now accessible by the guest, nothing guarantees that it remains "0".

I understand your goal is hide the holes in the memory space in case you are requesting a chunk of memory of which only a portion may actually be mapped and this kindof hacks around it. I think the function should take input an array of `void*` with the size of the requested memory block (in terms of pages). The LibVMI would map in each requested pfn in the range when possible to its appropriate pointer in the array. The caller then could loop through that array and check if the pointer is NULL, in which case you can add your padding. This is a bit more logic on the caller side but I think it's more robust. What do you think?

icedevml:
We use `PROT_READ` when calling Xen, so at the current state the mapped memory is read only:
```cpp
xen->libxcw.xc_map_foreign_pages(xen->xchandle, xen->domainid, PROT_READ, pfns, size);
```
thus you can't really change the contents of the "zero page". This is probably bad, because I could imagine that some users would want to modify the mapped area.

From our perspective, such a hack is handy if we would like to use something like Yara's [yr_rules_scan_mem](https://yara.readthedocs.io/en/v3.7.0/capi.html#c.yr_rules_scan_mem) to check the memory contents against some signature.

Maybe splitting it into two different functions would be OK? These functions may potentially call some common abstraction.

```cpp
// returns array output_ptrs[num_pages] where each ptr corresponds to a single page, non-existent pages left as NULLs
status_t vmi_mmap_guest(vmi_instance_t vmi, vmi_pid_t pid, addr_t vaddr, size_t count, void **output_ptrs, int prot);

// returns a single pointer, guaranteed continous buffer, non-existent pages faked
status_t vmi_mmap_guest_readonly_continous(vmi_instance_t vmi, vmi_pid_t pid, addr_t vaddr, size_t count, void **block_ptr);
```

I think that with proper docs the users would not be confused as much.

smaresca:
@icedevml I support that proposal (splitting into two functions).  I agree with you that there are major integration benefits from providing the continuous region, especially if the function behavior is clearly documented.  There are many, many tools written with the same assumption as that Yara function, and accommodating that convention will facilitate the use of such tools/libraries.

Is there any alternative to zero padding that might allow inference of invalid regions? (e.g., filling instead with a highly-nonrandom yet detectable value? perhaps setting a flag? I would prefer not to abuse errno for such things, though)

tklengyel:
> thus you can't really change the contents of the "zero page"

I meant that you map the page _into the guest_ at https://github.com/libvmi/libvmi/pull/786/files#diff-6751e1b2087817e662fcf1f47cbdb303R641 so the guest can modify it at that point. This was also an issue that lead to XSA-99. That's what I meant.

icedevml:
Ah, you are absolutely right. Do you see any way to overcome these problems?

tklengyel:
The only way to make that make that work "safe" is to pause the guest while that zero page is mapped in or to do additional EPT protection to that page. Neither options are particularly good for this API. I would probably try to find a way to make that zero page something that's only on the host side and plug it in when needed. 

I would suggest to try to find a way to take the pages Xen can get us which will be mapped non-continuously in the process' virtual address space and take the zero page's address and try to make Linux construct for you a continuous memory mapping with these. Perhaps experimenting with `mremap` would worth a try to make it appear that that the memory is continuous?

icedevml:
Thanks for your hints. I'm going to experiment with the abovementioned.

tklengyel:
@icedevml I'm also wondering, did you do a performance comparison of doing the memcpy into a local buffer (add adding the padding there) vs this direct mapping done by the hypervisor? In my experience the difference between the two is barely noticeable in most use-cases.

icedevml:
i'm trying to PoC something with `mremap` but it seems that it doesn't work well with foreign memory mapped by Xen. Even though it succeeds, any attempt to use the new pointer causes SIGBUS.

I don't want to spend too much time investigating related issues, so maybe let's stick to @tklengyel [initial proposition](https://github.com/libvmi/libvmi/pull/786#issuecomment-510568060) completely (at least now). I will also try to make some benchmarks later.

icedevml:
Okay, drafted new version of `vmi_mmap_guest`, now focusing on some benchmarks.

libvmi/driver/xen/xen.c
```diff
@@ -2636,6 +2636,17 @@ xen_read_page(
...
 }
 
+void *
+xen_mmap_guest(
+    vmi_instance_t vmi,
+    unsigned long *pfns,
+    unsigned int size)
+{
+    xen_instance_t *xen = xen_get_instance(vmi);
+    void *memory = xen->libxcw.xc_map_foreign_pages(xen->xchandle, xen->domainid, PROT_READ, pfns, size);
```
tklengyel: Just return `xen->libxcw.xc_map_foreign_pages`, no need for the local variable here

libvmi/libvmi.h
```diff
@@ -1201,6 +1201,26 @@ status_t vmi_read_va(
...
+ * @param[in] vmi LibVMI instance
+ * @param[in] pid Pid of the virtual address space
+ * @param[in] vaddr Virtual address to read from
+ * @param[in] num_pages Number of guest pages to be mapped
+ * @param[out] access_ptrs Array of size [num_pages] containing pointers to the respective guest's pages
+ */
+status_t vmi_mmap_guest(
+    vmi_instance_t vmi,
+    vmi_pid_t pid,
+    addr_t vaddr,
```
tklengyel: How about taking an `access_context_t` here instead? That would allow a caller to directly specify a physical address making the API useful in scenarios where paging has not been yet enabled (or there is no OS initialized yet).

icedevml:
I did some very brief measurements using Windows 7 SP 1 Virtual Machine and DRAKVUF as a LibVMI consumer. I've modified DRAKVUF a little bit, so it saves a memory dump of process' main module when it makes `NtTerminateProcess` syscall. Two implementations:

* "old idea" - with the originally-proposed `vmi_mmap_guest` implementation which returns a single pointer, filling holes by zero page hack; caller simply puts everything to a file with `fwrite` and calls `munmap` afterwards
* "new idea" - with the current `vmi_mmap_guest` implementation which returns an array of pointers; caller allocates a huge buffer using `calloc`, calls function and then iterates through the resulting pointers, `memcpy`-ing pages into a huge buffer and `munmap`-ing them, finally puts everything to a file with `fwrite` (note that this buffer is not extremely necessary in this particular case, but I wanted test to be fair around providing continous memory)

Some numbers that I've captured:

measurement | cur. idea (A) | old idea (B)
-- | -- | --
1 | 409415 [ns] | 919688 [ns]
2 | 443784 | 904619 
3 | 432359 | 904060
4 | 446238 | 915532
5 | 448290 | 921740
6 | 440664 | 926193
7 | 461094 | 920405
8 | 431051 | 921796
**avg** | **439112 [ns]** | **916754 [ns]**

The new implementation is 2.2 times faster even taking into account that the caller is doing more work because of `memcpy`-ing stuff. Funny thing is that only 15 pages out of 166 were ever available (well, I guess this is how virtual memory works).

tklengyel:
Thanks for doing those benchmarks, that's quite an interesting find. Happy to see it payed off and we have more confidence in that this is the right implementation to go with :)

PR #787, opened 2019-07-14
Fixed iteration over Linux tasks.

mleinhos:
Iteration worked fine, unless requested entry wasn't found. In that case it got stuck in an infinite loop. This commit fixes that behavior.

### DIFF
--- a/libvmi/os/linux/memory.c
+++ b/libvmi/os/linux/memory.c
@@ -31,13 +31,14 @@
 #include "os/linux/linux.h"
 #include "driver/driver_wrapper.h"
 
-/* finds the task struct for a given pid */
+/* finds the task struct for a given pid, with init_task pointing to actual task */
 static addr_t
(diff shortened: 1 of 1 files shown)


tklengyel:

libvmi/os/linux/memory.c
```diff
@@ -54,25 +55,28 @@ linux_get_taskstruct_addr_from_pid(
...
      */
-    next_process = vmi->init_task;
-    list_head = next_process;
+    curr_proc = vmi->init_task;
+    vmi_read_addr_va(vmi, curr_proc + tasks_offset, 0, &curr_entry);
+
+    list_head = curr_entry;
 
     do {
-        vmi_read_32_va(vmi, next_process + pid_offset, 0, (uint32_t*)&task_pid);
```
mleinhos: As this line indicates, next_process actually pointed to a task_struct, not to a list_head as claimed by the above (removed) comment.

PR #790, opened 2019-07-22
latest bareflank

ssentanoe:
support latest bareflank, the name2domid and vice versa is still a temporary solution, in this way bareflank handler knows who monitor whom. And for now, we need to use stdout for error printing so that boxy can print the error message (from domU)


tklengyel:

PR #793, opened 2019-07-22
Add example on using instruction emulation as callback response

mtarral:
This PR adds an API example of using breakpoints based on opcode emulation instead of singlestepping.


tklengyel:

examples/breakpoint-emulate-example.c
```diff
@@ -0,0 +1,215 @@
...
+
+struct cb_data {
+    char *symbol;
+    addr_t vaddr;
+    emul_insn_t emul;
+};
+
+event_response_t int3_cb(vmi_instance_t vmi, vmi_event_t *event)
+{
+    vmi = vmi;
```
tklengyel: Please change this to `(void)vmi;` to silence warnings generated on clang-scan

mtarral: Done

PR #794, opened 2019-07-23
Feature summary

mtarral:
This patches adds description for the packages and a feature summary when `cmake` is done.

It also disables the features on which dependencies couldn't be satisfied.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,11 @@ endif()
 #-----------------------------------------------------------------------------
 include(FeatureSummary)
 find_package(PkgConfig REQUIRED)
+set_package_properties(PkgConfig PROPERTIES
+    DESCRIPTION "Helper tool to compile applications and libraries"
+    URL "https://www.freedesktop.org/wiki/Software/pkg-config/"
+    TYPE REQUIRED
+    PURPOSE "Find and configure multiple LibVMI dependencies")
 pkg_search_module(GLIB REQUIRED glib-2.0)
 # cleanup GLIB_LDFLAGS (remove -l prefix)
 string(REGEX REPLACE "-l" "" GLIB_LDFLAGS ${GLIB_LDFLAGS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,7 @@ endif()
 #-----------------------------------------------------------------------------
 #                              DEPENDENCIES
 #-----------------------------------------------------------------------------
+include(FeatureSummary)
 find_package(PkgConfig REQUIRED)
 pkg_search_module(GLIB REQUIRED glib-2.0)
 # cleanup GLIB_LDFLAGS (remove -l prefix)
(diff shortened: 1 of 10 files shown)


mtarral:
The test failing is `without xenstore`

I found out that in the actual build system, explicitely disabling Xenstore never worked, because we were configuring `config.h.in` before checking this option in `libvmi/CMakeLists.txt`, and only after the `find_package(Xenstore)` directive, which is responsible for setting `HAVE_XENSTORE_H` for example.

I'm working on it and see if I can understand the issue.

mtarral:
@tklengyel Libvmi Xen driver doesn't compile without Xenstore.

This is an issue that should have been catched by Travis when @aisaila was making PR to introduce the watch domain feature, but it didn't, the build system had a bug (that this PR is fixing now btw).

Using `INTRODUCE_TOKEN`
https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L574

Conditional compilation on Xenstore that brings `INTRODUCE_TOKEN`
https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen.h

tklengyel:
Yes, the domain watcher relies on Xenstore. It has to be wrapped in ifdefs and not compiled if without xenstore is specified.

mtarral:
The build is still failing.
I missed 2 functions in my patch:
https://travis-ci.org/libvmi/libvmi/jobs/566499643#L436

- `xen_events.c:process_domain_watch()`: https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L1154
- `xen_events.c:check_domain_shutdown`: https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L1134

@tklengyel , if you can add `ifdefs` to protect these functions, i can rebase this PR.

tklengyel:
How did that pass the CI in the first place in the original PR? I'm just going to revert it.

mtarral:
> How did that pass the CI in the first place in the original PR?
The actual build system has a bug, `HAVE_XENSTORE` is always set.
This is fixed by this PR, where we moved the detection of XenStore when we have to configure the Xen driver build, and we test beforehand if `ENABLE_XENSTORE` is set.
https://github.com/libvmi/libvmi/pull/794/files#diff-68b5d5b58f7a676bca6caf17aeb45a34R162

mtarral:
I updated PR this only include the `feature_summary` patches.

You can check
- description
- type (required, recommended, optional)

A more general question about the build system behavior:
Do you want to disable features when we detect a missing dependency, or adopt a simpler and cleaner behavior to simply mark the `find_package(.... )` as `REQUIRED` and stop the configuration ?
~~~CMake
if (ENABLE_CONFIGFILE)
    find_package(FLEX)
    find_package(BISON)
    if (NOT FLEX_FOUND OR NOT BISON_FOUND)
        set(ENABLE_CONFIGFILE OFF CACHE BOOL "Enable config file" FORCE)
        message(WARNING "Cannot find flex or bison: config file parsing will be
        disabled")
else ()
~~~

I'm asking this because it lead to confusion from a user perspective, where the user explicitely asked for `-DENABLE_CONFIGFILE=ON` for example, and he might not have seen the warning displayed, and would have expected an error in case of missing dependency.

tklengyel:

tklengyel:
Why are rekall profiles being disabled on Travis under cmake? `libjson-c-dev` is installed.

tklengyel:
Same for `vmifs`, libfuse-dev is installed.

mtarral:
> Why are rekall profiles being disabled on Travis under cmake? libjson-c-dev is installed.

The feature is enabled, but i put the wrong name in the feature summary declaration:
~~~CMake
add_feature_info(ENABLE_PROFILES ENABLE_PROFILES "Enable Rekall's JSON profiles")
~~~
which should be
~~~CMake
add_feature_info(REKALL_PROFILES REKALL_PROFILES "Enable Rekall's JSON profiles")
~~~

> Same for vmifs, libfuse-dev is installed.

`ENABLE_VMIFS` is disabled by default.

I was looking at the original PR `switch to cmake` because I remembered that you asked for this behavior, but looking back at the discussion today I believe I misinterpreted some comments.
https://github.com/libvmi/libvmi/pull/674#issuecomment-431684493

I can add 2 commits to fix the REKALL_PROFILES in the feature summary and VMIFS enabled by default.

mtarral:
@tklengyel the release build fails, now that it compiles vmifs:
~~~
/home/travis/build/libvmi/libvmi/tools/vmifs/vmifs.c:90:16: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
445     if (offset < memsize && size) {
446                ^
447/home/travis/build/libvmi/libvmi/tools/vmifs/vmifs.c: At top level:
448cc1: error: unrecognized command line option ‘-Wno-cast-function-type’ [-Werror]
~~~

can you quickfix `vmifs.c` ?

tklengyel:
@mtarral the master branch compiles just fine on Travis. If it breaks on this PR then it's because of the changes this PR is making.

mtarral:
> If it breaks on this PR then it's because of the changes this PR is making.

Yes, because this PR enables the compilation of `vmifs` by default now, contrary to the master branch:
https://github.com/libvmi/libvmi/blob/master/CMakeLists.txt#L62

And it doesn't compile with RELEASE flags

mtarral:
> If the cmake system has bugs regarding vmifs I'm not touching that.
Merge state

CMake introduces the notion of Debug and Release builds.
Release flags:
~~~CMake
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -Wno-cast-function-type -Werror -O2")
~~~

--> so `vmifs.c` simply doesn't compile with `-Werror`.

The flag is not enforced by automake:

tklengyel:
I see, in that case please just add the fix for vmifs in this PR

tklengyel:
LGTM

PR #795, opened 2019-07-31
xen: domain watcher feature depends on xenstore

mtarral:
This PR fixes an issue found by #794 .

The domain watcher feature depends on `libxenstore` and shouldn't be compiled if this library is not found on the system.


tklengyel:

Issue #796, opened 2019-08-01
Safety check in xen_request_page_fault is too restrictive


icedevml:
The [xen_request_page_fault](https://github.com/libvmi/libvmi/blob/fd34c6f4faed34e8eccd20be01d03ce6967b3bfb/libvmi/driver/xen/xen.c#L2512) method has too restrictive safety check that doesn't allow to cause a page fault in kernel mode, even though it would be valid (in some situations).

```cpp
#ifdef ENABLE_SAFETY_CHECKS
#if defined(I386) || defined(X86_64)
    uint64_t value;

    if (xen_get_vcpureg_hvm(vmi, &value, SS_ARBYTES, vcpu) == VMI_FAILURE)
        return VMI_FAILURE;

    if (((value >> 5) & 3) != 3) {
        errprint("%s not in user mode\n", __FUNCTION__);
        return VMI_FAILURE;
    }
#endif
#endif
```

A page fault can be succesfully triggered when we are inside a Windows system call handler (which is not user mode, in fact), just because the process' virtual memory is still "attached" in such a case.

I don't know how to resolve it other way than simply removing this check.

tklengyel:
Removing the safety check is fine, the reason I requested it to be added was because it was only tested to work while in usermode. If your tests confirm that under certain situation it works while in kernel mode, then it makes sense to remove the restriction.

tklengyel:
@libvmi-jenkins This is OK to test

icedevml:
A simple example is that you can cause a page fault being inside DRAKVUF's syscall hook and it should work perfectly fine. If you request for an invalid virtual address then the system will crash through `KiSystemServiceHandler` SEH, which is also fine.

Related PR #797

PR #797, opened 2019-08-01
Remove too restrictive check in xen_request_page_fault

icedevml:
Related discussion in #796.


tklengyel:

PR #803, opened 2019-08-07
Try to trick cmake into defining things correctly for Xenstore

tklengyel:


mtarral:
So you need a couple of things to fix Xenstore in the build system:
remove the detection early on:
https://github.com/libvmi/libvmi/pull/794/files#diff-af3b638bc2a3e6c650974192a53c7291L31

move it here:
https://github.com/libvmi/libvmi/pull/794/files#diff-68b5d5b58f7a676bca6caf17aeb45a34R162

set the Xenstore variables to `FORCE`, since we are in a child cmakelists.txt

and update the config.h after processing the `libvmi` directory
https://github.com/libvmi/libvmi/pull/794/files#diff-af3b638bc2a3e6c650974192a53c7291R112

Or we could merge the feature summary directly, and fix the Xenstore conditional compilation afterwards, now that the travis tests works

tklengyel:
This PR has got it working somewhat. I pretty much hate cmake at this point so I'm very much considering reverting back to autotools in my private repo and not bothering with cmake anymore. The only upside of cmake i was considering was that perhaps it will work under Windows as well but that's not the case. So beside regressions and bugs we didn't gain anything at this point by switching.

mtarral:
> I pretty much hate cmake at this point so I'm very much considering reverting back to autotools in my private repo and not bothering with cmake anymore.

I understand your frustration.
We had regressions because I was the only one working on the CMake and it worked, but I missed corner cases in the review, so you can blame it on me if you want.
I have done my best to fix them when I had the time.

But if judge by some of our users review, they like the benefits of CMake for integration:
https://github.com/tklengyel/drakvuf/pull/625#issuecomment-490114109

> The only upside of cmake i was considering was that perhaps it will work under Windows as well but that's not the case

Gerhart tried to compile Libvmi on Windows, but he was stopped by the dependencies, not CMake itself.
If we could find a way to build glib and json-c on Windows, we can have a Libvmi Windows build.
But this also stopped my own investigation of this idea.

Despite the issues we have had so far, I still think it's a win for users and developers coming to the project, because it's more likely that they will know CMake than autotools if they want to integrate their tools or customize the build.

What I can do is split the feature_summary PR into 2 distinct PRs, one that fixes the build as you wanted (disable driver if the dependencies are not detected), and the one that brings the feature summary.

tklengyel:
At this point I would not revert upstream exactly because of users. But I find tweaking cmake and making corrections to it an annoying timesink I can't deal with. So as you say, you are the only one maintaining it at this point. Adding the fact that the Jenkins CI was effectively broken for months now because of changes you made to the build system without adjusting the CI is problematic.

PR #810, opened 2019-08-16
Copy SS register for events

tklengyel:


smaresca:
@tklengyel ah excellent, thanks.

PR #814, opened 2019-09-06
Fixes for VMI against Windows mem snapshot (file mode)

mleinhos:
Analysis against a Windows 7 Xen VM worked fine, but failed against a memory snapshot of that same VM. I traced this down to a couple of oddities, namely:
- The technique for finding the System process was incorrectly checking for a string match
- Structure offsets were not pulled from Rekall if the Rekall init failed. However, the overall libvmi init could still succeed, but without those offsets.

### DIFF
--- a/libvmi/os/windows/core.c
+++ b/libvmi/os/windows/core.c
@@ -917,6 +917,7 @@ init_from_rekall_profile_real(vmi_instance_t vmi, reg_t kpcr_register_to_use)
         dbprint(VMI_DEBUG_MISC, "**KernBase PA=0x%"PRIx64"\n", windows->ntoskrnl);
 
         if (windows->ntoskrnl && kdvb && kernbase_offset) {
+            /* This could fail because the address mode is unknown. */
             if ( VMI_FAILURE == vmi_read_addr_pa(vmi, windows->ntoskrnl + kdvb + kernbase_offset, &windows->ntoskrnl_va) )
                 goto done;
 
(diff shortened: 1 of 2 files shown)


tklengyel:

libvmi/os/windows/core.c
```diff
@@ -1046,10 +1047,20 @@ init_core(vmi_instance_t vmi)
...
+    /* Fall be here too if the Rekall profile based init
+     * fails. N.B. the Rekall profile init could fail against a file
+     * because the address mode is unknown, so another attempt is made. */
+    ret = init_from_kdbg(vmi);
     if ( VMI_FAILURE == ret )
-        ret = init_from_kdbg(vmi);
+        goto done;
 
+    dbprint(VMI_DEBUG_MISC, "** Retrying init from Rekall profile.\n");
+    (void) init_from_rekall_profile_real(vmi, (reg_t)0);
```
mleinhos: I'll change this to capture the return status.

libvmi/os/windows/core.c
```diff
@@ -917,6 +917,7 @@ init_from_rekall_profile_real(vmi_instance_t vmi, reg_t kpcr_register_to_use)
         dbprint(VMI_DEBUG_MISC, "**KernBase PA=0x%"PRIx64"\n", windows->ntoskrnl);
 
         if (windows->ntoskrnl && kdvb && kernbase_offset) {
+            /* This could fail because the address mode is unknown. */
             if ( VMI_FAILURE == vmi_read_addr_pa(vmi, windows->ntoskrnl + kdvb + kernbase_offset, &windows->ntoskrnl_va) )
                 goto done;
```
mleinhos: Although I didn't change this block, the logic seems wrong. If a read fails, we bail. Right after that, we try the read again.

In the case of a file, the first read will fail because we don't have the addr mode.

tklengyel:
@mleinhos What's the status of this PR? Still working on it?

mleinhos:
Thanks for the reminder @tklengyel . I'm pushing a small change around line 1060. I have not addressed my concern around line 920.

tklengyel:
@libvmi-jenkins Retest this please

tklengyel:
@libvmi-jenkins Retest this please

tklengyel:
@libvmi-jenkins Retest this please

tklengyel:
This PR seems to break LibVMI for Windows XP guests.

tklengyel:
@libvmi-jenkins Retest this please

mleinhos:
Hmm. Can you pinpoint where? Unfortunately I don't have any XP VMs on hand.

tklengyel:
It also causes Jenkins itself to crash. Here is the last XP test that failed: http://libvmitest.zentific.com:8080/job/PR_LibVMI_Xen_Test_Winxpsp2/lastFailedBuild/console

mleinhos:
I've attempted to repro with an XP memory snapshot, but neither this
branch, nor `master`can find the Windows structures in the image to
bootstrap the init function.

I'll look into this more as time permits.

tklengyel:

PR #816, opened 2019-10-01
xen: check and force ALPT2M_external before setting altp2m dom state

mtarral:
This PR implement the idea we previously discussed in https://github.com/tklengyel/drakvuf/pull/681#issuecomment-519947749 (enforcing ALPT2M_external in `vmi_slat_set_domain_state`


tklengyel:
No, this can't be merged into LibVMI. There are legitimate usecases where the altp2m setting is not external-only. Enforcing that setting for DRAKVUF was fine because there it is a requirement. Here you just made it impossible to use #VE with LibVMI.

mtarral:
The goal was to remove a direct call to a xen library in Drakvuf (working on #679 )

if I understand correctly, you also want Drakvuf to run on multiple hypervisors, and your priority is `boxy`.

How do you suggest we should proceed to remove these calls (`xc_hvm_param_get/set`) ?

PR #817, opened 2019-10-01
Add vmi_toggle_single_step_vcpu

### LINKED ISSUE #815: Implementing vmi_toggle_single_step_vcpu
Hi,

I'm following up on an old issue I had regarding the API, where I wanted to control singlestepping without `TOGGLE_SINGLESTEP` callback response.

In `libvmi/events.h` an API already exists: `vmi_stop_single_step_vcpu`, but
the `start` API is missing.

As we already discussed this [before](https://172.255.51.119/xen/devel/550756), you recommended me to implement it, and to merge both of them in a single API:

Which I did:
~~~C
status_t
vmi_toggle_single_step_vcpu(
    vmi_instance_t vmi,
    vmi_event_t* event,
    uint32_t vcpu,
    bool enabled)
{
#ifdef ENABLE_SAFETY_CHECKS
    if (!vmi || !event)
        return VMI_FAILURE;

    if (!(vmi->init_flags & VMI_INIT_EVENTS))
        return VMI_FAILURE;
#endif

    if (enabled) {
        SET_VCPU_SINGLESTEP(event->ss_event, vcpu);

        gint *key = g_malloc0(sizeof(gint));
        if (!key)
            return VMI_FAILURE;

        *key = vcpu;

        if (!g_hash_table_insert(vmi->ss_events, key, event)) {
            g_free(key);
            return VMI_FAILURE;
        }

        return driver_start_single_step(vmi, &event->ss_event);
    } else {
        UNSET_VCPU_SINGLESTEP(event->ss_event, vcpu);

        gint key = vcpu;

        if (!g_hash_table_remove(vmi->ss_events, &key))
            return VMI_FAILURE;

        return driver_stop_single_step(vmi, vcpu);
    }
}
~~~

However I've noticed that it doesn't work as expected.
For example with this code:

~~~C
    memset(&single_event, 0, sizeof(vmi_event_t));
    single_event.version = VMI_EVENTS_VERSION;
    single_event.type = VMI_EVENT_SINGLESTEP;
    single_event.callback = single_step_callback;
    single_event.ss_event.enable = 1;
    SET_VCPU_SINGLESTEP(single_event.ss_event, 0);
    vmi_register_event(vmi, &single_event);
    while (!interrupted ) {
        printf("Waiting for events...\n");
        vmi_events_listen(vmi,500);
    }
    printf("stopping singlestep\n");
    if (VMI_FAILURE == vmi_toggle_single_step_vcpu(vmi, &single_event, 0, false)) {
        fprintf(stderr, "fail to disable sstep\n");
    }
    printf("stopped. VM should be running\n");
    sleep(10);
    printf("starting sstep\n");
    if (VMI_FAILURE == vmi_toggle_single_step_vcpu(vmi, &single_event, 0, true)) {
        fprintf(stderr, "fail to enable sstep\n");
    }
    sleep(1);
    printf("Finished with test.\n");
~~~

When it reaches `stopped. VM should be running`, the MTF flag has been disabled, but the VM is not executing because a singlestep event is waiting for processing on the ring.

Therefore it doesn't work as intended.

@tklengyel: how should we implement this function so that the previous example works as expected ?
Should we clean the ring when we disable MTF ?

Should solve #665

mtarral:
This PR replaces `vmi_stop_single_step_vcpu` by a new API: `vmi_toggle_single_step_vcpu`, adding a boolean parameter to toggle the singlestep on-demand.

Notes:
- the singlestep example has been updated to demonstrate how this API should be used, as well as make use of `atexit` to better handle errors and ensure `vmi_destroy()` has been called while keeping the example code minimal.
- the API description better reflects in which context it should be called
- `g_try_malloc0` is used instead of `g_malloc0`, since [it doesn't abort on failure](https://developer.gnome.org/glib/unstable/glib-Memory-Allocation.html#g-try-malloc0) and provide error recovery: 

It should fix #815

### DIFF
--- a/examples/singlestep-event-example.c
+++ b/examples/singlestep-event-example.c
@@ -27,16 +27,23 @@
 #include <stdio.h>
 #include <inttypes.h>
 #include <signal.h>
+#include <unistd.h>
 
 #include <libvmi/libvmi.h>
 #include <libvmi/events.h>
 
-vmi_event_t single_event;
+static vmi_event_t single_event = {0};
+static vmi_instance_t vmi = {0};
 
 static int interrupted = 0;
 static void close_handler(int sig)
 {
     interrupted = sig;
+}
+
+void exit_cleanup()
+{
+    vmi_destroy(vmi);
 }
 
 event_response_t single_step_callback(vmi_instance_t vmi, vmi_event_t *event)
@@ -52,14 +59,12 @@ event_response_t single_step_callback(vmi_instance_t vmi, vmi_event_t *event)
 
 int main (int argc, char **argv)
 {
-    vmi_instance_t vmi;
-
     struct sigaction act;
 
     char *name = NULL;
 
     if (argc < 2) {
-        fprintf(stderr, "Usage: single_step_example <name of VM> \n");
+        fprintf(stderr, "Usage: %s <name of VM> \n", argv[0]);
         exit(1);
     }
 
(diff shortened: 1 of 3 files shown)


mtarral:
@tklengyel any thoughts on this one ?

tklengyel:
Removing an API entirely is not a good route. You need to first declare it deprecated but keep it for compatibility. Eventually down the road if everyone is happy with the new function it can be removed. Till then, it needs to remain.

examples/singlestep-event-example.c
```diff
@@ -82,23 +87,91 @@ int main (int argc, char **argv)
...
+    }
+    // process a few more singlestep events
+    for (int i=0; i < 5; i++) {
+        printf("Waiting for events...\n");
+        if (VMI_FAILURE == vmi_events_listen(vmi,500)) {
+            fprintf(stderr, "Failed to listen on events\n");
+            return 1;
+        }
+    }
+    printf("Finished with test.\n");
```
tklengyel: Perhaps add a comment that all singlestep events are disabled by `vmi_destroy`.

libvmi/events.c
```diff
@@ -1060,10 +1061,31 @@ vmi_stop_single_step_vcpu(
         return VMI_FAILURE;
```
tklengyel: You should add a safety check that checks if this was called from an event callback. For that use-case there is an event_response_t flag that should be used.

examples/singlestep-event-example.c
```diff
@@ -82,23 +87,91 @@ int main (int argc, char **argv)
...
+
+    // toggle singlestep back ON
+    // pause before changing VM state
+    if (VMI_FAILURE == vmi_pause_vm(vmi)) {
+        fprintf(stderr, "Failed to pause VM\n");
+        return 1;
+    }
+    printf("Restarting singlestep\n");
+    for (unsigned int vcpu=0; vcpu < num_vcpus; vcpu++) {
+        if (VMI_FAILURE == vmi_toggle_single_step_vcpu(vmi, &single_event, vcpu, true)) {
```
tklengyel: Perhaps here use `vmi_shutdown_single_step` to demonstrate that it's equivalent to what you do above with `vmi_shutdown_single_step`

mtarral: This loop is restarting the singlestep, `vmi_shutdown_single_step` will shut it down.
I'm not sure what you meant here ?

tklengyel: Ah yea, sorry, I thought this was just turning it off again.

libvmi/events.c
mtarral: Done

examples/singlestep-event-example.c
mtarral: Done

tklengyel:

tklengyel:

tklengyel:

PR #825, opened 2019-10-14
mmap_guest: check calloc return value

mtarral:
This commit fixes a check of `calloc` in `vmi_mmap_guest`


tklengyel:

PR #826, opened 2019-10-14
xen: properly close xen dynamic library handles on error

mtarral:
This PR closes the `libxenctrl/libxenstore` handles upon error.


tklengyel:

PR #830, opened 2019-10-24
fix memory leaks and improper free() in kvm driver

redspot:
Leaks were detected in the KVM driver using Google's AddressSanitizer:
https://github.com/google/sanitizers/wiki/AddressSanitizer

Flags used to build our software (not libvmi): -g3 -fsanitize=address -fno-omit-frame-pointer

Conditions:
I am using LibVMI for malware unpacking research, with the host system based on Xen.

After program startup failed, due to stale/stuck VMs, ASAN reported the leaks in kvm_init(), which I wasn't even using. After a quick code inspection, I noticed the failure to g_free(kvm) where appropriate, and one use of free(kvm), which I believe fails on something allocated by g_malloc*() family.

Project Reference:
https://github.com/carter-yagemann/vmi-unpack


tklengyel:

PR #832, opened 2019-10-24
Clean-up Xen events shutdown a bit

tklengyel:
Only reset EPT permissions if there are still active events registered. Otherwise it's just a waste of time (and it's quite time-consuming for large guests).


tklengyel:

Issue #778, opened 2019-06-03
How to make event-example in example support  KVM ?


yourlittle:
event-example in example does not support KVM, Becaues KVM does not expose APIs equivalent to the Xen event APIs.It's can use qemu to  achieve it?

mtarral:
Hi,
please have a look at https://github.com/KVM-VMI/kvm-vmi/tree/kvmi (branch `kvmi`), where we are implementing a new VMI API which brings KVM hardware events into LibVMI

Thanks.

mtarral:
@yourlittle a PR has been opened offering basic event support on KVM:

https://github.com/KVM-VMI/libvmi/pull/9

@tklengyel maybe we can close this issue ?

smaresca:
I think we should keep it open until the support is formally merged into LibVMI.

tklengyel:
> I think we should keep it open until the support is formally merged into LibVMI.

+1

Wenzel:
Hi @yourlittle ,

I can't help you if you don't give me more details.
how did you setup kvm-vmi, which branch are you using ? (`kvmi`) I suppose ?

you are supposed to give the kvmi socket path as the second argument.
What you have here is `10002`, what is it supposed to mean ? is it the socket path ?

Open a new issue on kvm-vmi/libvmi so we can go into details there.

yourlittle:
Oh! What you mean is that I need to use the KVM provided in the KVM-VMI project, not just install libvmi.

yourlittle:
So,In order to achieve this function, you modified the KVM kernel?

Wenzel:
The setup is explained on the Wiki:
https://github.com/KVM-VMI/kvm-vmi/wiki/KVM-VMI-setup

yourlittle:
when I execute"$ make -j4 bzImage" commands, there always report errors and stop compiling
![TIM截图20190917162124](https://user-images.githubusercontent.com/50984180/65024369-4f7d9580-d967-11e9-9ae5-c8c8da1721a9.png)

yourlittle:
Configure, build and install QEMU
![TIM截图20190918171116](https://user-images.githubusercontent.com/50984180/65134758-6ab7d600-da37-11e9-86b5-f4be03a93dd3.png)

mtarral:
@yourlittle 
[...]

PR #844, opened 2019-12-05
Propose new KVM driver based on KVMi subsystem

### LINKED ISSUE #827: Integration of the new KVM driver based on KVMi APIs
Hi,

As the new LibVMI KVM driver is progressing fast and is already [offering basic event handling](https://github.com/KVM-VMI/libvmi/pull/9), I wanted to have the LibVMI maintainers opinion on how and when we could merge our [`kvmi`](https://github.com/KVM-VMI/libvmi/tree/kvmi) branch upstream.

What are your requirements ?
- polishing the driver to support `event_response` ?
- having a test suite ?
- having a new KVMi release based on an LTS kernel ? (latest v6 is on 5.0.0-rc7)

Thanks.

mtarral:
This PR proposes a brand new KVM driver, based on the experimental KVMi subsystem currently being developed by BitDefender.

The driver's code has been splitted:
- `kvm_legacy.c`: legacy kvm driver (GDB and qemu memaccess patch)
- `kvm.c` the new driver based in KVMi
- `kvm_common.c`: common functions share by both drivers.

This PR also changes some of the examples to include a new `kvmi_socket` optional parameter.

cc @tklengyel , @adlazar, @mdontu

Fixes #827 , #778


tklengyel:

mtarral:
Astyle test fails because of the includes copied from KVM:
https://travis-ci.org/libvmi/libvmi/jobs/621162984#L248

Can we exclude them ?

The rest of the tests fails because the new driver isn't integrated with autotools:
https://travis-ci.org/libvmi/libvmi/jobs/621162986#L497

We need to include the `libvmi/driver/kvm/include` header location, and link with `libkvmi`
CMake equivalent:
https://github.com/libvmi/libvmi/pull/844/files#diff-da0b452291e9058696a8da3487c538d7R9

mtarral:
@tklengyel can you take care of autotools integration and push your commits on this PR so the Travis passes ?
I don't know autootols :/

tklengyel:
@mtarral no, sorry, this is your PR, please make sure it can pass the CI gates properly

tklengyel:
> Astyle test fails because of the includes copied from KVM:
> https://travis-ci.org/libvmi/libvmi/jobs/621162984#L248
>
> Can we exclude them ?

I don't see why we would want to do that.

> We need to include the libvmi/driver/kvm/include header location, and link with libkvmi

Either update the Travis CI system to set that up for you or explicitly disable any type of CI testing for the KVM driver. If you do that though, I would also request that you make a note in the README that says the KVM drivers are not tested at all. This will likely lead to a high number of issues down the road of people complaining about it being broken, so I would advise against that.

libvmi/driver/kvm/kvm_private.h
```diff
@@ -33,15 +33,28 @@
 
 #include "private.h"
 #include "libvirt_wrapper.h"
+#ifndef KVM_LEGACY_DRIVER
+# include "driver/kvm/include/kvmi/libkvmi.h"
```
tklengyel: You probably should just do `#include "include/kvmi/libkvmi.h"` instead of specifying the full path from the top.

mtarral:
@tklengyel I fixed the CI, it's green now :tada: 

I added support in autotools to add `libkvmi`'s headers and compile with it,
as well as adding a new option `--enable-kvm-legacy` (disabled by default)

libvmi/driver/kvm/include/asm/kvm.h
```diff
@@ -0,0 +1,364 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
```
tklengyel: Could you explain this comment? Is this header under GPL2? I don't think we can mix GPL and LGPL

libvmi/driver/kvm/include/asm/kvmi.h
```diff
@@ -0,0 +1,124 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
```
tklengyel: Same here. Usually ABI headers are exempt from license restrictions to avoid issues (at least thats the case with Xen). Isn't that the case here too?

libvmi/driver/kvm/include/asm/kvm.h
mtarral: cc @adlazar, @mdontu

smaresca: In kernel headers, that line specifically means that the header is placed outside of the Linux GPL license requirements, allowing it to be included into non GPL compliant user-space application code.

Anything built with LibVMI would constitute a user-space application falling safely within that exception; HOWEVER, the placement of that header with that marking within LibVMI is inherently problematic. Such headers are meant to be included by reference, not embedded entirely.

libvmi/driver/kvm/include/asm/kvmi.h
smaresca: my earlier comment applies here also.

libvmi/driver/kvm/include/asm/kvm.h
smaresca: To elaborate:

If both headers are intended to be included in the installation of a kernel with kvmi, then relicensing is unnecessary as far as I can tell (and embedding of the header within LibVMI should be unnecessary).

The kvmi driver would literally just need to #include the kvmi.h etc headers (which should still retain the "PDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note"), and we're all set AND LibVMI safely remains LGPL.

In addition to the doing the above, if LibVMI-specific material has been added into the headers in question, it should be extracted into a much smaller header, LGPL licensed, and included in the LibVMI kvmi subtree.

mdontu: I agree. I think we can 'unbundle' kvm.h. It's part of the kernel UAPI and hence it should not see changes that break backwards compatibility.

I think we used qemu as example which also bundles them (since 2011). I don't know why the chose to do that.

mtarral:
What about removing `libkvmi` entirely from the kvm driver sources ?

We would expect the KVM user to install this library system-wide (or tweak the build parameters)
In the end, we don't really want to embbed this library in Libvmi, as it should evolve on its own.
@mdontu, @tklengyel ?

mdontu:
> What about removing `libkvmi` entirely from the kvm driver sources ?
> 
> We would expect the KVM user to install this library system-wide (or tweak the build parameters)
> In the end, we don't really want to embbed this library in Libvmi, as it should evolve on its own.
> @mdontu, @tklengyel ?

I'm OK with that.

tklengyel:
If you mean libkvmi would be like libxc is on Xen, then yes, that's a viable route to take.

mtarral:
> If you mean libkvmi would be like libxc is on Xen, then yes, that's a viable route to take.

Yes, that's what I meant.
I will push the fix when I have the time

tklengyel:
@mtarral While libkvmi is not merged upstream to KVM perhaps it would be appropriate to add a submodule to LibVMI to make it easy to find.

mtarral:
@tklengyel since the KVMi subsystem has not been merged upstream too, LibVMI users who wants to use the new KVM driver will have to go through the KVM-VMI Setup page:
https://github.com/KVM-VMI/kvm-vmi/wiki/KVM-VMI-setup

Which contains the setup instructions the whole stack:
- KVM
- QEMU
- Libkvmi

What do you think ? Should we still add a submodule ?

tklengyel:
I guess just adding a section to the `README` that says for KVMi to follow instruction at that site would be enough. But for the Travis we should add a test-case that goes through the motion of installing all the dependencies for this new driver so that at least it's compile tested.

README.rst
```diff
@@ -113,50 +113,32 @@ can then be transferred to the XenServer dom0 and run natively.
...
-  2) Enable GDB access to your KVM VM.  This is done by adding
-     '-s' to the VM creation line or, by modifying the VM XML
-     definition used by libvirt as follows:
+https://github.com/KVM-VMI/kvm-vmi
 
-     - Change:
+Follow this wiki page to setup KVM-VMI:
 
-       .. code::
+https://github.com/KVM-VMI/kvm-vmi/wiki/KVM-VMI-setup
```
Wenzel: @tklengyel I updated the `README` already in my PR, you can review the instructions.

Regarding the Travis, I will get it fixed today.

mtarral:
@tklengyel the Travis is green except `astyle`:
> /home/travis/.travis/functions: line 109: /usr/bin/astyle: Argument list too long

Wenzel:
@tklengyel and we have a :green_heart: :construction_worker_man: !

tklengyel:
Jenkins is failing because either though the build was performed with `./configure --disable-kvm` the linker was still trying to link with kvmi: `/usr/bin/ld: cannot find -lkvmi`

tklengyel:
You will also need to introduce a kvmi wrapper similar to what we have for libvirt (see https://github.com/libvmi/libvmi/blob/master/libvmi/driver/kvm/libvirt_wrapper.c#L72). This is necessary to be able to compile a single version of LibVMI that can then be deployed across different hypervisors. The wrapper effectively looks up the required libraries at runtime based on what hypervisor is detected, instead of doing that at compile time. For example if we ever want to get LibVMI into debian/centos/etc. then it wouldn't make sense to have separate packages (libvmi-xen/libvmi-kvm/etc) for the different settings. And of course it would also not make sense to require kvmi libraries to be installed on a Xen host, or vice-verse.

Makefile.am
```diff
@@ -135,6 +146,10 @@ libvmi_libvmi_la_CFLAGS= -fvisibility=hidden $(GLIB_CFLAGS) $(JSONC_CFLAGS)
 libvmi_libvmi_la_LDFLAGS= $(GLIB_LIBS) $(JSONC_LIBS) -no-undefined
 libvmi_libvmi_la_LDFLAGS+= -version-info $(VERSION)
 
+if !WITH_KVM_LEGACY
```
tklengyel: This whole block is uneccesary, the `AC_CHECK_LIB(kvmi, kvmi_init_vsock` in the configure script will automatically add the required libraries to the linker invocation, you don't have to manually specify. It's also why Jenkins failed.

tklengyel: With the wrapper added for kvmi the `AC_CHECK_LIB` call would also need to go btw since no hard linking would be done to kvmi anymore.

mtarral:
@tklengyel I included `libkvmi_wrapper.h/c` to dynamically load `libkvmi`.

I removed the linking in `Makefile.am`, `configure.ac` and `libvmi/CMakeLists.txt`

Also I modified some of the examples to accept an optional `socket` parameter, so they can be compatible with the new KVM driver, and I took the freedom to add a `examples/README.md` file, explaining what the examples are doing, and which of them are not compatible yet.

examples/msr-event-example.c
```diff
@@ -48,8 +46,12 @@ static void close_handler(int sig)
 
 int main (int argc, char **argv)
 {
-    vmi_instance_t vmi;
-    struct sigaction act;
+    vmi_instance_t vmi = {0};
+    vmi_event_t msr_event = {0};
+    struct sigaction act = {0};
+    vmi_init_data_t *init_data = alloca(sizeof(vmi_init_data_t)
```
tklengyel: alloca is generally discouraged since it breaks stack protections

examples/process-list.c
```diff
@@ -84,14 +87,21 @@ int main (int argc, char **argv)
                     config_type = VMI_CONFIG_JSON_PATH;
                     config = (void*)optarg;
                     break;
+                case 's':
+                    init_data = alloca(sizeof(vmi_init_data_t)
```
tklengyel: alloca is generally discouraged since it breaks stack protections

libvmi/driver/kvm/kvm_events.c
```diff
@@ -0,0 +1,835 @@
...
+/*
+ * Helpers
+ */
+static void
+fill_ev_common_kvmi_to_libvmi(
+    struct kvmi_dom_event *kvmi_event,
+    vmi_event_t *libvmi_event)
+{
+    //      standard regs
+    libvmi_event->x86_regs->rax = kvmi_event->event.common.arch.regs.rax;
```
tklengyel: Would be faster to have a local pointer to `x86_regs` declared so you don't do double dereferences.

libvmi/driver/kvm/kvm_events.c
```diff
@@ -0,0 +1,835 @@
...
+            }
+        }
+    }
+    // the rpl struct should be like this
+    //    struct {
+    //        struct kvmi_vcpu_hdr hdr;
+    //        struct kvmi_event_reply common;
+    //        // Event specific structs
+    //        ...
+    //    };
```
tklengyel: This type of structure being passed around is really odd. Instead of it being casted and passed around, just allocate each header individually and pass the required pointers to the functions that needs them.

libvmi/driver/kvm/kvm_events.c
```diff
@@ -0,0 +1,835 @@
...
+
+    // call user callback
+    event_response_t response = call_event_callback(vmi, libvmi_event);
+
+    // reply struct
+    struct {
+        struct kvmi_vcpu_hdr hdr;
+        struct kvmi_event_reply common;
+        struct kvmi_event_msr_reply msr;
+    } rpl = {0};
```
tklengyel: See my previsous comment, this type of construct should be avoided.

libvmi/driver/kvm/kvm_private.h
```diff
@@ -30,18 +30,34 @@
...
+#else
+    void *kvmi;
+    void *kvmi_dom;
+    libkvmi_wrapper_t libkvmi;
+    pthread_mutex_t kvm_connect_mutex;
+    pthread_cond_t kvm_start_cond;
+    unsigned int expected_pause_count;
+    // dispatcher to handle VM events in each process_xxx functions
+    status_t (*process_event[KVMI_NUM_EVENTS])(vmi_instance_t vmi, struct kvmi_dom_event *event);
+#endif
```
tklengyel: So instead of this ifdeffery, it would be better to split the legacy driver and the new driver completely. Why do you want to rule out the possibility of building a single libvmi instance that can work with both?

libvmi/driver/kvm/libkvmi_wrapper.c
```diff
@@ -0,0 +1,63 @@
+
```
tklengyel: License info missing

libvmi/driver/kvm/libkvmi_wrapper.h
```diff
@@ -0,0 +1,82 @@
+#ifndef LIBKVMI_WRAPPER_H
```
tklengyel: License info missing

libvmi/msr-index.c
```diff
@@ -1,26 +1,4 @@
-/* The LibVMI Library is an introspection library that simplifies access to
```
tklengyel: You are removing the license here

examples/cr3-event-example.c
```diff
@@ -0,0 +1,131 @@
...
+    printf("CR3 write happened: Value=0x%"PRIx64"\n", event->reg_event.value);
+    return VMI_EVENT_RESPONSE_NONE;
+}
+
+int main (int argc, char **argv)
+{
+    vmi_instance_t vmi;
+    status_t status;
+    vmi_mode_t mode;
+    vmi_init_data_t *init_data = alloca(sizeof(vmi_init_data_t)
```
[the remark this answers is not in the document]
mtarral: I used `alloca` because it was easy to manage for an example code, where we are focused on showing API usage.
And the compiler didn't trigger any warnings, even with `cmake -DHARDENING=ON`.

Should I replace this with a `malloc/free` ?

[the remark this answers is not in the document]
tklengyel: Yes please, the examples should exhibit good coding practices.

libvmi/driver/kvm/kvm_events.c
mtarral: The issue I was attempting to solve is that the reply might contain event specific fields.
And we need to set these event specified in the event handler.

Example with `process_register()`

~~~C
    // reply struct
     struct {
         struct kvmi_vcpu_hdr hdr;
         struct kvmi_event_reply common;
         struct kvmi_event_cr_reply cr;
     } rpl = {0};


     // TODO: how can the callback specifiy a new value for the MSR ?
     // libvmi_event.reg_event.xxx
     rpl.cr.new_val = kvmi_event->event.cr.new_value;

     // and then calling the generic "process_callback_response" function, passing rpl pointer.
~~~

I was following the example provided in `kvmi`:
https://github.com/KVM-VMI/kvm/blob/kvmi/tools/kvm/kvmi/kvmi-test.c#L83
cc @mdontu, @adlazar 

> nstead of it being casted and passed around, just allocate each header individually and pass the required pointers to the functions that needs them.

So, if I understand correctly,  I would do something like this
~~~C
// allocate header
struct kvmi_vcpu_hdr hdr;
// allocate common
 struct kvmi_event_reply common;
// allocate event specific struct if need
struct kvmi_event_cr_reply cr;

int size = sizeof(hdr) + sizeof(common) + sizeof(cr);

// pass them to reply function
process_callback_response(..., &hdr, &common, (void*)&cr, size );
~~~

:thinking:

libvmi/driver/kvm/kvm_private.h
mtarral: So adding a `KVM_LEGACY` driver is doable, I just didn't think we would keep both at the same time.

tklengyel: I don't know, I think it might make sense to just have a single build of LibVMI that can target a bunch of different setups. I mean, that's kinda the point of LibVMI :)

libvmi/driver/kvm/kvm_events.c
tklengyel: Well, it looks to me `kvmi_vcpu_hdr` and `kvmi_event_reply` are standard headers all events would use, so I would update the prototype of process_callback_response to take the pointers to those structs directly. If the `kvmi_event_cr_reply` is variable based on the event type, it would be fine to pass that as a void* and then cast it when needed to its respective type in the part of process_callback_response that needs it. Does that make sense?

mtarral: I think we can even remove this part
~~~C
   struct kvmi_vcpu_hdr *hdr = (struct kvmi_vcpu_hdr *)rpl;
     hdr->vcpu = kvmi_event->event.common.vcpu;

     struct kvmi_event_reply *common = (struct kvmi_event_reply *)( hdr + 1 );
     common->event = kvmi_event->event.common.event;
     common->action = KVMI_EVENT_ACTION_CONTINUE;
~~~

and make the `process_xxxx` functions do the reply.

Example:
~~~C
process_register(vmi_instance_t vmi, struct kvmi_dom_event *kvmi_event)
{
  // handle register event ...
  // call callback
  event_response_t response = call_event_callback(vmi, libvmi_event);
  // prepare reply struct
    struct {
         struct kvmi_vcpu_hdr hdr;
         struct kvmi_event_reply common;
         struct kvmi_event_cr_reply cr;
     } rpl = {0};

   int action = process_cb_response()

   rpl.hdr.vcpu = kvmi_event->event.common.vcpu;
   rpl.common.event = kvmi_event->event.common.event;
   rpl.action = action;

   // do the reply
   if (kvm->libkvmi.kvmi_reply_event(kvm->kvmi_dom, kvmi_event->seq, rpl, sizeof(rpl)))
         return VMI_FAILURE;

   return VMI_SUCCESS;
}

static int process_cb_response()
{
    // loop over all possible responses
     for (uint32_t i = VMI_EVENT_RESPONSE_NONE+1; i <=__VMI_EVENT_RESPONSE_MAX; i++) {
         event_response_t candidate = 1u << i;
         if (response & candidate) {
             switch (candidate) {
                 default:
                     errprint("%s: KVM - unhandled event reponse %u\n", __func__, candidate);
                     break;
             }
         }
     }

     return KVMI_EVENT_ACTION_CONTINUE; // or KVMI_EVENT_ACTION_DENY, ...
}
~~~

What do you think

libvmi/driver/kvm/kvm_private.h
mtarral: > I mean, that's kinda the point of LibVMI :)

Well yes, after all.
But I would rather keep this refactoring for another PR, as this one is really huge already.

tklengyel: Fair enough.

libvmi/driver/kvm/kvm.c
```diff
@@ -547,40 +165,190 @@ status_t
...
+            *value = sregs.gs.base;
+            break;
+        case MSR_EFER:
+            *value = msrs.entries[0].data;
+            break;
+        case MSR_STAR:
+            *value = msrs.entries[1].data;
+            break;
+        default:
+            dbprint(VMI_DEBUG_KVM, "--Reading register %"PRIu64" not implemented\n", reg);
```
namidairo: Looks like you have MSR_LSTAR implemented in kvm_get_vcpuregs but not here

mtarral: Good catch.
I also added `MSR_IA32_SYSENTER_CS/ESP/EIP`.

libvmi/driver/kvm/kvm_events.c
Wenzel: So progressing on the implementation, I would like to add support for
event responses like
- `VMI_EVENT_RESPONSE_SET_EMUL_READ_DATA`
- `VMI_EVENT_RESPONSE_SET_EMUL_INSN`

However, there is a catch.
Only the pagefault events can offer emulation in replies:
~~~C
struct kvmi_event_pf_reply {
	__u64 ctx_addr;
	__u32 ctx_size;
	__u8 singlestep;
	__u8 rep_complete;
	__u16 padding;
	__u8 ctx_data[256];
};
~~~

So there are 2 solutions to implement this

# Solution 1

Handle `VMI_EVENT_RESPONSE_SET_EMUL_READ_DATA` and `VMI_EVENT_RESPONSE_SET_EMUL_INSN` in the same common function: `process_cb_response`.

However, since we are passing a `void *rpl`, we have to cast and do pointer arithmetic, similar to
~~~C
    struct kvmi_vcpu_hdr *hdr = (struct kvmi_vcpu_hdr *)rpl;
    //  

    struct kvmi_event_reply *common = (struct kvmi_event_reply *)( hdr + 1 );
    // 


   case VMI_EVENT_RESPONSE_SET_EMUL_READ_DATA:
          struct kvmi_event_pf_reply *pf_reply = (struct kvmi_event_pf_reply *)( common + 1 );
          // handle emulation response
          memcpy(pf_reply.ctx_data, ....)
~~~

What I don't like with this solution is that we are making ugly casts, and aren't doing any sanity check.
If a `interrupt` callback sets this event response, the `* rpl` struct will be overwritten here.

Here comes solution 2

# Solution 2

What I would prefer is declaring a second function, with the only goal to handle this emulation event response, and which will be called from the `process_pagefault` handler
~~~C
            // call user callback
            event_response_t response = call_event_callback(vmi, libvmi_event);

            // handle emulation reply requests
            if (VMI_FAILURE == process_cb_response_emulate(vmi, response, libvmi_event, &rpl))
                return VMI_FAILURE;
~~~


@tklengyel which one do you prefer ?
Thanks.

.travis.yml
```diff
@@ -2,6 +2,15 @@ os: linux
 dist: bionic
 language: c
 
+# install libkvmi for all tests
+before_install:
```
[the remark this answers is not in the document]
Wenzel: the last commits reintroduce cloning kvm, for the sake of passing the CI.
Once you have you repo ready for libkvmi, I will switch to it.

examples/cr3-event-example.c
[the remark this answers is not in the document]
Wenzel: fixed.

libvmi/msr-index.c
Wenzel: fixed.

libvmi/driver/kvm/libkvmi_wrapper.h
Wenzel: fixed.

libvmi/driver/kvm/libkvmi_wrapper.c
Wenzel: fixed.

Wenzel:
The last commits fixed some of the changes you requested:
- alloca
- local pointer to `x86_regs` to avoid double dereferences
- license info missing or removed

It also introduces a few features:
- descriptor events support, with `descriptor-event-example`
- `fool-patchguard` example (corrupting an SSDT entry, and catching when PatchGuard will check the structure to return fake data) (I wasn't able to actually verify that it works because of a crash I found in KVMi)
- TSC Speed
- a README in `driver/kvm` that details the state of the Libvmi API implementation.

Wenzel:
The `master` has been merged, conflicts are resolved.

Wenzel:
@tklengyel this one is ready for another review.

Wenzel:
I have a question regarding the driver shutdown.

ATM, we are leaving the responsability of restoring the VM state to `libkvmi` functions, (see [`kvm_close_vmi`](https://github.com/KVM-VMI/libvmi/blob/kvmi/libvmi/driver/kvm/kvm.c#L410)) by calling:
~~~C
kvm->libkvmi.kvmi_domain_close(kvm->kvmi_dom, true);
kvm->libkvmi.kvmi_uninit(kvm->kvmi);
~~~

However, the Xen driver shutdown function doesn't assume that Xen shutdown VMI API will do the cleanup, and tracks the state of the events who have been enabled to disable them ([`xen_events_destroy`](https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L3245))

So, @tklengyel  should we implement the same logic in `kvm_events_destroy`, or should we have more expectations from `libkvmi` shutdown/close APIs ?


also cc @mdontu, @adlazar

tklengyel:
IMHO the fewer things the libkvmi APIs "assume" the better. For example, doing the event cleanup I think is better to be left to the user-side (ie LibVMI). The hypervisor API shouldn't decide what events and when to disable on its own. It might be the case the user doesn't want the API to clean things up when it closes it for whatever reasons. Or at least make it optional to do a full cleanup or not.

adlazar:
I will add a command that will control the clean-up process (all or nothing, with nothing being the default behavior), if this is useful.

tklengyel:
I wouldn't ask you guys to add that if you already have the full cleanup implemented as part of the close. But generally speaking a VM on Xen will continue to run fine even if you didn't disable the event and just closed the monitor vm_event interface. Except of course the EPT permissions. But for example with generic mem_events we don't even clear those with LibVMI.

adlazar:
I don't remember exactly how we start to do the clean-up when the introspection channel is closed, but the performance was one thing. We wanted to reduce the number of VM-exits to the not-introspected level. And once we've restored one thing, we did it for all :) This is done on the kernel side because the introspection tool might crash. Of course, not ours :)
If there is a use case to skip the implicit clean-up, we only need a new flag controlled by another command.

tklengyel:
I don't have a particular use-case in mind, but generally my experience has been the fewer things you do on the hypervisor side and let the user decide what they want to is has been always the better as it allows for more flexibility. It is certainly a question whether you want the VM to continue "normally" if the introspection app goes away unexpectedly and thus skipping the cleanup routines. From a security perspective, if the security app crashes, then I would assume its better to be fail-safe and not let the VM continue to run. But even that is probably better to have as something configurable since in some cases that's better, in some other cases it's not. So the more options there are, the more flexible the whole thing is to fit use-cases we don't even know about.

libvmi/driver/kvm/kvm_events.c
```diff
@@ -0,0 +1,987 @@
...
+    static bool pf_enabled = false;
+    unsigned char kvmi_access, kvmi_orig_access;
+    kvm_instance_t *kvm = kvm_get_instance(vmi);
+#ifdef ENABLE_SAFETY_CHECKS
+    if (!kvm || !kvm->kvmi_dom) {
+        errprint("%s: invalid kvm handle\n", __func__);
+        return VMI_FAILURE;
+    }
+#endif
+    // enable PF events the first time we call this function
```
Wenzel, line 890: @mdontu I can't remember, can I move this at the driver initialization function `kvm_events_init` ?
Or does it have a performance cost ?

adlazar: We enable this event during initialization, because we always track memory access. So, we didn't measure it. But, as this setting won't trigger more VM-exists by itself I wouldn't worry that much.

Wenzel:
The last commits add the following changes:
- move PF interception enabling at driver init
- track the state of enabled events and disable them at driver shutdown if needed
- free `kvmi_dom_event` in `vmi_events_listen` loop (was there at the beginning, but apparently it disappeared following a rebase)
- restore page permission if VMI access flag is `VMI_MEMACCESS_N`

.travis.yml
```diff
@@ -20,11 +30,12 @@ jobs:
 #
     - env:
         - TEST="astyle"
+        - ARTISTIC_STYLE_OPTIONS=".astylerc"
       install:
         - sudo apt-get update
         - sudo apt-get -q -y install astyle
       script:
-        - export ARTISTIC_STYLE_OPTIONS=".astylerc" && astyle -Q `find . | egrep "\.c$|\.h$"` > astyle.out
+        - find . -type f \( -name "*.c" -or -name "*.h" \) -exec astyle -Q "{}" >> astyle.out \;
```
tklengyel, line 38: Use instead 
```
astyle -Q `find . -type f -name "*.c" -o -name ".h"

File: examples/cr3-event-example.c [hunk 20]
```diff
@@ -0,0 +1,138 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+ * memory in a target virtual machine or in a file containing a dump of
+ * a system's physical memory.  LibVMI is based on the XenAccess Library.
+ *
+ * Copyright 2011 Sandia Corporation. Under the terms of Contract
```
tklengyel, line 5: No need for this copyright notice for Sandia as you don't work for them

Wenzel: fixed.

.travis.yml
Wenzel: fixed.

Wenzel:
@tklengyel 
I think the CI is blocked.
Travis didn't execute and Jenkins status is busy since yesterday.
Maybe you can restart Travis jobs ?

tklengyel:
> Travis didn't execute

That usually indicates that the travis config is malformed and needs to be fixed. Jenkins is offline and will be offline for a couple months at least.

Wenzel:
Actually now I think it's a bug with Github.
On this PR: https://github.com/KVM-VMI/libvmi/pull/30 you can see that Travis is still busy (even though I merged the PR).
And when you follow the Travis link: https://travis-ci.org/github/KVM-VMI/libvmi/builds/699702545 it's done.
(the whitespace check is always red on kvm-vmi/libvmi, but green when  I open a PR here, I'm not sure why)

Wenzel:
About the Travis issue, I opened a ticket on the Github support last week.
They should reply soon I hope

Wenzel:
@mdontu, @tklengyel I used the new libkvmi repo in the Travis CI instead of cloning the entire KVM repo for each test.
The Travis seems to be back now that I have pushed some commits, that's good because I had zero replies from Github support.

This PR is ready for another review.

Wenzel:
@tklengyel can you trigger the jenkins also ?
thanks

tklengyel:
@drakvuf-jenkins retest this please

.travis.yml
```diff
@@ -147,6 +160,32 @@ jobs:
...
+      install:
+        - sudo apt-get update
+        - sudo apt-get -q -y install bison flex libjson-c-dev libvirt-dev libxen-dev libfuse-dev
+      script:
+        - mkdir build && cd build
+        - cmake -DENABLE_KVM_LEGACY=ON ..
+        - make
+
+#
+# KVM legcy build (autotools)
```
tklengyel, line 177: /r/legcy/legacy

libvmi/driver/kvm/kvm.h
```diff
@@ -106,10 +153,19 @@ driver_kvm_setup(vmi_instance_t vmi)
     driver.check_id_ptr = &kvm_check_id;
     driver.get_name_ptr = &kvm_get_name;
     driver.set_name_ptr = &kvm_set_name;
+    driver.write_ptr = &kvm_write;
     driver.get_memsize_ptr = &kvm_get_memsize;
+# ifndef ENABLE_KVM_LEGACY
```
tklengyel, line 158: Could we group these functions at the end under a single `#ifndef`?

libvmi/driver/kvm/kvm_events.c
```diff
@@ -0,0 +1,1079 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+ * memory in a target virtual machine or in a file containing a dump of
+ * a system's physical memory.  LibVMI is based on the XenAccess Library.
+ *
+ * Copyright 2011 Sandia Corporation. Under the terms of Contract
```
tklengyel, line 5: The Sandia notice doesn't apply, this is new code

tklengyel:
This is looking good to me, I only spotted some minor things that really aren't blockers. Lots of kudos @Wenzel for getting this done! 🥇 🥳

Wenzel:
@tklengyel thanks for you review, I've just fixed the minor issues you have found.
I'm excited to bring this new driver to the LibVMI community !

libvmi/driver/kvm/include/asm/kvm.h
smaresca: @Wenzel Thank you for addressing

libvmi/driver/kvm/include/asm/kvmi.h
smaresca: @Wenzel Thank you for addressing

smaresca:
> @tklengyel thanks for you review, I've just fixed the minor issues you have found.
> Are we waiting for @smaresca's approval also ?
Thank you for addressing my license and header-related concerns. There are a few remaining Sandia references in libvmi/driver/kvm files though that need attention.

Wenzel:
According to @tklengyel's comment: https://github.com/libvmi/libvmi/pull/844#discussion_r461944327
I thought I needed to remove the Sandia license only in new code/new files.
(therefore I should also remove the license in `kvm_common.c` and `kvm_legacy.c`)
is that what you meant @smaresca ?

or did you mean that since the files have changed so much, the Sandia license wasn't applying anymore ?

libvmi/driver/kvm/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+- KVMi: the new [KVM virtual machine introspection API](https://static.sched.com/hosted_files/kvmforum2019/f6/Advanced%20VMI%20on%20KVM%3A%20A%20progress%20Report.pdf)
+- Legacy: the legacy KVM driver for LibVMI, using either `GDB` or the `fast-memaccess` patches available in `libvmi/tools/qemu-kvm-patch`
+
+## LibVMI API Implementation
+
+This section will give an implementation status of the LibVMI API on the new KVM driver.
+
+- [x] r/w physical memory
+- [ ] VCPU registers (x86 only)
```
smaresca, line 15: @Wenzel Does this file need updates? Example: It seems vcpu read/write is documented as unavailable, but I did see get/set implemented.

Wenzel: Go catch, I should have checked the sub checkboxes here

smaresca:
> According to @tklengyel's comment: [#844 (comment)](https://github.com/libvmi/libvmi/pull/844#discussion_r461944327)
> I thought I needed to remove the Sandia license only in new code/new files.
> (therefore I should also remove the license in `kvm_common.c` and `kvm_legacy.c`)
> is that what you meant @smaresca ?
Ah, disregard my comment about the other Sandia lines. The notices I cited do indeed need to stay in place.

libvmi/driver/kvm/README.md
adlazar: libkvmi supports setting general purpose registers only, for now. No special registers and no MSRs.

README.rst
namidairo, line 131: Don't forget to update the references to libvmi in there to point to this repository? Or alternatively point the submodule to reference whatever commit id this ends up being in libvmi/libvmi.

libvmi/driver/kvm/README.md
Wenzel: Now I remember, I didn't check the boxes because of these restrictions.
I suppose I should add a sub-level of boxes on both read and write sections, and check them for general purpose registers, so it's more clear what is supported ATM.

README.rst
Wenzel: This link references a Wiki page of KVM-VMI repo.
But on Github, the Wiki lives inside its own repo, separated from KVM-VMI/kvm-vmi.

Maybe this Setup page should be included in KVM's README, and referenced from the main Libvmi README file ?

libvmi/driver/kvm/README.md
smaresca: @Wenzel  yes please, thank you

README.rst
smaresca: > 
> 
> This link references a Wiki page of KVM-VMI repo.
> But on Github, the Wiki lives inside its own repo, separated from KVM-VMI/kvm-vmi.
> 
> Maybe this Setup page should be included in KVM's README, and referenced from the main Libvmi README file ?

👍

Wenzel: Actually I'm afraid that duplicating the setup page in both repos will make the situation more difficult to maintain,
and since libvmi is based on (KVM + QEMU + libkvmi), I would rather keep the documentation on KVM-VMI's Wiki and focus my effort on keeping this up to date.

smaresca:
Echoing Tamas: I'm very happy to see this come to fruition. Thanks to all Mathieu especially but also to the substantial set of reviewers/contributors we've had on this one.

Wenzel:
Thanks @smaresca :) :ok_hand:
Shall we merge then ? no more blockers ?

PR #847, opened 2019-12-07
Initialize OS type from provided JSON profile

tklengyel:


tklengyel:

PR #849, opened 2019-12-07
Use PKG_CHECK_MODULES to find json-c library

tklengyel:


tklengyel:

PR #851, opened 2019-12-07
Arm cross compile

rageagainsthepc:


rageagainsthepc:
I am still having trouble with flex and bison because cmake tries to check their versions by executing them with `--version`. This does not work of course because those are arm binaries. @Wenzel maybe you have an idea?

mtarral:
@rageagainsthepc why did you closed this PR ?

rageagainsthepc:
I thought that your PR will make my PR obsolete :O

mtarral:
Not really, because my PR will test that libvmi compiles on an ARM CPU,
whereas yours test whether we can generate an ARM build on an x86 CPU, which is cross-compilation.

IIRC, @tklengyel , some of Libvmi users have this use case ? (building libvmi for ARM on x86, and then copying the build artifacts on the said ARM device)

tklengyel:
@mtarral for the most part we just wanted to test that LibVMI compiles for ARM. Actually doing cross-compilation is not necessary if we have the ability to compile the code on real ARM hardware. That said, if you can make the necessary changes to the cmake system to allow cross-compilation, I see that as a plus.

rageagainsthepc:
I have two issues:
1. I need to specify which repo in `/etc/apt/sources.list` should be searched for a given architecture. This means that I need to modify every line in addition to adding new ones. I could do this via a shell command in the test but it won't be pretty. Maybe someone has got a better idea?
2. Like I've mentioned before, the `--version` check has to be circumvented somehow. Should copy the standard cmake modules for flex and bison and modify them accordingly?

PR #853, opened 2019-12-11
travis: add arm64 architecture

mtarral:
Since october 2019, Travis has an alpha support for ARM:
https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support

This PR tests building LibVMI on `arm64` via Travis.


mtarral:
According to the [docs](https://docs.travis-ci.com/user/multi-cpu-architectures/)

Since we have a `matrix.include`, we need to specify the arch explicitely for each test, otherwise the first one of the array is taken (`amd64`)

> Please note, that explicitly included builds inherit the first value in an array:

~~~Yaml
arch:
  - amd64
  - arm64
  - ppc64le
  - s390x
jobs:
  include:
   - os: linux
     env: LIB_PATH="/usr/bin/shared/x86_64/v1"
   - os: linux
     env: LIB_PATH="/usr/bin/shared/x86_64/v2"
~~~

> For example, the above .travis.yml, would result in running both jobs with the environmental variable LIB_PATH assigned different values being run only on amd64 architecture.

mtarral:
@tklengyel the `arm64` travis test passed.

How do you want to work with the Travis config ?
I would recommend using the automatic matrix expansion, so that any test written will get duplicated for `amd64` and `arm64`.

Or we can keep the explicit `matrix.include` and add a few `arm64` tests.

tklengyel:
> so that any test written will get duplicated for amd64 and arm64

If we can get that working that would be ideal.

mtarral:
> If we can get that working that would be ideal.

I will make a few tests here to see how we can work this out.

Was it only ARM builds that were retaining the autotools in the repo ?
see also my question from https://github.com/libvmi/libvmi/pull/851#issuecomment-564625428

tklengyel:
No, there were additional regressions at the time with the cmake system that I didn't have time to deal with that lead me to putting back the automake scripts. In the meanwhile, I'm on the opinion of keeping both in place. I'm certainly not able to deal with cmake issues and I plain find it frustrating. So I'm not using it and don't plan on using it. But if some folks prefer it, I don't have anything against it being in the repo. I'm just not going to maintain it/fix it if it breaks in the future.

rageagainsthepc:
@tklengyel Could you elaborate where your frustration with cmake comes from?

tklengyel:
@rageagainsthepc I find it extremely hard to debug when it's not working. And there has been many times it failed for one reason or another.

rageagainsthepc:
Ok, so while I understand that your mind is pretty much set on this topic, let me just say that you could always open issues for those regressions and I will try to put the time in and get them resolved.

tklengyel:
@rageagainsthepc good to hear that. As I said, I don't have anything against having it in the tree, it's just I don't have the bandwidth and patience to deal with it. So if someone is willing to maintain it going forward, that's great!

mtarral:
@tklengyel so, I reworked this PR a bit, duplicated the tests for ARM, and only removed the Bareflank build test on the ARM side because it's not available for this architecture.

We have 2 tests failing, because of unused variables, due to conditional compilation:
https://travis-ci.org/github/libvmi/libvmi/jobs/664372721#L490

PR #856, opened 2019-12-17
Implement fast singlestep

skvl:
On break point event eight context switches occures.

With fast single step it is possible to shorten path for two context switches
and gain 35% spead-up.

### DIFF
--- a/libvmi/driver/xen/xen_events.c
+++ b/libvmi/driver/xen/xen_events.c
@@ -614,11 +614,15 @@ void process_response ( event_response_t response, vmi_event_t *event, vm_event_
                     case VMI_EVENT_RESPONSE_VMM_PAGETABLE_ID:
                         rsp->altp2m_idx = event->slat_id;
                         break;
-                    case VMI_EVENT_RESPONSE_EMULATE_NOWRITE:
+                    case VMI_EVENT_RESPONSE_FAST_SINGLESTEP:
+                        rsp->fast_singlestep.p2midx = event->next_slat_id;
+                        break;
+                    case VMI_EVENT_RESPONSE_EMULATE_NOWRITE: // TODO Remove redundant case
                         rsp->flags |= event_response_conversion[VMI_EVENT_RESPONSE_EMULATE];
                         break;
                     case VMI_EVENT_RESPONSE_SET_EMUL_READ_DATA:
                         if ( event->emul_read ) {
+                            // TODO Remove redundant conversion
                             rsp->flags |= event_response_conversion[VMI_EVENT_RESPONSE_EMULATE];
 
                             if ( event->emul_read->size < sizeof(event->emul_read->data) )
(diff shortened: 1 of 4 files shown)


skvl:
This needs a Xen patch from "[XEN PATCH v1 1/1] x86/vm_event: add fast single step"

tklengyel:
Very nice optimization, looking forward to getting this merged to Xen and the other projects!

libvmi/driver/xen/xen_events_abi.h
```diff
@@ -472,6 +477,7 @@ typedef struct vm_event_st_412 {
         struct vm_event_mov_to_msr_411        mov_to_msr;
         struct vm_event_desc_access           desc_access;
         struct vm_event_singlestep            singlestep;
+        struct vm_event_fast_singlestep       fast_singlestep;
```
tklengyel: You won't be able to add this to `vm_event_st_412`, or even `vm_event_st_413`. This feature will likely only get merged into Xen 4.14, so you'll need to base your changes on the current xen-unstable branch. In there, the vm_event structure has already changed, so we'll need to implement a new `vm_event_st_414` here. I usually wait with such changes to add support in LibVMI for until the Xen RC's are coming since there could be other changes made to the vm_event structure before the merge window is closed.

libvmi/driver/xen/xen_events.c
```diff
@@ -1883,6 +1888,9 @@ status_t process_requests_412(vmi_instance_t vmi, uint32_t *requests_processed)
         rsp->reason = vmec.reason;
         rsp->altp2m_idx = vmec.altp2m_idx;
 
+        if (rsp->flags & VM_EVENT_FLAG_FAST_SINGLESTEP)
+          rsp->u.fast_singlestep.p2midx = vmec.fast_singlestep.p2midx;
```
tklengyel: This will also likely will need to move to the new functions that will handle the Xen 4.14 vm_event ABI ( VM_EVENT_INTERFACE_VERSION 0x00000006).

libvmi/driver/xen/xen_events_abi.h
skvl: May be we could back-port the feature into https://github.com/tklengyel/xen ?

tklengyel: You can backport it to any version, but as long as its not upstream I wouldn't merge it to LibVMI master. So if you are using with Xen 4.12 I suggest you have a fork for that.

skvl: OK. I will fix it in a while.

libvmi/driver/xen/xen_events.c
tklengyel: I've added the plumbing for vm_event version 6 so this addition should be a lot easier.

skvl: Thank you!

tklengyel:
@skvl Patch merged to Xen http://xenbits.xen.org/gitweb/?p=xen.git;a=commit;h=f801b40d20c01eb41bed3157e266a18d34260259

tklengyel:
@skvl Are you planning to rebase this patch? Now that the Xen support is present in 4.14-unstable I think we could test and merge it.

skvl:
Hello. Sure. Thank You very much!

tklengyel:
Closing with #893 merged instead

PR #857, opened 2019-12-18
Add support to vm_event version 6

tklengyel:
Also switch internal Xen ABI structure and function names to reflect vm_event version instead of Xen versions.


tklengyel:

PR #868, opened 2020-01-13
Make accessible Windows build number

skvl:
For https://github.com/tklengyel/drakvuf/pull/766

### DIFF
--- a/libvmi/accessors.c
+++ b/libvmi/accessors.c
@@ -95,6 +95,45 @@ vmi_get_winver(
 #endif
 }
 
+uint16_t
+vmi_get_win_buildnumber(
+    vmi_instance_t vmi)
+{
+#ifndef ENABLE_WINDOWS
+    errprint("**LibVMI wasn't compiled with Windows support!\n");
+    return 0;
+#else
+
+#ifndef ENABLE_JSON_PROFILES
+    errprint("**LibVMI wasn't compiled with JSON profiles support!\n");
+    return 0;
+#endif
+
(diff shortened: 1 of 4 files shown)


libvmi/accessors.c
```diff
@@ -95,6 +95,34 @@ vmi_get_winver(
 #endif
 }
 
+uint16_t
+vmi_get_win_buildnumber(
```
tklengyel: The only issue I see is that this right now only works if a JSON profile was provided, in all other cases the value returned would just be 0. I would add a safety check to see if LibVMI was initialized with a Json profile or not (check if `vmi->json.handler == NULL` and print an error if it wasn't.

skvl: @tklengyel, done!

libvmi/accessors.c
```diff
@@ -95,6 +95,45 @@ vmi_get_winver(
...
+    errprint("**LibVMI wasn't compiled with Windows support!\n");
+    return 0;
+#else
+
+#ifndef ENABLE_JSON_PROFILES
+    errprint("**LibVMI wasn't compiled with JSON profiles support!\n");
+    return 0;
+#endif
+
+    if (vmi->json.handler == NULL) {
```
tklengyel: This should be moved to be under ` #ifdef ENABLE_SAFETY_CHECKS `

tklengyel:

libvmi/accessors.c
skvl: @tklengyel, done!

PR #869, opened 2020-01-13
Add tool to catch processes execution statistics in Windows

skvl:
This is used to determine most suitable process to inject into.

### DIFF
--- a/examples/win-proc-stat.c
+++ b/examples/win-proc-stat.c
@@ -1,0 +1,476 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+ * memory in a target virtual machine or in a file containing a dump of
+ * a system's physical memory.  LibVMI is based on the XenAccess Library.
+ *
+ * Author: Sergey Kovalev (valor@list.ru)
+ *
+ * This file is part of LibVMI.
+ *
+ * LibVMI is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or (at your
+ * option) any later version.
+ *
+ * LibVMI is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <glib.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <libvmi/libvmi.h>
+#include <libvmi/events.h>
+#include <json-c/json.h>
+
+/* Structure to pass data into event callback */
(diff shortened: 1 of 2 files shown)


examples/win-proc-stat.c
```diff
@@ -0,0 +1,476 @@
...
+
+static const char* win_offset_names[__WIN_OFFSETS_MAX][2] = {
+    [KPCR_PRCB] = {"_KPCR", "Prcb" },
+    [KPCR_PRCBDATA] = {"_KPCR", "PrcbData" },
+    [KPRCB_CURRENTTHREAD] = { "_KPRCB", "CurrentThread" },
+    [ETHREAD_CID] = {"_ETHREAD", "Cid" },
+    [CLIENT_ID_UNIQUETHREAD] = {"_CLIENT_ID", "UniqueThread" },
+};
+
+static bool rekall_lookup_in_json(
```
tklengyel: I don't think you should do this manually, there are already functions available in `libvmi_extra.h` that you can use to look up values in the json profiles, which works with either Rekall or Volatility profiles.

examples/win-proc-stat.c
```diff
@@ -0,0 +1,476 @@
...
+    g_hash_table_insert(config, g_strdup("os_type"), g_strdup("Windows"));
+    g_hash_table_insert(config, g_strdup("rekall_profile"), g_strdup(rekall_profile));
+
+    /* initialize the libvmi library */
+    vmi_instance_t vmi = NULL;
+    vmi_init_error_t err = VMI_SUCCESS;
+
+    if (VMI_FAILURE ==
+            vmi_init_complete(&vmi, domain, init_flags, NULL,
+                              VMI_CONFIG_GHASHTABLE, config, &err)) {
```
tklengyel: You can pass in the path directly to `vmi_init_complete` now, no need to create the GHashTable here. Just use VMI_CONFIG_JSON_PATH as the identifier. LibVMI can figure out the OS type from the profile too.

skvl: @tklengyel, done!

examples/win-proc-stat.c
```diff
@@ -0,0 +1,381 @@
...
+    GHashTable* stats;
+};
+
+/*
+ * Structures and functions to store stat data
+ */
+typedef union pid_tid {
+    struct {
+        uint64_t pid : 32;
+        uint64_t tid : 32;
```
tklengyel: This is bit odd looking, why not just use `uint32_t pid` & `uint32_t tid`?

examples/win-proc-stat.c
```diff
@@ -0,0 +1,381 @@
...
+                profile_json,
+                win_offset_names,
+                __WIN_OFFSETS_MAX,
+                offsets))
+        printf("Failed to find offsets for array of structure names and subsymbols.\n");
+
+    return offsets;
+}
+
+static int32_t get_current_thread_id(vmi_instance_t vmi, vmi_event_t* event)
```
tklengyel: I'm afraid you copied a large chunk of this code verbatim from DRAKVUF. You can't do that, the license of DRAKVUF is incompatible with LibVMI's LGPL. I would suggest you close this PR and add this as an extra tool to DRAKVUF using the DRAKVUF license.

PR #870, opened 2020-01-17
Allow @ character in FILENAME token in libvmi config file parser

disaykin:


tklengyel:

PR #892, opened 2020-04-22
windows: get major and minor version numbers

skvl:


tklengyel:
Looking at this I get the feeling that we really ought to combine all these `winver` functions we have right now. Currently we have:
```
vmi_get_winver
vmi_get_win_buildnumber
vmi_get_winver_str
vmi_get_winver_manual
```
and now we would have to two more.

How about we declare all the above ones deprecated and introduce a single function that can get you all these. Say `vmi_get_windows_build_info`. That returns a single structure that has all these infos in it?

skvl:
Sounds great!

skvl:
@tklengyel I have added `vmi_get_windows_build_info` and remove some other functions.

Though I don't remove `vmi_get_winver` as a convenient shortcut for a lot of places in _drakvuf_.

Also I don't remove `vmi_get_winver_str` and `vmi_get_winver_manual` because it could be used in special cases.

tklengyel:
Please don't remove existing functions, we just mark them deprecated in a comment for now. Other people might be actively using them and removing them as such would break their builds.

skvl:
@tklengyel fixed!

tklengyel:

libvmi/accessors.c
```diff
@@ -63,6 +63,51 @@ vmi_get_ostype(
         vmi->os_type;
 }
 
+bool
+vmi_get_windows_build_info(
+    vmi_instance_t vmi, win_build_info_t* info)
+{
+    if (!info) {
```
tklengyel, line 70: This should be under ENABLE_SAFETY_CHECKS

skvl: @tklengyel, done!

tklengyel:
Thanks, looks good!

libvmi/os/windows/core.c
```diff
@@ -84,24 +86,24 @@ win_ver_t pe2version(vmi_instance_t vmi, addr_t kernbase_pa)
...
-            major_os_version=oh32->major_os_version;
-            minor_os_version=oh32->minor_os_version;
+            *major=oh32->major_linker_version;
+            *minor=oh32->minor_linker_version;
             break;
         case IMAGE_PE32_PLUS_MAGIC:
-            major_os_version=oh32plus->major_os_version;
-            minor_os_version=oh32plus->minor_os_version;
+            *major=oh32plus->major_linker_version;
+            *minor=oh32plus->minor_linker_version;
```
tklengyel, line 94: This subtle change from `major_os_version` to `major_linker_version` is causing breakage for Windows 10 VMs

PR #895, opened 2020-05-07
Switch pe2version back to using major_os_version

tklengyel:
Fix regression introduced in #892

/cc @skvl


skvl:
@tklengyel I'm very sorry for the stupid error.

tklengyel:
No worries, just took me a while to even notice the change. I was staring at the patch for like half an hour :)

PR #896, opened 2020-05-07
Add SIGINT handler for win-offsets

chivay:
In some cases `vmi-win-offsets` may fail and start to consume a lot memory.
On the other hand stopping it with Ctrl+C leaves the VM frozen.
Adding the SIGINT handler allows to manually stop it and exit cleanly.

### DIFF
--- a/examples/win-offsets.c
+++ b/examples/win-offsets.c
@@ -32,12 +32,27 @@
 #include <stdio.h>
 #include <inttypes.h>
 #include <glib.h>
+#include <signal.h>
 
+vmi_instance_t vmi;
+GHashTable* config;
+
+void clean_up(void)
+{
+    vmi_resume_vm(vmi);
+    vmi_destroy(vmi);
+    if (config)
+        g_hash_table_destroy(config);
+}
+
+void sigint_handler()
+{
+    clean_up();
+    exit(1);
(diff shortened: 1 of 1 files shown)


icedevml:
@drakvuf-jenkins This is OK to test.

icedevml:
:+1: the exact reason is that sometimes LibVMI has trouble finding the kernel and goes for a full scan, user interruption is causing pause request being leaked and the VM pause count remains = 1 forever

examples/win-offsets.c
```diff
@@ -32,12 +32,21 @@
...
 #include <glib.h>
+#include <signal.h>
+
+vmi_instance_t vmi;
+
+void sigint_handler()
+{
+    vmi_resume_vm(vmi);
+    vmi_destroy(vmi);
+    exit(1);
```
tklengyel, line 43: You would leak the `config` ghashtable if you exit here. It's not like it would matter for this particular application but this sets a bad example. Just make sure that ghashtable gets freed, make the `done:` part into a standalone function and call it in both main and the sigint handler.

chivay: :+1:

chivay:
Fixed.

PR #898, opened 2020-05-13
Harden vmi-win-offsets against KPTI

### LINKED ISSUE #897: LibVMI for Windows 10 doesn't always initialize properly with KPTI enabled
From what I've understood, LibVMI's `windows_init` is searching for kernel's virtual address in order to find its Page Global Directory (KPGD) physical location in the next step.

We've encountered some stability problems particularly with Windows 10, that LibVMI randomly gets into trouble finding right kernel VA/KPGD and starts doing a full scan (sometimes even a few different full scans).

I belive this is due to the fact that with KPTI the kernel may not be mapped at all, when we are in usermode and this may confuse the LibVMI search procedure. From what I've experimented, there is a strong correlation between being in user mode and full scans.

A hint may be to detect that we caught the VM in user mode (this snippet is useful for that: https://github.com/libvmi/libvmi/pull/797/files ) and retry until we catch it in a kernel mode. This is however pretty tricky - if there is some program doing simple computations consuming 100% CPU time, then the kernel is active for a small fraction of time and it is really hard to randomly catch it.

/cc @chivay and thanks for help debugging/understanding this

icedevml:
When doing LibVMI init, normally LibVMI is just pausing the target VM at arbitrary moment and then inspecting VCPU=0 context, trying to compute all necessary parameters, like kernel VA. Unfortunately this doesn't work well with KPTI - if the VM gets paused while VCPU=0 is executing some usermode code, there is no such thing as "kernel VA" anymore because it's not mapped for security reasons. In such case, LibVMI gets confused and has a trouble finding offsets.

This PR tries to harden `vmi-win-offsets` against KPTI by introducing additional preflight checks:

1. We do partial initialization of the LibVMI just to have the JSON profile initialized, but doing any OS-specific offset initialization.
2. The CR3 event is created and we inspect all such events for VCPU=0.
3. If we detect that we've just paused inside the system process, we clear the event, leave the VM paused and perform proper LibVMI initialization.

resolve #897

again thanks @chivay

### DIFF
--- a/examples/win-offsets.c
+++ b/examples/win-offsets.c
@@ -22,8 +22,14 @@
 
 #define _GNU_SOURCE
 
+#define LIBVMI_EXTRA_JSON
+
 #include <libvmi/libvmi.h>
 #include <libvmi/peparse.h>
+#include <libvmi/events.h>
+#include <libvmi/libvmi_extra.h>
+#include <json-c/json.h>
+
 #include <stdlib.h>
 #include <string.h>
 #include <stdbool.h>
@@ -33,13 +39,36 @@
 #include <inttypes.h>
 #include <glib.h>
 #include <signal.h>
+#include <unistd.h>
+#include <getopt.h>
 
 vmi_instance_t vmi;
 GHashTable* config;
+vmi_event_t cr3_event;
+
+int find_pid4_success = 0;
+unsigned long found_vcpu_id;
+
+addr_t offset_kpcr_prcb;
+addr_t offset_kprcb_currentthread;
+addr_t offset_eprocess_uniqueprocessid;
+addr_t offset_kthread_process;
+
+int enable_debug = 0;
+
+void dp(const char* format, ...)
+{
+    va_list argptr;
+    va_start(argptr, format);
+
+    if (enable_debug)
+        vfprintf(stderr, format, argptr);
+
+    va_end(argptr);
+}
 
 void clean_up(void)
 {
-    vmi_resume_vm(vmi);
     vmi_destroy(vmi);
     if (config)
         g_hash_table_destroy(config);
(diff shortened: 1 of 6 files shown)


icedevml:
I've tested it with Windows 10 build 1909 and Windows 7 x64. For both cases it works properly.

Probably some adjustment is needed for 32 bit systems, I will check that later.

icedevml:
Got it working for Windows 10 build 1909, Windows 7 both x86 and x64. Unfortunately there is some crash occurring from time to time, related issue #899.

tklengyel:
So I'm a little confused, if you already have the kernel's json generated, why do you need the windows offsets? The offsets getting defined in `libvmi.conf` is a legacy feature that I think should get deprecated eventually.

libvmi/libvmi.h
```diff
@@ -847,6 +847,34 @@ os_t vmi_init_os(
...
+ * @param[out] error Optional. If not NULL and the function returns VMI_OS_UNKNOWN,
+ *                   this will specify the stage at which initialization failed.
+ * @param[in] os_specific_init If set to FALSE - only initialize the JSON profile
+ *                             and paging, but don't perform OS-specific offset
+ *                             initialization.
+ * @return VMI_OS_UNKNOWN when the configuration didn't work for the VM, otherwise
+ *         the OS type that has been initialized (ie. VMI_OS_WINDOWS or
+ *         VMI_OS_LINUX).
+ */
+os_t vmi_init_os_partial(
```
tklengyel, line 871: I don't particularly like this function. It is very confusing. As I understand you just want to init paging (there is already `vmi_init_paging` for that and the JSON parsing functions. I would be fine with adding a function that sets that up without all the other OS-init stuff LibVMI does but that should go into `libvmi_extra.h`. For example, we have `json_object* vmi_get_kernel_json(vmi_instance_t vmi) NOEXCEPT;`, would make sense to me to add `vmi_set_kernel_json` with the file path being passed as input.

icedevml:
> So I'm a little confused, if you already have the kernel's json generated, why do you need the windows offsets?

I'm looking for runtime offsets, the ones that may change every Windows boot (like kernel virtual address or KPGD).

When creating a snapshot in DRAKVUF Sandbox, we just want to determine these offsets at once, store them and use for further LibVMI initializations. At the moment we can't rely on automated initialization because it fails pretty often for Windows 10 with KPTI enabled ( https://github.com/libvmi/libvmi/issues/897 ).

tklengyel:
> I'm looking for runtime offsets, the ones that may change every Windows boot (like kernel virtual address or KPGD). It's not about the JSON offsets of structure members.

Ah OK, makes sense, that performance optimization trick to speed up the LibVMI setup time.

icedevml:
Not only performance but also stability. Code from this PR also ensures that the VM is paused inside `System` process before calling thr actual LibVMI Windows initialization. This makes sure that Kernel Page Table Isolation doesn't get in our way, which sometimes may cause LibVMI not being able to initialize at all.

icedevml:
@tklengyel I think now it is adjusted according to your remarks.

icedevml:
Outionally this
https://github.com/libvmi/libvmi/blob/cfacfbd845474f9e6a3e3693033ef6622251f90e/examples/win-offsets.c#L92-L146
and this
https://github.com/libvmi/libvmi/blob/cfacfbd845474f9e6a3e3693033ef6622251f90e/examples/win-offsets.c#L261-L286

can be implemented as a function:

```
vmi_pid_t vmi_get_current_pid(vmi_instance_t vmi, unsigned long vcpu)
```

which could be placed inside `libvmi_extra.h` if that sounds fair. Otherwise I propose to leave this as it is, but we will have to copy the abovementioned code to DRAKVUF.

tklengyel:
>  but we will have to copy the abovementioned code to DRAKVUF.

Why? In DRAKVUF we already have `drakvuf_get_current_process` and `drakvuf_get_process_pid`.

examples/win-offsets.c
```diff
@@ -51,57 +78,164 @@ void sigint_handler()
...
+    };
+
+    /*
+     * Get current GS_BASE and translate its VA to PA using current CR3.
+     * This may fail (most probably) if we are in user-mode DTB with KPTI hardening.
+     */
+    page_mode_t pm = vmi_get_page_mode(vmi, 0);
+    reg_t kpcr_reg = pm == VMI_PM_IA32E ? GS_BASE : FS_BASE;
+
+    if (VMI_FAILURE == vmi_get_vcpureg(vmi, &gs_base, kpcr_reg, 0)) {
```
tklengyel, line 108: Why not just use the registers that are passed down as part of the event?

examples/win-offsets.c
```diff
@@ -51,57 +78,164 @@ void sigint_handler()
...
 }
 
+event_response_t cr3_cb(vmi_instance_t vmi, vmi_event_t *event)
+{
+    dp("CR#: %llx, VCPU: %d\n", event->reg_event.value, event->vcpu_id);
+
+    if (event->vcpu_id != 0) {
+        /* LibVMI initialization procedure strongly
+         * relies on reading CPU context for VCPU 0
+         * so we only do care about this one. */
```
tklengyel, line 88: I guess this could also be adjusted. Nothing technically requires this to be vCPU 0, we just never bothered to update the code to perhaps check the other vCPUs or to pass it in as an input if we have a known vCPU with a good pagetable for the init to use. So I think going through all this trouble just to catch it on vCPU is a bit silly.

icedevml:
> Why? In DRAKVUF we already have drakvuf_get_current_process and drakvuf_get_process_pid.

Hm. You're 100% right, it's implemented pretty much the same way so I bet it would work OK.

examples/win-offsets.c
icedevml: Yes, VCPU=0 is hardcoded through whole initialization part. This could be easily changed in such way that user could hint which VCPU should be inspected.

libvmi/libvmi_extra.h
```diff
@@ -56,6 +56,18 @@ GSList* vmi_get_va_pages(
...
+ * Initialize kernel's JSON profile from particular config type.
+ * @param[in] vmi Instance
+ * @param[in] config_mode The type of OS configuration that is provided.
+ * @param[in] config Configuration is passed directly to LibVMI (ie. in a string
+ *                   or in a GHashTable) or NULL of global config file is used.
+ */
+os_t vmi_init_profile(
+    vmi_instance_t vmi,
+    vmi_config_t config_mode,
+    void *config) NOEXCEPT;
```
tklengyel, line 69: I don't understand this function. You wanted to initialize the JSON interface, but then why would this function take a string or a ghashtable as input instead of a `const char* path` to the json file?

tklengyel: I guess I would be OK with this since it matches the other init functions input params. But I think you need to add more information what this function actually enables and what won't work the way a user would expect. For example, while enabling the json parser routines will get you offsets and structure sizes, things like `vmi_get_symbol_addr_from_json` won't work the same way as when the os part of LibVMI is initialized since we don't yet know the KASLR offset.

So I think there needs to be better description before we can add this to the API.

tklengyel: Actually never mind that example, for some reason I thought that function adjusts the values to KASLR but that's the other LibVMI symbol translation function that does that. I guess I should get some sleep :P

icedevml: I've added some more explanations.

examples/win-offsets.c
icedevml: fixed

examples/win-offsets.c
icedevml: I've added `windows->init_vcpu` in the `windows_instance`. Using this value user may hint LibVMI to inspect the context of particular vCPU when doing `vmi_init_os()`. If not set, it's 0 by default, so it should be compatible with the old behavior of checking VCPU=0.

icedevml:
I've altered the code so it doesn't rely on VCPU=0 anymore. The `cr3_cb` will be satisfied PID=4 being executed on any VCPU. LibVMI will be hinted about which VCPU context it should inspect during initialization. This hint is passed through config key:

https://github.com/libvmi/libvmi/blob/077cc8aadbfb4ba929cf871a76c21d5fbb511ce1/examples/win-offsets.c#L304-L306

icedevml:
Hm, there must be some mistake in the code, DRAKVUF doesn't play well with this PR. I will check it.

icedevml:
Ok, it was unrelated. There is some mistake happening when debug prints are enabled, fixed in #903

icedevml:
Ok. I've tested DRAKVUF on Windows 10 with KPTI. The `vmi-win-offsets` seems to work OK and it allows to obtain the KPGD value which can be then plugged into DRAKVUF using `-k <kpgd>` switch. When this value is manually provided, DRAKVUF also doesn't experience any startup problems anymore.

I've implemented a convenience `-k` switch here, which alters the program behavior to only print the KPGD value, so this can be easily used in combo with DRAKVUF:

```
KPGD=$(vmi-win-offsets name vm-1 -r $PROFILE -k)
drakvuf -k $KPGD -d vm-1 -r $PROFILE
```

examples/win-offsets.c
```diff
@@ -51,76 +79,241 @@ void sigint_handler()
...
+        }
+
+    if (argc - optind != 2) {
+        show_usage(argv[0]);
+        return 1;
+    }
 
-    if (strcmp(argv[1],"name")==0) {
-        domain = (void*)argv[2];
+    if (strcmp(argv[optind],"name")==0) {
```
tklengyel, line 211: Can we switch this to just use standard `getopt` flags? Like `--domain` and `--domid`?

libvmi/os/windows/core.c
```diff
@@ -610,6 +610,11 @@ void windows_read_config_ghashtable_entries(char* key, gpointer value,
         goto _done;
     }
 
+    if (strncmp(key, "win_init_vcpu", CONFIG_STR_LENGTH) == 0) {
```
tklengyel, line 613: The init vCPU should be passed in as `VMI_INIT_DATA_INIT_VCPU` instead of via the ghashtable since that only works with `VMI_CONFIG_GHASHTABLE`

tklengyel:

libvmi/os/windows/core.c
icedevml: I think if the appropriate key would be added to ` libvmi/config/grammar.y` then it will also work with `VMI_CONFIG_GLOBAL_FILE_ENTRY` and `VMI_CONFIG_STRING`.

Also `vmi_init_os` doesn't accept `vmi_init_data_t *init_data` as an argument. It is only accepted by `vmi_init` or `vmi_init_complete`. In case of `vmi-win-offsets` tool, we don't yet know KPGD value at the moment when we call `vmi_init()`.

examples/win-offsets.c
icedevml: done

libvmi/os/windows/core.c
icedevml: I've added an entry to `grammar.y`. I think it would be much easier this way, as `vmi_init_data_t` can not be easily used here, for reasons I've outlined above.

icedevml:
After some additional testing I've found a problem in `kpcr_find1`. Looks like it somehow implicitly assumes that we are on VCPU=0. Gonna also fix this one.

icedevml:
I will add some extra checks whether `SelfPcr` really points to our KPCR and whether `KPCR.Number` is 0. Otherwise this is not initial PCR and the first method can't be used.

libvmi/os/windows/core.c
tklengyel: True.. but then the problem is this can't be used with VMI_CONFIG_JSON_PATH.. I think it would make more sense to just update the OS init code in LibVMI to try all vCPUs instead of just vCPU 0.

icedevml:
@tklengyel I've modified this PR one more time. Instead of `init_vcpu` we could just pass the CR3 of any system process as `win_kpgd` parameter and then perform `vmi_init_os`. This obviously will be not the true KPGD but LibVMI is aware of that and will correct it to the proper value. Such solution also works ok with KPTI and is much simpler in terms of code.

About reworking LibVMI to inspect all VCPUs in a loop - I was digging into it for a few hours, but I suppose that we do need much more sanity checks in the initialization process to accomplish that. It's just that in some corner cases, some methods return `VMI_SUCCESS` but are actually setting improper values in `windows_instance` structure. This propagates further and the initialization process is confused. I don't want to make too big changes in the initialization process itself, because eventually we may decrease stability due to regressions. Maybe later and in a separate PR.

icedevml:
So this PR would just fix `vmi-win-offsets` in particular and I will think about the rest when I have some more power/time/sth

tklengyel:
```
New defect(s) Reported-by: Coverity Scan
Showing 3 of 3 defect(s)


** CID 358748:    (PRINTF_ARGS)


________________________________________________________________________________________________________
*** CID 358748:    (PRINTF_ARGS)
/home/travis/build/libvmi/libvmi/examples/win-offsets.c: 85 in cr3_cb()
79         clean_up();
80         exit(1);
81     }
82     
83     event_response_t cr3_cb(vmi_instance_t vmi, vmi_event_t *event)
84     {
>>>     CID 358748:    (PRINTF_ARGS)
>>>     Argument "event->vcpu_id" to format specifier "%lu" was expected to have type "unsigned long" but has type "unsigned int".
85         dp("CR#: %llx, VCPU: %lu\n", event->reg_event.value, event->vcpu_id);
86     
87         if (find_pid4_success) {
88             dp("Skip callback as we already found System process\n");
89             goto failed;
90         }
/home/travis/build/libvmi/libvmi/examples/win-offsets.c: 154 in cr3_cb()
148          */
149         if (pid != 4) {
150             dp("Current PID=%llx, skip until we reach PID=4\n", (unsigned long long)pid);
151             goto failed;
152         }
153     
>>>     CID 358748:    (PRINTF_ARGS)
>>>     Argument "event->vcpu_id" to format specifier "%lu" was expected to have type "unsigned long" but has type "unsigned int".
154         dp("Stopped inside system process, VCPU %lu, CR3=%llx\n", event->vcpu_id, event->reg_event.value);
155         find_pid4_success = 1;
156         found_cr3 = event->reg_event.value;
157     
158         /*
159          * Remove CR3 event and leave the VM paused inside System process,

PR #900, opened 2020-05-14
Rework vmi-dump-memory

chivay:
At the beginning I wanted to add support for pausing the VM when dumping memory, however I've found some bigger issues (it seems that `--sparse` is not working because  `if (argc != 3)` makes the program exit, missing null getopt at the end of the array causing UB, etc.).

Summary:
* Reworded page as frame, since it seems more appropriate in this context
* Added working sparse file support for both unmapped frames and zero-only frames
* Added pausing the VM as default behavior (this seems reasonable)
* Added SIGINT handler to cleanly stop the program
* Added `--progress` switch for tracking progress every 32MiB dumped
* Added `--no-pause` switch to disable VM pausing if it's desired

### DIFF
--- a/examples/dump-memory.c
+++ b/examples/dump-memory.c
@@ -24,171 +24,206 @@
  * along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include <config.h>
-#include <libvmi/libvmi.h>
 #define _GNU_SOURCE
+
 #include <stdlib.h>
 #include <string.h>
-#include <sys/mman.h>
 #include <stdio.h>
 #include <limits.h>
 #include <getopt.h>
+#include <signal.h>
+#include <sys/mman.h>
+#include <config.h>
+#include <libvmi/libvmi.h>
 
-#define PAGE_SIZE 1UL << 12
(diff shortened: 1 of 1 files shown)


tklengyel:

tklengyel:
```
** CID 358747:  Control flow issues  (UNREACHABLE)
/home/travis/build/libvmi/libvmi/examples/dump-memory.c: 65 in bareflank_setup()
59         printf("Using this example on Bareflank is not safe.\n");
60         printf("You have to adjust it to match your machine before running it.\n");
61         return 1;
62     
63         /* The following is an example based on the e820 map described in
64          * notes/memory_map.txt. */
>>>     CID 358747:  Control flow issues  (UNREACHABLE)
>>>     This code cannot be reached: "uint32_t e820_entries = 5U;".
65         uint32_t e820_entries = 5;
66     
67         memory_map_t *memmap = malloc(sizeof(memory_map_t) + sizeof(addr_t) * 2 * e820_entries);
68         memmap->count = e820_entries;
69         memmap->range[0][0] = 0xfff;
70         memmap->range[0][1] = 0x57fff;

PR #907, opened 2020-05-29
Off-by-one error in xen driver renders last physical page inaccessible

thomasdangl:
The max_physical_address member of the vmi_instance data structure must by definition point past the end of the last physical page.
However, in the current implementation of the Xen driver, this member points *at* the last physical page, therefore rendering it inaccessible as we hit a sanity check in the [memory cache](https://github.com/libvmi/libvmi/blob/master/libvmi/driver/memory_cache.c#L123) for HVM guests.

### DIFF
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -790,7 +790,7 @@ xen_get_memsize(
     }
 
     *allocated_ram_size = XC_PAGE_SIZE * pages;
-    *max_physical_address = xen->max_gpfn << XC_PAGE_SHIFT;
+    *max_physical_address = (xen->max_gpfn + 1) << XC_PAGE_SHIFT;
 
     return VMI_SUCCESS;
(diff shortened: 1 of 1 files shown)


icedevml:
Thanks for noticing this.

Side note: I would also add that if you allocate some extra physical pages for the VM dynamically, these are also inaccessible if LibVMI is compiled with memory cache enabled, due to the exact same sanity check you've mentioned, so also be aware of that.

thomasdangl:
Thanks for your hint. I fixed this issue by calling vmi_get_max_physical_address after allocating the new page, but before accessing them through the cache.
If this is something we would ever want to support in the library, we should look for a proper fix.

icedevml:
> If this is something we would ever want to support in the library, we should look for a proper fix.

Yup, this is a "not implemented yet" thing.

tklengyel:
Thanks!

PR #913, opened 2020-07-10
do not skip the pid=1 entry in tasklist

dl9rdz:
The code of #787 starts with curr_proc pointing to inittask (pid=0), curr_entry to the linked list of the next (pid=1).
The first loop iteration analyzes curr_proc (pid=0), then advances curr_entry to the linked list of the successor (pid=2) and curr_proc to that entry (pid=2).  pid=1 (the second entry of the task list) is never used.

The simple fix starts with curr_entry pointing to the linked list of the inittask.


tklengyel:

tklengyel:
Thanks!

PR #916, opened 2020-07-30
Travis: Use clang 10 for static analysis 

Wenzel:
* add `SCAN_BUILD` cmake var to configure `scan-build` binary used for static analysis
on Ubuntu, multiple version of the `clang-tools` package are available:
- `clang-tools`
- `clang-tools-6`
- `clang-tools-7`
- etc..

this variable allow to control which `scan-build` binary will be used.

* update static analysis in Travis to use `clang-tools-7`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -142,9 +142,10 @@ jobs:
       install:
         - sudo apt-get update
         - sudo apt-get -q -y install bison flex libjson-c-dev libvirt-dev libxen-dev libfuse-dev
+        - sudo apt-get install clang-tools-10
       script:
         - mkdir build && cd build
-        - cmake ..
+        - cmake -DSCAN_BUILD='scan-build-10' ..
         - make static_analysis_test
 
 #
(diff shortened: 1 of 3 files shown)


tklengyel:
Isn't clang 7 the default already?

Wenzel:
I believe that clang 6 is the default:
https://packages.ubuntu.com/bionic/devel/clang-tools

However, we can go up to `clang-tools-10`:
https://packages.ubuntu.com/bionic/clang-tools-10

But It spotted small issues in the Xen driver that I wasn't sure how to solve.
I haven't had a deep look though.

Wenzel:
I updated the PR to use `clang-tools-10` and fixed the issues raised in the Xen driver.

mtarral:
I don't understand why Travis refuses to trigger a new build

tklengyel:
I don't see any error on the Travis side either.

Wenzel:
I added a comment just to change the commit and to be able to push --force.
Travis works now :man_shrugging: :man_shrugging: :man_shrugging:

PR #918, opened 2020-08-09
New changes to support win7, and rectify some SYSENTER based examples

0xarash:
LibVMI not work on some of the newer build of Windows 7, so I added a symbol to solve this.

Also another matter is that it seems like there are some aliases for SYSENTER registers instead of using MSR index directly. using MSR registers directly caused event-example.c (and some other) to fail on KVM, I changed the registers to be similar to xen.c.


tklengyel:

PR #920, opened 2020-08-10
Pause VM in Xen HVM setregs calls for safety

tklengyel:
As the comment in these functions noted, calling these while the VM is unpaused was unsafe. It is unclear why we never made these functions safe from the start, but since the VM pause/unpause calls are reference counted in Xen it is always safe to do this, the VM only actually gets resumed if it was running at the start of the function.


tklengyel:

PR #921, opened 2020-08-22
json handlers: updated to look for struct fields recursively

mleinhos:
Enables searching for fields buried in embedded, anonymous unions or structures, e.g. `mm_struct` or `page`. Note that recent-ish Linux kernel changes have put fields of interest in `mm_struct` inside an anonymous struct, which have complicated their discovery from the VMI side. For example, check out https://elixir.bootlin.com/linux/v5.3.18/source/include/linux/mm_types.h#L370


mleinhos:
@tklengyel It checks out with the more thorough testing.

PR #924, opened 2020-09-11
json profiles: recurse down all embedded structs to find field

mleinhos:
The conditional recursion previously merged doesn't handle a case in an old kernel. In 2.6, dentry contains a named union, `d_u`. Rather than making assumptions about whether it's an anonymous struct that holds the field the caller wants, we just recurse down every field. Sure, it's a tad slower, but this function shouldn't be used after initialization.


tklengyel:

PR #934, opened 2020-10-29
some slice fix about compile compatibility

npc1054657282:
1. support glib2 < 2.40 in libvmi/events.c (or a compile error occured under low version glib2)
2. support to compile examples together with libvmi.so, beacuse some examples use C99 feature
3. add an enable_kvm_legacy patch with qemu-2.10


tklengyel:

PR #940, opened 2020-11-06
kvm: sanity check mem_access flags for invalid EPT configurations

Wenzel:
* adds sanity check for `vmi_mem_access_t` flags in `kvm_set_mem_access`, as the Xen driver does already.

fix https://github.com/KVM-VMI/libvmi/issues/38

### DIFF
--- a/libvmi/driver/xen/xen_events_private.h
+++ b/libvmi/driver/xen/xen_events_private.h
@@ -61,6 +61,7 @@
 #include <xenctrl.h>
 #include <libvmi/events.h>
 
+#include "arch/intel.h"
 #include "xen_events_abi.h"
 
 /*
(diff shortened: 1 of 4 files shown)


libvmi/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set(libvmi_src
     arch/arm_aarch32.c
     arch/arm_aarch64.c
     driver/driver_interface.c
+    driver/driver_utils.c
```
tklengyel, line 18: I think this would be more properly abstracted into the `arch` interface given the EPT settings are architecture-specific, not hypervisor specific.

libvmi/events.h
```diff
@@ -45,6 +45,9 @@ extern "C" {
 #pragma GCC visibility push(default)
 
 #include <stdbool.h>
+#include <inttypes.h>
+
+#include <libvmi/libvmi.h>
```
tklengyel, line 50: Unrelated changes, please remove

libvmi/driver/xen/xen_events_private.h
```diff
@@ -153,30 +154,10 @@ static const unsigned int event_response_conversion[] = {
...
-    }
-
-    return VMI_SUCCESS;
-}
-
 static inline status_t
 convert_vmi_flags_to_xenmem(vmi_mem_access_t page_access_flag, xenmem_access_t *access)
 {
-    if ( VMI_FAILURE == vmi_flags_sanity_check(page_access_flag) )
+    if ( VMI_FAILURE == intel_mem_access_sanity_check(page_access_flag) )
```
tklengyel, line 160: I guess at some point we also want to ensure this check is only performed when we are actually running on x86.

PR #941, opened 2020-11-07
Fatal bug fix about cmake

npc1054657282:
1. In Cmake, The variable in subdirectories cannot be found in parent directories. That is, even if LibvmiRequest is found, HAVE_LIBVMI_REQUEST cannot be defined in config.h.
It is a first-aid fix and the coding in cmake is not very pretty after the fix. I think the /CMakeLists.txt , /libvmi/CMakeLists.txt , and /cmake/modules/*.cmake should be optimized throughly in the future.
2. In kvm_legacy.c, there is a function 'ceil', which needs the library 'libm'. 
It seems that, without HAVE_LIBVMI_REQUEST, the branch to the function 'ceil' will be optimized by gcc. When I use cmake to test kvm_legacy with HAVE_LIBVMI_REQUEST(after I fix another bug), I meet the error with unreferenced symbol 'ceil'. To add the link library 'libm' can resolve it.
/*3. I find there are two config flags "HAVE_HVMMEM_ACCESS_T", "HAVE_XENMEM_ACCESS_T" in config.h.in, but they are never set in all CMakelists.txt. Though I do not test libvmi on xen, but it seems a bit strange and I would open an issue about it.*/
Sorry, I find the "check_type_size" in FindXen.cmake, so the third point is not an issue.


tklengyel:
Thanks!

PR #942, opened 2020-11-20
Fool patchguard fix mem event

Wenzel:
fix fool-patchguard example:

- fix syscall entry detection
- fix emulating custom read buffer
- decode instructions with [libddisasm](https://github.com/bitdefender/bddisasm)
- determine read access size
- detect Patchguard when read instruction is a `xor` (assuming)
- tested on Windows 7 x64 and Windows 10 x64


examples/fool-patchguard.c
```diff
@@ -53,14 +142,117 @@ event_response_t read_callback(vmi_instance_t vmi, vmi_event_t *event)
...
+        NdToText(&rip_insn, 0, sizeof(insn_str), insn_str);
+        printf("Read on KiServiceTable %s entry (size: %ld)\n", cb_data->target, overlap.size);
+        printf("\t0x%"PRIx64": %s\n", event->x86_regs->rip, insn_str);
+
+        // assume PatchGuard if read with a XOR
+        if (rip_insn.Instruction == ND_INS_XOR) {
+            printf("\tXOR read: Patchguard check !\n");
+        }
+
+        emul_read_t emul_read = {0};
```
tklengyel, line 232: This is not good. If you pass `emul_read` with `dont_free` it needs to live on the heap or needs to be global. If it's on the stack nothing guarantees that it won't get corrupted when this function returns.

tklengyel:
So I think this is OK but looks to me it's way over-engineered. Why do you need to disassemble to determine the read size?

A single memory read can go as high as 8-bytes. Let's say you hooked a pointer (8-byes) then any read at or near the hooked pointer would need to be emulated:

```
| --7 bytes-- | -- hooked 8 bytes -- | --7-bytes-- |
| -- 15-bytes area needs emul  -- |  -- 8 bytes -- |
```

So you just have a backup of that 22-bytes and if the EPT fault triggers in the 15-bytes area where emulation is needed you just copy 8-bytes from this backup to the read emul buffer. Then you don't have to care whether the read was actually for single byte or up to 8 bytes.

vlutas:
> So I think this is OK but looks to me it's way over-engineered. Why do you need to disassemble to determine the read size?
> 
> A single memory read can go as high as 8-bytes. Let's say you hooked a pointer (8-byes) then any read at or near the hooked pointer would need to be emulated:
> 
> ```
> | --7 bytes-- | -- hooked 8 bytes -- | --7-bytes-- |
> | -- 15-bytes area needs emul  -- |  -- 8 bytes -- |
> ```
> 
> So you just have a backup of that 22-bytes and if the EPT fault triggers in the 15-bytes area where emulation is needed you just copy 8-bytes from this backup to the read emul buffer. Then you don't have to care whether the read was actually for single byte or up to 8 bytes.

I agree with you @tklengyel that for this particular use-case, it can be simplified, but a single-memory read could exceed 8-bytes - for example, if the kernel decides to relocate a particular page to another physical-address, you may get 16 or even 32-byte reads via SSE/AVX instructions from memcpy & friends.

tklengyel:
@vlutas Hm, and those still only trigger a single EPT violation? Last I experimented with this anything larger then an 8-byte fetch triggered a separate EPT violation.

vlutas:
You get at most one EPT violation per accessed physical address per operand. If the accessed operand would cross a page boundary (for example, 8 bytes in page P1 and the next 8 bytes in page P2), then you could get two distinct EPT violations (first one on P1, and a second EPT violation on P2, if the first one is resolved). 
In addition, you usually get an EPT violation for the first accessed address that triggers one - for example, if you have a **MOVSQ** instructions that copies 8 bytes from **[RSI]** to **[RDI]**, you would get a read EPT violation on **[RSI]**, if the page is not readable. Even if the page at **[RDI]** is not writable, you would only get an EPT violation on it if you somehow handlde the first EPT violation, and re-execution of the **MOVSQ** instruction doesn't trigger it anymore (for example, by marking the **[RSI]** page as readable). If you have a **REP MOVSQ** instruction, then you would indeed get an EPT violation for each 8 bytes accessed, but this is a special case.
In essence, it's the same with large access. A single 16/32/64 bytes access would trigger a single EPT violation, as long as the operand is contained within the same single-page, because a single page-walk is made for it.

tklengyel:
Cool stuff, makes total sense. I just did some experiments with sse2 and indeed I'm fetching 128-bytes and trigger only a single EPT violation. TIL :)

examples/fool-patchguard.c
Wenzel: I fixed this with a global variable for `emul_read`

Wenzel:
Regarding the over-engineering aspect of this example, it's true, but it also serve as a demonstration of how things should be done right in a bigger app, how to handle the corner cases, etc.
I could add a note regarding this in the callback, saying the we could have done this much simpler but for the sake of the completeness and demonstration, we kept the disassembler.

What do you think ?

examples/fool-patchguard.c
```diff
@@ -28,22 +28,115 @@
...
 #include <libvmi/libvmi.h>
 #include <libvmi/events.h>
 
+// maximum size of an x86 instruction
+#define MAX_SIZE_X86_INSN 15
+#define KISERVICE_ENTRY_SIZE sizeof(uint32_t)
+
+// we use emul_read with 'dont_free' so it can't be allocated on the stack
+// as it could be corrupted
+static emul_read_t g_emul_read;
```
tklengyel, line 43: ` = { .dont_free = 1 };`

tklengyel: I'm also thinking that it's perhaps time to change this ABI a bit and instead of having a `dont_free` field, just supply a pointer to a `free` function if one needs to be called.

examples/fool-patchguard.c
```diff
@@ -53,14 +146,118 @@ event_response_t read_callback(vmi_instance_t vmi, vmi_event_t *event)
...
+    if (is_zone_read(&read_zone, &syscall_entry_zone, &overlap)) {
+        // overlap !
+        // get insn string
+        char insn_str[ND_MIN_BUF_SIZE];
+        NdToText(&rip_insn, 0, sizeof(insn_str), insn_str);
+        printf("Read on KiServiceTable %s entry (size: %ld)\n", cb_data->target, overlap.size);
+        printf("\t0x%"PRIx64": %s\n", event->x86_regs->rip, insn_str);
+
+        memset(&g_emul_read, 0, sizeof(g_emul_read));
+        g_emul_read.dont_free = true;
```
Wenzel, line 232: @tklengyel I was setting the `dont_free` attribute here, as I reset the struct before.
I will change this to `1` since it's an `int`

libvmi/driver/kvm/kvm_events.c
```diff
@@ -135,7 +135,7 @@ process_cb_response_emulate(
                             rpl->pf.ctx_size = libvmi_event->emul_read->size;
                             // set linear address
                             // TODO: ARM support
-                            rpl->pf.ctx_addr = libvmi_event->x86_regs->rip;
+                            rpl->pf.ctx_addr = libvmi_event->mem_event.gla;
```
tklengyel, line 138: This seems to be an unrelated fix for a bug.

Wenzel: right, i will make a separate PR

tklengyel:
@Wenzel I think the example would be a lot more meaningful if instead of corrupting the ssdt - which is kinda pointless - you instead actually breakpointed one of the syscall functions and protected that. I understand that the example just ough to demonstrate how to fool patchguard but then might as well make it more useful. Perhaps you can even extend it and just breakpoint all syscall functions (fyi https://j00ru.vexillium.org/syscalls/nt/64).

examples/fool-patchguard.c
tklengyel: If a variable is a static global then you don't need to initialize it to 0, it's already initialized to 0.

Wenzel:
> @Wenzel I think the example would be a lot more meaningful if instead of corrupting the ssdt - which is kinda pointless - you instead actually breakpointed one of the syscall functions and protected that. I understand that the example just ough to demonstrate how to fool patchguard but then might as well make it more useful. Perhaps you can even extend it and just breakpoint all syscall functions (fyi https://j00ru.vexillium.org/syscalls/nt/64).

I agree, I just wanted to test whether the API actually worked in this example at the beginning, that's where it came from.
It could be a good demo to have an example breakpointing all syscall and avoiding patchguard detection.
Let me try to implement it, in another PR

Wenzel:
So, I was looking at implementing the example with breakpoints on all syscalls.
The thing is that I wanted to go with on the instruction emulation route, to fake the original opcode replaced by the breakpoint.
However, the KVMi API doesn't allow one to emulate a custom buffer on interrupt events, it's only available for pagefault events at this point.

We could switch to a recoil-based implementation for breakpoints (singlestep, etc) but it's tedius to implement, and implies that we pause the guest to avoid race conditions, so it won't scale.

Wenzel:
In the meantime I pushed the fixes according to your comments.

tklengyel:
That seems to be an implementation limitation only for KVM and should be easy enough fix to by allow that emulation information to be returned for more event types (in this case software interrupts). I don't have anything against merging an example with a limited functionality, just saying it would be nicer to get these things fixed and have something that has more utility.

tklengyel:
Could you add the example to be built with autotools as well?

PR #943, opened 2020-11-30
kvm: fix pf ctx_addr for SET_EMUL_READ_DATA event response

Wenzel:
After a documentation clarification of the KVMi API:
- https://github.com/KVM-VMI/kvm/pull/42
- https://github.com/KVM-VMI/kvm/pull/43

this PR fixes the context address used when emulating custom read data as a libvmi event response for KVM.
The context address that should be used is the one where the read actually happens, and not where RIP is located.

Related comment on another PR: https://github.com/libvmi/libvmi/pull/942#discussion_r531627986


Wenzel:
Wow, Travis has been running since ` Running for 6 hrs 8 min 59 sec `, and not finished yet.
What the heck

PR #958, opened 2021-01-30
Debug print in volatility_ist can segfault, fix

tklengyel:


icedevml (approved):
Huh, I've seen this kind of mistake in debug prints so many times and it's still there. Maybe some full code review should be done in terms of debug prints.

tklengyel:
@icedevml we do test builds with debug options enabled but evidently the static analysis isn't spotting all such potential issues `¯\_(ツ)_/¯`

icedevml:
huh

PR #990, opened 2021-09-07
Fix reading of long windows unicode strings (length > 4KiB)

disaykin:
Fix reading of long windows unicode strings (length > 4KiB).
Add sanity check that length <= maximum_length.


disaykin:
Compilation test must be successful after merging of #991.

disaykin:
@tklengyel  This MR fixes some cases with long names in drakvuf

PR #992, opened 2021-09-22
Introduce kvmi support for vmifs

thomasdangl:
The current vmifs implementation does not yet support the new kvmi driver.
This pull requests adds the missing support.


tools/vmifs/vmifs.c
```diff
@@ -147,11 +148,18 @@ int main(int argc, char *argv[])
         return 1;
     }
 
-    if (VMI_FAILURE == vmi_get_access_mode(NULL, domain, init_flags, NULL, &mode))
+    if (argc == 5) {
+        init_data = malloc(sizeof(vmi_init_data_t) + sizeof(vmi_init_data_entry_t));
+        init_data->count = 1;
+        init_data->entry[0].type = VMI_INIT_DATA_KVMI_SOCKET;
+        init_data->entry[0].data = strdup(argv[4]);
```
Wenzel, line 155: you need to free the `strdup` memory on error / exit, as well as the `malloc` call above

like this:
https://github.com/libvmi/libvmi/blob/master/examples/process-list.c#L269

PR #995, opened 2021-10-29
Include libvmi.h in events.h

rageagainsthepc:
Right now the order of includes matters. For example, you cannot include `events.h` before `libvmi.h` because it requires types defined in `libvmi.h`. This is especially a problem if you want to use code formatters that sort includes alphabetically (aside from the usual reasons for avoiding such behavior). I therefore propose moving all typedefs in `libvmi.h` to a separate header.

### DIFF
--- a/libvmi/events.h
+++ b/libvmi/events.h
@@ -45,6 +45,7 @@ extern "C" {
 #pragma GCC visibility push(default)
 
 #include <stdbool.h>
+#include <libvmi/libvmi.h>
 
 /*---------------------------------------------------------
  * Event management


tklengyel:
I don't think this is the way to go about it. IMHO would be a lot saner to just make `events.h` include `<libvmi/libvmi.h>`

rageagainsthepc:
Sure, I could do that. It feels a little bit odd though, considering `events.h` will then expose most of the functionality contained in this library. I've always thought of `libvmi.h` as the main header. But then again, if you are going to use the functionality provided by `events.h` you will always need `libvmi.h` as well. 🤔

rageagainsthepc:
I've implemented the changes. For the record, I think this is bad design because we are effectively merging the two APIs defined in `libvmi.h` and `events.h` just because we need access to some custom types. If this is desired though, why not merge `events.h` into `libvmi.h` all together?

tklengyel:
Because libvmi.h is already huge plus events are not supported on all hypervisors. I don't see any downside in ensuring libvmi.h is properly included in events.h since events.h is not a separate api but an extension of libvmi.h

libvmi/events.h
```diff
@@ -46,6 +46,8 @@ extern "C" {
 
 #include <stdbool.h>
 
+#include "libvmi.h"
```
tklengyel, line 49: No, you should not use `""` in a public header. It must be `<>`

libvmi/os/windows/core.c
```diff
@@ -73,7 +73,7 @@ win_ver_t pe2version(vmi_instance_t vmi, addr_t kernbase_pa, uint16_t* major, ui
     uint16_t optional_header_type = 0;
     struct optional_header_pe32 *oh32 = NULL;
     struct optional_header_pe32plus *oh32plus = NULL;
-    uint8_t pe[VMI_PS_4KB];
+    uint8_t pe[VMI_PS_4KB] = {0};
```
tklengyel, line 76: Unrelated change

libvmi/os/windows/peparse.c
```diff
@@ -437,7 +437,7 @@ peparse_get_export_table(
     size_t export_header_size = 0;
 
 #define MAX_HEADER_BYTES 1024   // keep under 1 page
-    uint8_t image[MAX_HEADER_BYTES];
+    uint8_t image[MAX_HEADER_BYTES] = {0};
```
tklengyel, line 440: Unrelated change

libvmi/os/windows/core.c
rageagainsthepc: Without this change the CI will not pass. I don't know why it did not happen before, were the build images updated recently? I'm going to open another PR instead.

libvmi/events.h
rageagainsthepc: Is that part of some convention I am not aware of? With `""` the preprocessor will look in the current directory of the file first which is exactly where `libvmi.h` will be located. It's not like we are using some public header of another project.

tklengyel: What are you talking about? This is a public header

rageagainsthepc: Let me phrase it differently: `libvmi.h` and `events.h` wil be deployed to the same directory (e.g. `/usr/include/libvmi`). If the preprocessor sees `#include "libvmi.h"` in `events.h` it will look for `libvmi.h` in the same directory as `events.h` before it will consider any other include rules. We are not dependent on any global include paths here, at least not to my knowledge. That being said, I don't mind using `<>` here although I think it would make more sense to use `""`. I've already changed it btw.

Issue #1002, opened 2021-12-10
Build fails on Fedora 35 due to bad -L setting


flyn-org:
Building libvmi-0.14.0 on Fedora 35 with the following commands results in the indicated error:
```
$ mkdir build
$ cd build
$ cmake ..
$ make VERBOSE=1
[...]
/usr/bin/cc  -Wall -Wextra -g -g3 -O0  -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -rdynamic CMakeFiles/map-addr.dir/map-addr.c.o -o map-addr  -Wl,-rpath,/home/mike/Download/libvmi-0.14.0/build/libvmi ../libvmi/libvmi.so.0.0.14 -L/usr/lib64glib-2.0 
/usr/bin/ld: ../libvmi/libvmi.so.0.0.14: undefined reference to `g_direct_hash'
/usr/bin/ld: ../libvmi/libvmi.so.0.0.14: undefined reference to `g_tree_new_full'
/usr/bin/ld: ../libvmi/libvmi.so.0.0.14: undefined reference to `g_slist_free_full'
[...]
collect2: error: ld returned 1 exit status
```
Notice the use of `-L/usr/lib64glib-2.0`. A `/` is missing from the path, and the directory contains headers, not libraries.

flyn-org:
I am not sure if this is a Fedora bug. See also https://bugzilla.redhat.com/show_bug.cgi?id=2031182.

Wenzel:
hi @flyn-org,
I had a similar issue while trying to fix a bug on the Python bindings:
https://github.com/libvmi/python/pull/76#issuecomment-983403966

Could you try to compile with autotools to check if it's not a CMake issue ? (I'm thinking some weird string replacement that 's gone wrong)

besser82:
See https://src.fedoraproject.org/rpms/libvmi/c/3d132c9990c0377fcf1b7c7faea159d4d5e9722a for a fix.

Wenzel:
Thanks @besser82 
So these are the problematic lines:
https://github.com/libvmi/libvmi/blob/master/CMakeLists.txt#L32
https://github.com/libvmi/libvmi/blob/master/libvmi/CMakeLists.txt#L40

This has been added to cleanup the get a "clean" `glib_ldflags`, in order for to be used when generating the libtool file:
The libtool libraries dependencies are appended in this variable, by prefixing with `-l` each time
[...]

PR #1003, opened 2021-12-15
Fix handling of glib link libraries in cmake (#1002)

besser82:
The `target_link_libraries` command is supposed to be called using the contents of `GLIB_LIBRARIES` instead of `GLIB_LDFLAGS`.

Fixes: #1002

Signed-off-by: Björn Esser <besser82@fedoraproject.org>


besser82:
I've reworked my PR to be less intrusive about the minimum required cmake version.  Using `GLIB_LIBRARIES` and not `GLIB_LINK_LIBRARIES`, the minimum cmake version can stay to be 3.1.

@Wenzel Can you please re-trigger the test pipeline?

PR #1007, opened 2021-12-23
Fix static analyzer warning about usage of enum value in boolean cont…

disaykin:
…ext in ENTRY_PRESENT(vmi->os_type, ...)

### DIFF
--- a/libvmi/arch/amd64.c
+++ b/libvmi/arch/amd64.c
@@ -319,7 +319,7 @@ GSList* get_pages_ia32e(vmi_instance_t vmi, addr_t npt, page_mode_t npm, addr_t 
 
                 uint64_t pgd_value = pgd_page[pgde_index];
 
-                if (ENTRY_PRESENT(vmi->os_type, pgd_value)) {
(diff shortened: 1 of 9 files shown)


libvmi/arch/amd64.c
```diff
@@ -319,7 +319,7 @@ GSList* get_pages_ia32e(vmi_instance_t vmi, addr_t npt, page_mode_t npm, addr_t
 
                 uint64_t pgd_value = pgd_page[pgde_index];
 
-                if (ENTRY_PRESENT(vmi->os_type, pgd_value)) {
+                if (ENTRY_PRESENT(vmi->os_type != VMI_OS_UNKNOWN, pgd_value)) {
```
tklengyel, line 322: This should be `VMI_OS_WINDOWS == vmi->os_type` here and all other spots. Entry present needs to know if tranisition pages are in use and those are Windows specific.

tklengyel (changes requested):

libvmi/arch/amd64.c
disaykin: Done

tklengyel (approved):

PR #1010, opened 2022-02-15
Disk reading interface for Xen based VMs and example

blsvntn:
Hello! This PR adds support for reading disks of Xen virtual machines.
Disk identification is based on reading xenstore `/local/domain/<domain_id>/device/vbd` properties.

List of disks identifiers can be retrieved via `vmi_get_disks` function. Next, disk ID string is passed to `vmi_read_disk` function.
Supported disk formats:
- raw
- QCOW2
For QCOW2 disks it can detect if clusters are compressed and decompress Zlib.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -145,31 +145,19 @@ pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libvmi.pc
 
 AM_CPPFLAGS = -I$(top_srcdir)/libvmi
-if WITH_ZLIB
-    AM_CPPFLAGS += $(ZLIB_CFLAGS)
-endif
+AM_CPPFLAGS += $(ZLIB_CFLAGS)
 
 AM_CFLAGS = $(CFLAGS)
-if WITH_ZLIB
-    AM_CFLAGS += $(ZLIB_CFLAGS)
-endif
+AM_CFLAGS += $(ZLIB_CFLAGS)
 
 AM_YFLAGS = -y -d
 AM_LDFLAGS = $(LDFLAGS)
-if WITH_ZLIB
-    AM_LDFLAGS += $(ZLIB_LIBS)
-endif
+AM_LDFLAGS += $(ZLIB_LIBS)
 
 lib_LTLIBRARIES= libvmi/libvmi.la
 libvmi_libvmi_la_SOURCES= $(h_public) $(h_private) $(drivers) $(os) $(c_sources)
-libvmi_libvmi_la_CFLAGS= -fvisibility=hidden $(GLIB_CFLAGS) $(JSONC_CFLAGS)
-if WITH_ZLIB
-    libvmi_libvmi_la_CFLAGS += $(ZLIB_CFLAGS)
-endif
-libvmi_libvmi_la_LDFLAGS= $(GLIB_LIBS) $(JSONC_LIBS) -no-undefined
-if WITH_ZLIB
-    libvmi_libvmi_la_LDFLAGS += $(ZLIB_LIBS)
-endif
+libvmi_libvmi_la_CFLAGS= -fvisibility=hidden $(GLIB_CFLAGS) $(JSONC_CFLAGS) $(ZLIB_CFLAGS)
+libvmi_libvmi_la_LDFLAGS= $(GLIB_LIBS) $(JSONC_LIBS) $(ZLIB_LIBS) -no-undefined
 
 libvmi_libvmi_la_LDFLAGS+= -version-info $(VERSION)
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -145,14 +145,11 @@ pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libvmi.pc
 
 AM_CPPFLAGS = -I$(top_srcdir)/libvmi
-AM_CPPFLAGS += $(ZLIB_CFLAGS)
 
 AM_CFLAGS = $(CFLAGS)
-AM_CFLAGS += $(ZLIB_CFLAGS)
 
 AM_YFLAGS = -y -d
 AM_LDFLAGS = $(LDFLAGS)
-AM_LDFLAGS += $(ZLIB_LIBS)
 
 lib_LTLIBRARIES= libvmi/libvmi.la
 libvmi_libvmi_la_SOURCES= $(h_public) $(h_private) $(drivers) $(os) $(c_sources)
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,6 +91,8 @@ if WITH_XEN
     h_public    += libvmi/events.h
     drivers     += libvmi/driver/xen/altp2m.c \
                    libvmi/driver/xen/altp2m_private.h \
+                   libvmi/driver/xen/vbd.c \
+                   libvmi/driver/xen/vbd_private.h \
                    libvmi/driver/xen/xen.h \
                    libvmi/driver/xen/xen_private.h \
                    libvmi/driver/xen/xen.c \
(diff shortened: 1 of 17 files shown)


tklengyel:
Thanks for the PR! I think this is a worthwhile addition. My only recommendation would be to abstract out the vbd portion of the code from under the `driver/xen` to a new `disk` folder. There is no reason to keep that tied to Xen, might be handy for other setups too. Looks to me the only Xen specific info is finding the disk path through Xenstore, so that makes sense to be kept in the Xen-specific driver base, but the rest could be their own thing.

blsvntn:
Thanks for the reply, Tamas! I moved disk-reading code from `driver/xen` to `disk` folder, also xenstore specific functions to `driver\xen\xen.c`

examples/read-disk-example.c
```diff
@@ -0,0 +1,165 @@
+/* The LibVMI Library is an introspection library that simplifies access to
+ * memory in a target virtual machine or in a file containing a dump of
+ * a system's physical memory.  LibVMI is based on the XenAccess Library.
+ *
+ * Copyright 2011 Sandia Corporation. Under the terms of Contract
+ * DE-AC04-94AL85000 with Sandia Corporation, the U.S. Government
+ * retains certain rights in this software.
+ *
+ * Author: Bryan D. Payne (bdpayne@acm.org)
```
tklengyel, line 9: I think the Author is different + you can remove the Sandia blob above ;)

libvmi/disk/vbd.c
```diff
@@ -0,0 +1,518 @@
...
+        fclose(f);
+        return VMI_FAILURE;
+    }
+
+    fclose(f);
+
+    return VMI_SUCCESS;
+}
+
+static int xen_vbd_qcow2_uncompress_cluster(unsigned char *dest, size_t dest_size, unsigned char *src, size_t src_size)
```
tklengyel, line 56: I don't see anything Xen specific in this function except its name

libvmi/disk/vbd.c
```diff
@@ -0,0 +1,518 @@
...
+        ret = 0;
+    } else {
+        ret = -EIO;
+    }
+
+    return ret;
+}
+
+/* Free L1 table and close file object */
+static void xen_vbd_qcow2_close(QCowFile *qcowfile)
```
tklengyel, line 83: Same here, why would this be Xen specific?

libvmi/disk/vbd.c
```diff
@@ -0,0 +1,518 @@
...
+* You should have received a copy of the GNU Lesser General Public License
+* along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <zlib.h>
+
+#include "private.h"
+#include "driver/xen/xen.h"
+#include "driver/xen/xen_private.h"
+#include "vbd_private.h"
```
tklengyel, line 26: I would prefer `disk/vbd_private.h` here.

tklengyel:
This looks largely fine now, please just address the issues spotted by the CI and the couple comments I left. Thanks!

blsvntn:
I see that some issues remains, trying to figure out how to fix them

configure.ac
```diff
@@ -267,6 +267,7 @@ AM_CONDITIONAL([MAKE_TESTS], [test x$have_check = xyes])
 
 PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.16],[],[AC_MSG_ERROR(GLib 2.16 or newer not found. Install missing package and re-run)])
 PKG_CHECK_MODULES([JSONC], [json-c], [have_jsonc='yes'], [have_jsonc='no'])
+PKG_CHECK_MODULES([ZLIB], [zlib], [], [AC_MSG_ERROR([zlib not found])])
```
tklengyel, line 270: I don't see why ZLIB is being added as a dependency, where is it being used?

blsvntn: It is used to inflate compressed cluster in `vbd_qcow2_uncompress_cluster` (`libvmi/disk/vbd.c`).

tklengyel: I see. Please make the zlib dependency optional, if its not installed the qcow2 parsing should just be disabled. Thanks

Makefile.am
```diff
@@ -145,19 +145,32 @@ pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libvmi.pc
 
 AM_CPPFLAGS = -I$(top_srcdir)/libvmi
-AM_CPPFLAGS += $(ZLIB_CFLAGS)
+if WITH_ZLIB
```
tklengyel, line 148: No need to make these conditional here, if zlib is not found the flags will just be empty

configure.ac
```diff
@@ -293,6 +294,10 @@ AC_CHECK_LIB(json-c, json_object_get_uint64, [AC_DEFINE([JSONC_UINT64_SUPPORT],
     [then]
         AC_CHECK_HEADERS([xenstore.h xs.h])
         AC_DEFINE([HAVE_LIBXENSTORE], [1], [Define to 1 to enable Xenstore support.])
+        [if test "$have_zlib" = "yes"]
+        [then]
+            AC_DEFINE([HAVE_ZLIB], [1], [Define to 1 to enable QCOW2 disk parsing.])
```
tklengyel, line 299: `PKG_CHECK_MODULES` should already define in config.h whether zlib is found, you don't need to do this

libvmi/disk/vbd.c
```diff
@@ -0,0 +1,501 @@
...
+* LibVMI is distributed in the hope that it will be useful, but WITHOUT
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+* FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
+* License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <zlib.h>
```
tklengyel, line 21: You'll need to wrap this in an ifdef to make sure it compiles even if zlib isn't installed

configure.ac
blsvntn: I've tried to remove explicit defininition, but zlib didn't appear in config.h

tklengyel: Ah yes sorry it sets the cflags and libs variables not config.h. Still, you could just define HAVE_ZLIB in the PKG_CONFIG_CHECH line:

```
PKG_CHECK_MODULES([ZLIB], [zlib],
   [AC_DEFINE([HAVE_ZLIB], [1], [Found ZLIB])],[])

File: configure.ac [hunk 9]
```diff
@@ -267,6 +267,7 @@ AM_CONDITIONAL([MAKE_TESTS], [test x$have_check = xyes])
 
 PKG_CHECK_MODULES([GLIB], [glib-2.0 >= 2.16],[],[AC_MSG_ERROR(GLib 2.16 or newer not found. Install missing package and re-run)])
 PKG_CHECK_MODULES([JSONC], [json-c], [have_jsonc='yes'], [have_jsonc='no'])
+PKG_CHECK_MODULES([ZLIB], [zlib], [AC_DEFINE([HAVE_ZLIB], [1], [Found ZLIB])], [])
```
tklengyel, line 270: Looks like on Centos the pkg-config macro requires the last `[]` to be filled-in with something so that it doesn't error out. Just set it to `[AC_MSG_NOTICE(No zlib found)]`

Makefile.am
```diff
@@ -143,13 +145,20 @@ pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libvmi.pc
 
 AM_CPPFLAGS = -I$(top_srcdir)/libvmi
+AM_CPPFLAGS += $(ZLIB_CFLAGS)
```
tklengyel, line 148: I don't think you need to add this to the pre-processor flags here

Makefile.am
```diff
@@ -143,13 +145,20 @@ pkgconfigdir = $(libdir)/pkgconfig
...
+
+AM_CFLAGS = $(CFLAGS)
+AM_CFLAGS += $(ZLIB_CFLAGS)
```
tklengyel, line 151: This is also not required. For the library you already add `ZLIB_CFLAGS` below.

Makefile.am
```diff
@@ -143,13 +145,20 @@ pkgconfigdir = $(libdir)/pkgconfig
...
+
 AM_YFLAGS = -y -d
 AM_LDFLAGS = $(LDFLAGS)
+AM_LDFLAGS += $(ZLIB_LIBS)
```
tklengyel, line 155: Same for the libs.

libvmi/driver/xen/xen.c
```diff
@@ -165,7 +166,183 @@ xen_put_memory(
...
+xen_get_vbd_state(
+    vmi_instance_t vmi,
+    const char* device_id)
+{
+    unsigned int len;
+    bool result = false;
+    xen_instance_t *xen = xen_get_instance(vmi);
+    xs_transaction_t xth = XBT_NULL;
+
+    struct xs_handle *xsh = xen->libxsw.xs_open(0);
```
tklengyel, line 191: Any particular reason you open a new xs handle and not just use the one already open?

blsvntn: `xen_get_vbd_state` function is called at the beginning of `xen_read_disk` and there is no opened handle at this time, or am i missing something?

tklengyel: The xen instance has an active xenstore handle after `xen_init`.

PR #1022, opened 2022-04-13
Build vbd only when xen is enabled

realhezhe:
vbd was added for xen based VMs since
bdee00fac9b4 ("Disk reading interface for Xen based VMs and example")
and should be built only when xen is enabled, otherwise there would not be
necessary xen headers and cause the following failure.

xen_private.h:38:10: fatal error: xenctrl.h: No such file or directory

Signed-off-by: He Zhe <zhe.he@windriver.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,8 +17,7 @@ h_private = \
     libvmi/os/os_interface.h \
     libvmi/driver/driver_interface.h \
     libvmi/driver/driver_wrapper.h \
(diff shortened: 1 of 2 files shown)


tklengyel:
I don't think this is the right solution, I think vbd.c should just not include the Xen headers. The vbd code itself should be hypervisor agnostic.

tklengyel:
Thanks, I merged a separate PR that should fix the issue.

PR #1024, opened 2022-05-04
Add kvm support for swap/clear events

rageagainsthepc:
This PR aims to align the behavior of `kvm_events_listen` to the behavior of `xen_events_listen`. This includes handling vmi clear and swap events, but also making sure all pending events are processed after waiting for the next event.


tklengyel:
LGTM

PR #1026, opened 2022-05-05
Pop pending events after pausing vm

rageagainsthepc:
In some edge cases the pause event might not be the first event in the queue. This causes `kvm_resume_vm` to fail. Therefore we should process all pending events immediately after pausing the vm and let the pause event be inserted into `pause_events_list`.

### DIFF
--- a/libvmi/driver/kvm/kvm.c
+++ b/libvmi/driver/kvm/kvm.c
@@ -993,7 +993,7 @@ kvm_pause_vm(
 {
     kvm_instance_t *kvm = kvm_get_instance(vmi);
     // already paused ?
-    if (kvm->expected_pause_count)
+    if (kvm->paused)
         return VMI_SUCCESS;
 
     // pause vcpus
@@ -1002,9 +1002,9 @@ kvm_pause_vm(
         return VMI_FAILURE;
     }
 
-    kvm->expected_pause_count = vmi->num_vcpus;
+    kvm->paused = true;
 
-    dbprint(VMI_DEBUG_KVM, "--We should receive %u pause events\n", kvm->expected_pause_count);
+    dbprint(VMI_DEBUG_KVM, "--We should receive %u pause events\n", vmi->num_vcpus);
 
     return VMI_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


rageagainsthepc:
Apparently it's not possible to process pause events immediately in `kvm_pause_vm`, but we should at least process all events right before resuming so we can be sure that all pause events are waiting in `pause_events_list`.

rageagainsthepc:
Fyi, I've just discovered that there is still a bug in the implementation.

Wenzel:
@rageagainsthepc can you describe the bug you found ?

rageagainsthepc:
@Wenzel Sure. My assumption that all pause events have been sent by kvm by the time we call `kvm_resume_vm` was apparently incorrect. I assume that's why you wait for new events until you reach `expected_pause_count`. I am going to rework this PR soon.

libvmi/driver/kvm/kvm.c
```diff
@@ -1027,54 +1016,35 @@ kvm_resume_vm(
...
-        // if no pause event is waiting in the list, pop next one
-        if (!ev) {
-            if (VMI_FAILURE == kvm_get_next_event(kvm, &ev, 1000)) {
-                errprint("Failed to get next KVMi event\n");
-            }
-            if (!ev) {
-                // no new events
-                // report error
+        // wait until the pause event for the current vcpu is received
+        while (!kvm->pause_events_list[vcpu]) {
```
rageagainsthepc, line 1027: I wonder: Maybe it makes sense to wait for all pause events before starting to unpause vcpus? Or is it irrelevant that there may be a couple of milliseconds between two `reply_continue()` calls? :thinking:

rageagainsthepc:
It seems to work for me now. I guess this is ready for review again. :+1:

libvmi/driver/kvm/kvm.c
Wenzel: I don't think it's necessary to wait for all VCPU pause events to be received here
if you have 64 VCPUs, you want to resume them ASAP.

Wenzel:
LGTM on the review, but I haven't tested the PR.

rageagainsthepc:
What about this PR? Is there something left to be done?

PR #1030, opened 2022-05-11
vbd: Add support for JSON format of qcow2 backing file description

disaykin:
Example:

    'json:{"file": {"driver": "file", "filename": "/tmp/file.qcow2"}}'

### DIFF
--- a/libvmi/disk/vbd.c
+++ b/libvmi/disk/vbd.c
@@ -167,11 +167,11 @@ static status_t vbd_qcow2_open(QCowFile *qcowfile, const char *filename)
             g_stpcpy(qcowfile->backing_file, backing_file_name);
         } else if (!strncmp(backing_file_name, "json:", 5)) {
             char prefix[] = "\"filename\": \"";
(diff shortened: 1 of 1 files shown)


libvmi/disk/vbd.c
```diff
@@ -165,6 +165,14 @@ static status_t vbd_qcow2_open(QCowFile *qcowfile, const char *filename)
         /* Check if backing file has absolute path then simply copy it to structure */
         if (!strncmp(backing_file_name, "/", 1)) {
             g_stpcpy(qcowfile->backing_file, backing_file_name);
+        } else if (!strncmp(backing_file_name, "json:", 5)) {
+            char prefix[] = "\"filename\": \"";
+            char *filename = strstr(backing_file_name, prefix);
```
tklengyel, line 170: Please use a different variable name, there is already a `filename` in this scope as function input.

disaykin: Done

PR #1037, opened 2022-05-30
Do not require macro definitions for libvmi_extra.h

rageagainsthepc:
Users of `libvmi_extra.h` are required to define various macros although the underlying functionality should be activated via feature flags.

### DIFF
--- a/libvmi/libvmi_extra.h
+++ b/libvmi/libvmi_extra.h
@@ -22,15 +22,16 @@
  * @file libvmi_extra.h
  * @brief The Extra LibVMI API is defined here.
  *
- * To use GLib functions compile with -DLIBVMI_EXTRA_GLIB
- * To use JSON functions compile with -DLIBVMI_EXTRA_JSON
+ * To use JSON functions compile with -DENABLE_JSON_PROFILES
  */
 #ifndef LIBVMI_EXTRA_H
 #define LIBVMI_EXTRA_H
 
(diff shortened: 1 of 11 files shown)


libvmi/libvmi_extra.h
```diff
@@ -22,15 +22,16 @@
...
- * To use GLib functions compile with -DLIBVMI_EXTRA_GLIB
- * To use JSON functions compile with -DLIBVMI_EXTRA_JSON
+ * To use JSON functions compile with -DENABLE_JSON_PROFILES
  */
 #ifndef LIBVMI_EXTRA_H
 #define LIBVMI_EXTRA_H
 
-#ifdef LIBVMI_EXTRA_GLIB
+#ifdef HAVE_CONFIG_H
+#include <config.h>
```
tklengyel, line 31: This is a public header and config.h cannot be included here.

tklengyel:
I think you are missing the point of the current setup in `libvmi_extra.h`. There are features only available to external tools linking with LibVMI if they also link with some other third party library (glib, json-c). Just linking with LibVMI isn't sufficient to compile using these functions, you must also link with the respective third party lib. At the same time, we don't want to force all users to have to do that. So if you want to use these functions you must link with the required library AND declare the respective define to make the functions available.

rageagainsthepc:
Oh, you're right. I've totally missed this. That makes sense. Never mind then. :smile:

PR #1038, opened 2022-06-02
Kvm: Fix events not being processed during shutdown 

rageagainsthepc:


rageagainsthepc:
But perhaps there is a cleaner way to fix this. I agree that it is weird to call user defined callbacks during shutdown.

rageagainsthepc:
I would say something is clearly wrong here because comments like this one https://github.com/libvmi/libvmi/blob/master/libvmi/driver/kvm/kvm_events.c#L173 indicate that we expect to process events during shutdown, yet this piece of code can never be reached during shutdown.

libvmi/driver/kvm/kvm_events.c
```diff
@@ -170,10 +170,6 @@ process_cb_response(
         errprint("%s: invalid kvm or kvmi handles\n", __func__);
         return VMI_FAILURE;
     }
-    // Note: libvmi_event can be NULL
```
rageagainsthepc, line 173: Correct me if I'm wrong but afais this comment is not accurate because `events_destroy` is called after `driver_destroy`.

libvmi/driver/kvm/kvm_events.c
```diff
@@ -231,10 +227,12 @@ call_event_callback(
     vmi_instance_t vmi,
     vmi_event_t *libvmi_event)
 {
-    event_response_t response;
-    vmi->event_callback = 1;
-    response = libvmi_event->callback(vmi, libvmi_event);
-    vmi->event_callback = 0;
+    event_response_t response = VMI_EVENT_RESPONSE_NONE;
```
rageagainsthepc, line 230: Maybe @Wenzel can weigh in on this. Is it save to set `VMI_EVENT_RESPONSE_NONE` as a default response for all event types?

Wenzel: I think we should mimic the same behavior that already exists in the Xen driver.

libvmi/driver/kvm/kvm_events.c
Wenzel: That's correct, I reviewed the code and `events_destroy()` is called after, so the ghashtables should contain the libvmi events

Wenzel:
So, If I understand correctly, the intent of this PR is to have a consistent behavior on `vmi_destroy` on both KVM and Xen, regarding how events are destroyed.

On xen, when [`xen_events_destroy()`](https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L3622) is called, it will [listen](https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L3657), [skip](https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L3301) waiting for new events, and [process the responses](https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L3322).

The KVM driver we are currently [skipping the event handling entirely](https://github.com/libvmi/libvmi/blob/master/libvmi/driver/kvm/kvm_events.c#L923), including sending a response, and this PR fixes that. The user callbacks are not called, but the events will get their replies.

libvmi/driver/kvm/kvm_events.c
rageagainsthepc: That's simply not possible since pauses are also events and they don't come through if you have single stepping activated, for example. Also, from the user side this will behave the same as the xen driver. At least that was the goal here.

rageagainsthepc:
@Wenzel If I understood correctly the xen driver doesn't have to wait for and process events while shutting down. So this behavior would actually be a little bit different here. But if the kvm driver doesn't reply to pending events it might never receive pause events which is a problem for the cleanup code. The single step example actually never really worked without producing errors (at least not to my knowledge). After one of my PRs the code has been restructured in such a way that it waits indefinitely for pause events as soon as we try to resume the vm. With these changes the single step example doesn't crash anymore but hangs indefinitely. This PRs fixes this bevhavior by still replying to all kvm events. This works in the case of single steps but I am not entirely sure whether it works with all event types hence the question if `VMI_EVENT_RESPONSE_NONE` is ok as a default response.

rageagainsthepc:
@Wenzel I have implemented the changes as discussed. You can have a look now.

Wenzel:
@rageagainsthepc thanks for this update.
I tested your PR, and there are still some corner cases where kvm_resume_vm goes into a while loop on vmi_destroy.
image

:arrow_up: Here, we have 6 VCPUs, and running the `kvm_events_destroy` routine.
`kvm_resume_vm` goes into a while loop rtying to get the `PAUSE_EVENT` for VCPU 1.
However, VCPU 1 couldn't send his continue reply for the `CR_EVENT`, since the handler was unregistered:
https://github.com/libvmi/libvmi/blob/c2bc417da44ff8bbb94bebf26a50cd321c751943/libvmi/driver/kvm/kvm_events.c#L274

So this check is not enough / should be maybe placed somewhere else.
https://github.com/libvmi/libvmi/blob/c2bc417da44ff8bbb94bebf26a50cd321c751943/libvmi/driver/kvm/kvm_events.c#L231

I think in all the event handlers, you should have a check on `vmi->shutting_down` to skip getting the libvmi_event, calling the user callback, and just sending the reply.

Any way we could do this in a more cleaner fashion ?

rageagainsthepc:
@tklengyel Still waiting for feedback regarding the latest changes.

Wenzel:
I quickly tested this PR on a bare-metal host, and run the `cr3-event-example` with 8 VCPUs.

When I sent a CTRL-C, the code was stuck waiting for VCPU 6 to send its `PAUSE_EVENT`
Capture d’écran de 2023-04-11 03-38-09
Capture d’écran de 2023-04-11 03-38-23

Maybe there is a race-condition somewhere ?
I tested the `master` branch to comapre, and the bug also appears there.

rageagainsthepc:
Hm, stuff like that usually happens when there is an unprocessed single step event somewhere :thinking: In that case we won't receive any PAUSE_VCPU event for that vcpu, no matter how long we wait. Do you have the full log by chance?

Wenzel:
Can you try to repro this on your setup ?
I used the windows xp VM that is shipped by Ansible and configured 1 CPU with 8 cores.

Then i ran the cr3 event example with some basic logging enabled for KVM.

rageagainsthepc:
@Wenzel So, I tried to reproduce this problem on my side and I don't see this bug.
My settings:
Windows 10 21H2 VM with 8 VCPUS
Debian 10 host with newest kvmi-v7 kernel
libvmi with this branch and all debug logging enabled

I've executed the cr3-event example 10 times.

rageagainsthepc:
@Wenzel I can't reproduce the bug even with the latest kvm-v7 kernel. Does it occur every time you execute the cr3 event example?

rageagainsthepc:
@Wenzel @tklengyel This PR has been open for over a year now, is there any chance it will eventually be reviewed and merged? :) It's kind of an important bugfix for kvm and IIRC the issue @Wenzel was facing during testing is unrelated to these changes.

rageagainsthepc:
@tklengyel I've rebased this branch onto master, but now the complexity analysis gives me a worse score for `process_pagefault`. It looks like this function was already too long and now I received a slightly worse score because I've added a few more lines to it. I guess the correct course of action would be to split this function into several smaller ones, but I'd rather keep the changes as simple as possible. What would have me do in this case? Can't we just ignore the complexity analysis in this case? ^^

tklengyel:
Passing the complexity analysis is a nice-to-have, not a required CI check to pass. Unfortunately I have no way to test this PR at the moment. Once @Wenzel gives a thumbs up I'm happy to merge.

rageagainsthepc:
@tklengyel Unfortunately, @Wenzel hasn't reviewed the latest version in 8 months and he also doesn't respond to DMs. To me it feels like the kvm driver doesn't really have a maintainer anymore. So I guess my question is: How long are you planning to wait for his review?

tklengyel:
@rageagainsthepc if there is no maintainer for the KVM driver someone needs to step in for that role. Do you volunteer?

tklengyel:
@rageagainsthepc Let me know when you think it's ready to be merged. I won't be able to test it but don't want to keep it dangling forever either. In case someone finds a bug later we can address it then.

rageagainsthepc:
@tklengyel These changes have been used pretty extensively by me and my team throughout the year and so far we haven't encountered any problems. The only bugs I am currently aware of reside in KVMi itself (for example, the one @Wenzel had faced 8 months ago). I think it's ok to merge.

P.S.: I am sorry you weren't able to find a new maintainer. The community around KVMi doesn't seem to be very active right now. :/

PR #1039, opened 2022-06-02
 Kvm: Implement pause counter

rageagainsthepc:
If we pause the vm and then call `vmi_events_listen` the vm might be resumed unexpectedly if `swap_events` or `clear_events` have to be handled.


rageagainsthepc:
CodeQL check seems to be broken for some reason unrelated to this PR.

tklengyel:
The changes to the Xen driver are unnecessary. Pause is reference counted by Xen and the VM only gets resumed when the count goes to zero.

rageagainsthepc:
Oh damn, I must have missed this. I was sure that it was the same behavior for Xen.

PR #1041, opened 2022-06-22
CI: Checkout bitdefender libkvmi with specific revision

rageagainsthepc:


.github/workflows/compile.yml
```diff
@@ -26,8 +26,9 @@ jobs:
         run: |
           sudo apt-get update -q
           sudo apt-get install -y autoconf-archive flex bison libjson-c-dev libxen-dev libvirt-dev libfuse-dev clang build-essential cmake
-          git clone --depth=2 https://github.com/libvmi/libkvmi
+          git clone https://github.com/bitdefender/libkvmi
```
tklengyel, line 29: I rather not use a 3rd party repository here as it's not guaranteed that it won't get force-pushed. If you need to, you can send a PR to update the libkvmi repo to the latest head from bitdefender.

rageagainsthepc: Sure, but this would have the potential of breaking the libvmi master branch since it's not guaranteed that future libkvmi versions don't introduce breaking changes. So, in order to prove that my feature branch builds I would have to leave libvmi master in a broken state in that case.

tklengyel: Well, ideally we would be using git submodules for this, tagging specific commits, so pushing to the libkvmi repository won't break libvmi master under any circumstance.

rageagainsthepc: Ok, pushing to the libkvmi fork works fine for me. I just thought that this might give you a headache as the maintainer ;) I'm going to close this PR then. 👍

PR #1042, opened 2022-07-28
Fix memory cache remove

rageagainsthepc:
Stale pages are never removed because they are looked up by the location of the physical address on the stack instead of the actual physical address.


tklengyel:
Nice catch, forgot to convert it when I moved things over to use g_direct_hash

PR #1043, opened 2022-08-01
Implement SLAT for kvm

rageagainsthepc:


libvmi/slat.h
```diff
@@ -45,65 +45,66 @@ extern "C" {
...
-* Enables or disables slat for the domain
+* Enables or disables SLAT for the current VM.
 *
 * @param[in] vmi LibVMI instance
-* @param[in] state slat state of the domain
+* @param[in] state SLAT state of the domain
 * @return VMI_SUCCESS or VMI_FAILURE
 */
-status_t vmi_slat_set_domain_state (
+status_t vmi_slat_control (
```
tklengyel, line 65: Is renaming these functions absolutely necessary?

rageagainsthepc: The names of those two functions include Xen specific lingo. Domains aren't really a thing in other hypervisors as far as I am aware. I could also deprecate these functions instead of introducing breaking changes immediately, if you prefer. Or do you think the change is not worth the hassle? Personally I prefer to have hypervisor agnostic function names, but I will not die on this hill, so feel free to request changing it back. ;)

tklengyel: Deprecate the old names first, we can remove them later.

tklengyel:
This looks good to me ,let me know when it's ready for merging.

rageagainsthepc:
@tklengyel This branch is ready for merging now.

PR #1045, opened 2022-09-01
I/O event hooking, firmware type check

blsvntn:
Hello! This PR adds support for I/O event hooking and also new accessor - `vmi_get_firmware_type`, which get firmware type of Xen VM, according to xenstore `/local/domain/<d>/hvmloader/bios` property.

### DIFF
--- a/libvmi/driver/xen/libxc_wrapper.c
+++ b/libvmi/driver/xen/libxc_wrapper.c
@@ -84,6 +84,10 @@ static status_t sanity_check(xen_instance_t *xen)
 
         default:
         /* Things start to be a bit saner from 4.6 */
+        case 16:
+            if ( !w->xc_monitor_io )
+                break;
+            __attribute__ ((fallthrough));
         case 11:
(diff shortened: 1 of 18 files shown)


tklengyel:
Thanks for the PR but without the IO monitor bits being upstream in Xen I'm affraid we can't add this as-is. I would suggest you split the PR into two, first adding the get bios bits which are fine as-is. Then once you get the xc_monitor_io functionality upstream into Xen we can merge that into LibVMI as well.

libvmi/driver/xen/libxc_wrapper.c
```diff
@@ -84,6 +84,10 @@ static status_t sanity_check(xen_instance_t *xen)
 
         default:
         /* Things start to be a bit saner from 4.6 */
+        case 16:
+            if ( !w->xc_monitor_io )
+                break;
+            __attribute__ ((fallthrough));
```
tklengyel, line 90: Use `/* Fall-through */`

PR #1040, opened 2022-06-09
Harmonize KVM singlestep handling

thomasdangl:
Filters out failed singlesteps on KVM to reach the same behavior as on XEN.
Requires https://github.com/bitdefender/libkvmi/pull/14 to be merged first.

PR #1047, opened 2022-10-03
Revert "Harmonize KVM singlestep handling"

tklengyel:
Reverts libvmi/libvmi#1040

This PR is introducing regressions and breaks tools using singlestepping.

### DIFF
--- a/libvmi/driver/kvm/kvm_events.c
+++ b/libvmi/driver/kvm/kvm_events.c
@@ -634,7 +634,7 @@ process_singlestep(vmi_instance_t vmi, struct kvmi_dom_event *kvmi_event)
     event_response_t response = VMI_EVENT_RESPONSE_NONE;
     vmi_event_t *libvmi_event = NULL;
 
-    if (!vmi->shutting_down && !kvmi_event->event.ss.failed) {
+    if (!vmi->shutting_down) {
         // lookup vmi_event
         libvmi_event = g_hash_table_lookup(vmi->ss_events, GUINT_TO_POINTER(kvmi_event->event.common.vcpu));
 #ifdef ENABLE_SAFETY_CHECKS


rageagainsthepc:
@tklengyel Care to elaborate? `!kvmi_event->event.ss.failed` is required because sometimes singlesteps can fail, for example when the hypervisor injects an unrelated interrupt into the guest before the next instruction is executed (IIRC). If there are regressions the root cause is most likely located elsewhere. Reverting this change only means that you are now notified for completed singlestep events even if the RIP has not advanced to the next instruction.

tklengyel:
The regression manifests by a hang of the VM and qemu spinning at 100% CPU utilization. With this PR reverted everything works as expected and the VM is executing normally with no hangs.

rageagainsthepc:
Have you checked if you are using the lastest version of libkvmi? Specifically, a version where [this](https://github.com/bitdefender/libkvmi/commit/aac5787ffa30e9884a26d44dfab3f34cd1570235) commit is present?

tklengyel:
Yes, it was with the latest libkvmi (LibVMI wouldn't even compile otherwise).

rageagainsthepc:
True, but I don't know your set up. It could have been possible that you link against one version at compile time, but load a different version at runtime.
Honestly, this is very weird because it never came up when we were testing this change. And we are still on a version of libvmi where this change is a part of, but we have never experienced this behavior. Is there anything out of the ordinary with your set up?
Come to think of it, we did fix a bug that had similar behavior, but it was unrelated to this change. Maybe you just had bad luck? Are you on the latest kvm-v7 branch? https://github.com/KVM-VMI/kvm/commit/f76c34e529c34b9cf259073f81c3b2a440e1c1d5

tklengyel:
It was a fresh build at https://github.com/KVM-VMI/kvm/commit/cc45b7af077c4514020fd403b58b674b3b1c589e, with a couple extra unrelated fixes to make it compile on Ubuntu 22.04, https://github.com/tklengyel/linux/tree/kvm-vmi-v7-ubuntu-22.04-build. Nothing extraordinary in the code itself, using singlestepping with the event response route: https://github.com/intel/kernel-fuzzer-for-xen-project/blob/master/src/dmamonitor.c#L275

rageagainsthepc:
Hm, the return value of `vmi_set_mem_event()` is not checked. Could it be that the permissions of the page are not lifted correctly and the singlestep therefore fails to execute? That would explain the infinite loop. :thinking:

tklengyel:
If that was the case reverting this change would have no effect, but it fixes the issue.

rageagainsthepc:
My assumption was that the issue is only "fixed" because we don't check the failed state of the singlestep event anymore with the reverted state. But I have rewritten the mem-event-example to resemble your use case more closely and it turns out that this assumption wasn't true. Apparently mem events take precedence over singlesteps which makes sense. So, failing to lift permissions only results in the mem event being re-reported in a loop. However, I was not able to reproduce the bug you are experiencing. Both versions of libvmi work fine for me. :(

rageagainsthepc:
In case you are interested: https://github.com/libvmi/libvmi/compare/master...GDATASoftwareAG:libvmi:mem-event-test
Also, I think I should create a PR for these changes. :thinking:  The old mem-event-example had a race condition between clearing the mem event and processing the singlestep.

rageagainsthepc:
@tklengyel Could you check if my modified mem-event example is able to reproduce the bug on your end? It would be great to have some small sample code to be able to work with.

tklengyel:
I did try with the modified mem-event example and it seems to be working fine, except on CTRL+C it wouldn't exit and just hang there. Had to `kill -9` it to exit.

I tried once more with this change in LibVMI still there and the latest libkvmi and I can still see the same issue. EPT events stop coming in the dmamonitor tool and there was no error setting the EPT permission to clear. QEMU is still spinning at 100%. One thing I noticed however, is that on the VM console if I hit enter the boot process continues for a second to then freeze again. Seems that interrupt coming from the console kicked things into motion temporarily. However, after a while even that trick stops working and the VM just appears totally frozen with QEMU at 100%.

rageagainsthepc:
> I did try with the modified mem-event example and it seems to be working fine, except on CTRL+C it wouldn't exit and just hang there. Had to kill -9 it to exit.

This issue will most likely be fixed by https://github.com/libvmi/libvmi/pull/1038.
Ok, so the mem-event example is not sufficient in order to reproduce the bug. Hm... :thinking:

tklengyel:
Yea, I'm perplexed about the bug too. The EPT permission is cleared before enabling singlestep and then placed back in the singlestep callback. By having the callback for the failed singlestep we get a duplicate EPT event but then things work afterwards. We would be one step closer knowing what QEMU is spinning when this happens. I'll see if I can figure that out.

tklengyel:
Well, it seems that I'm getting the same hang with this PR reverted now, so it may have been just luck that with this reverted things appeared to be working. There seems to be some type of race-condition that triggers it.

rageagainsthepc:
Do you know whether the hang is entirely in the kernel? Or do we have a loop in libvmi?

tklengyel:
I think your guess that this is an EPT fault is the most reasonable. There are no callbacks issues by libkvmi at all but QEMU is shown as spinning at 100% by top. That usually happens when the vCPU doesn't exit out and is stuck in KVM somewhere. This would happen when an EPT fault is being hit over and over. I'm still trying to enable some debug messages to verify.

tklengyel:
OK, I'm getting the VM stuck in looping on vmexit 12, ie halt, which hits this path in KVM https://github.com/KVM-VMI/kvm/blob/kvmi-v7/arch/x86/kvm/x86.c#L7666

This comment seems relevant:

```
        /*
	 * TODO: we might be squashing a GUESTDBG_SINGLESTEP-triggered
	 * KVM_EXIT_DEBUG here.
	 */`
```

However, even in the v12 branch I don't see anything special required to handle this when MTF is enabled.

thomasdangl:
Just FYI, _KVM_GUESTDBG_SINGLESTEP_ relates to the trap flag in the FLAGS register, not the monitor trap flag in the VMCS. I think it is only used by the GDB stub when single-stepping in the debugger.

Looping on _EXIT_REASON_HLT_ doesn't sound good. Assuming _KVM_GUESTDBG_SINGLESTEP_ is disabled, there are only two things happening in that code:

1. Skipping over the current instruction (https://elixir.bootlin.com/linux/v5.4.217/source/arch/x86/kvm/vmx/vmx.c#L1632)
2. Injection of a debug trap if the trap flag was set by the guest (https://elixir.bootlin.com/linux/v5.4.217/source/arch/x86/kvm/x86.c#L6643)

So I guess we need to figure out why the guest does not advance past the HLT instruction. Unfortunately, I have not managed to reproduce the issue myself yet. Maybe you could disassemble the guest's kernel at the instruction pointer set by _skip_emulated_instruction_ to see what's going on there.

rageagainsthepc:
@tklengyel Since this seems to be an issue with kvmi and not this commit here, could you add it back again?

PR #1048, opened 2022-10-07
Rework mem-event-example

rageagainsthepc:
The original example clears page permissions by clearing the memory event. Since events are not cleared within callbacks, there is a brief period where we may receive a singlestep event before the mem event is cleared (and the memory permissions lifted). This results in unexpected behavior.

### DIFF
--- a/examples/mem-event-example.c
+++ b/examples/mem-event-example.c
@@ -22,11 +22,10 @@
 
 #include <stdlib.h>
 #include <string.h>
-#include <errno.h>
-#include <sys/mman.h>
 #include <stdio.h>
 #include <inttypes.h>
 #include <signal.h>
+#include <glib.h>
 
 #include <libvmi/libvmi.h>
 #include <libvmi/events.h>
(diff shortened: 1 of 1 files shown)


examples/mem-event-example.c
```diff
@@ -39,10 +37,26 @@ static void close_handler(int sig)
...
 }
 
-event_response_t mem_cb(vmi_instance_t vmi, vmi_event_t *event)
+event_response_t singlestep_cb(vmi_instance_t vmi, vmi_event_t *event)
 {
-    (void)vmi;
+    addr_t paddr = 0;
+    // We need to know the physical address of the event that triggered this callback to be able to
+    // reinstate the original memory permissions
+    if (vmi_pagetable_lookup(vmi, event->x86_regs->cr3, event->ss_event.gla, &paddr) == VMI_FAILURE) {
```
tklengyel, line 45: IMHO it makes a lot more sense to just stuff this information in `event->data` using GSIZE_TO_POINTER(event->mem_event.gfn) when the permissions are getting cleared. Saves you a pagetable walk.

rageagainsthepc: I guess it makes sense to get rid of the pagetable walk. I was trying to avoid having to put the events into global scope.
`GSIZE_TO_POINTER` expands to a simple type cast to `gpointer`, right? What's the benefit over using `(void *)`? Doesn't it only make sense if we are working with glibc objects?

tklengyel: It's more portable, see https://developer.gimp.org/api/2.0/glib/glib-Type-Conversion-Macros.html#GSIZE-TO-POINTER:CAPS

PR #1050, opened 2022-11-29
Return success if whole region is unmapped

rageagainsthepc:
I would like to propose to return `VMI_SUCCESS` if there are no pages to map. IMHO it shouldn't be viewed as an error, there is simply no work to be done. Otherwise it's going to be hard to distinguish between cases where `driver_mmap_guest()` returns an error or the whole region is simply not mapped.

### DIFF
--- a/libvmi/read.c
+++ b/libvmi/read.c
@@ -118,15 +118,15 @@ vmi_mmap_guest(
         buf_offset += vmi->page_size;
     }
 
-    if (pfn_ndx == 0) {
-        goto done;
-    }
+    void *base_ptr = NULL;
+    // do mmap only if there are pages available for mapping
+    if (pfn_ndx != 0) {
+        base_ptr = (char *) driver_mmap_guest(vmi, pfns, pfn_ndx);
 
-    void *base_ptr = (char *)driver_mmap_guest(vmi, pfns, pfn_ndx);
-
-    if (MAP_FAILED == base_ptr || NULL == base_ptr) {
-        dbprint(VMI_DEBUG_READ, "--failed to mmap guest memory");
-        goto done;
+        if (MAP_FAILED == base_ptr || NULL == base_ptr) {
+            dbprint(VMI_DEBUG_READ, "--failed to mmap guest memory");
+            goto done;
+        }
     }
 
     for (i = 0; i < num_pages; i++) {


rageagainsthepc:
The build fails due to unrelated compiler warnings. :(

tklengyel:
I'm fine with this PR as-is. For the CI failure, looks like debian bookworm's gcc now includes `-Wgnu-variable-sized-type-not-at-end` in `-Werror`. Should be fine to just silence that warning with `-Wno-gnu-variable-sized-type-not-at-end`.

tklengyel:
Actually, looks like the code in the KVM path is setting up structures in a way that is not safe. Those structures will need to be reshuffled so the variable length part of the structure is always at the end.

rageagainsthepc:
> Actually, looks like the code in the KVM path is setting up structures in a way that is not safe. Those structures will need to be reshuffled so the variable length part of the structure is always at the end.

I had a look at the code and it is though (in a way):
```C
    struct {
        struct kvm_msrs msrs;
        struct kvm_msr_entry entries[7];
    } msrs = { 0 };
```
The second member `entries` is used to reserve space on the stack for the zero length `kvm_msr_entry` array which is the last member of `kvm_msrs`. The code looks correct to me and I don't have any ideas on how to do it differently apart from allocating the `kvm_msrs` struct on the heap. I would simply ignore the warning for this file, or do you have an alternative proposal?

rageagainsthepc:
Fyi, I have made the mistake of jumping over the code that replaces `-1` with `NULL` in the result array. Should be fixed now.

tklengyel:
The error is `libvmi/driver/kvm/kvm.c:890:25: error: field 'msrs' with variable sized type 'struct kvm_msrs' not at the end of a struct or class is a GNU extension [-Werror,-Wgnu-variable-sized-type-not-at-end]`. The order within that struct needs to be the flipped  because `kvm_msrs` has variable length: https://elixir.bootlin.com/linux/latest/source/arch/x86/include/uapi/asm/kvm.h#L196.

tklengyel:
Looking at the libvmi code I'm not even sure the `entries` part of that struct is used for anything:
```
struct {
        struct kvm_msrs msrs;
        struct kvm_msr_entry entries[7];
    } msrs = { 0 };
```

/cc @Wenzel

rageagainsthepc:
> The error is `libvmi/driver/kvm/kvm.c:890:25: error: field 'msrs' with variable sized type 'struct kvm_msrs' not at the end of a struct or class is a GNU extension [-Werror,-Wgnu-variable-sized-type-not-at-end]`. The order within that struct needs to be flipped because `kvm_msrs` has variable length: https://elixir.bootlin.com/linux/latest/source/arch/x86/include/uapi/asm/kvm.h#L196.

I think you may have misunderstood me. The sole purpose of `struct kvm_msr_entry entries[7]` is to reserve space on the stack for the variable length struct `struct kvm_msrs msrs`. Flipping the order would break the code.
Those lines are equivalent in this context:
```C
msrs.entries[0].index = msr_index[MSR_IA32_SYSENTER_CS];
msrs.msrs.entries[0].index = msr_index[MSR_IA32_SYSENTER_CS];
```

tklengyel:
Yea, the issue is with trying to be clever about reserving that space on the stack with that syntax. See #1051 for a fix.

PR #1054, opened 2022-12-20
Fix x64 windows address translation

rageagainsthepc:
On newer versions of Windows (at least Win10 1909), some higher bits of the pfn value are used as flags for transition pages. This results in libvmi returning incorrect physical addresses for transition pages. Tested with `vmi_mmap_guest()` which used to fail a lot prior to this fix.

### DIFF
--- a/libvmi/arch/amd64.c
+++ b/libvmi/arch/amd64.c
@@ -25,12 +25,10 @@
  */
 
 #include <glib.h>
-#include <stdlib.h>
 #include <sys/mman.h>
 
 #include "private.h"
 #include "x86.h"
-#include "driver/driver_wrapper.h"
 #include "arch/amd64.h"
 
 /* PML4 Table  */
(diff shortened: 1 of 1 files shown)


rageagainsthepc:
I did not apply these changes to all the lines where we read entries pointing to the next page table layer because I assumed that page tables can't be "in transition" conceptually, right?

tklengyel:
While I'm not sure I do think pagetables can be in transition, Windows is known to move pagetables around and at point I would not be surprised to see pages holding pagetable entries to be marked in transition.

libvmi/arch/amd64.c
```diff
@@ -137,20 +135,35 @@ status_t get_pte_ia32e (vmi_instance_t vmi,
 }
 
 static inline
-uint64_t get_paddr_ia32e (addr_t vaddr, uint64_t pte)
+uint64_t get_paddr_ia32e (vmi_instance_t vmi, addr_t vaddr, uint64_t pte)
```
tklengyel, line 138: I would prefer if the vmi instance was not passed to these functions and instead a `max_bits` uint was passed in. It would be 51 by default but could be specified to be 44 where appropriate. Right now you specify 44 for all Windows where it sounds like that's only correct for some newer versions. Should that be more fine grained so it only applies where needed?

tklengyel (changes requested):

libvmi/arch/amd64.c
tklengyel: I would still like to avoid having to pass the `vmi` instance to these functions, so please adjust the prototype of these functions to take the number as input instead of doing a pointer deref and and if/else.

tklengyel:
> > While I'm not sure I do think pagetables can be in transition, Windows is known to move pagetables around and at point I would not be surprised to see pages holding pagetable entries to be marked in transition.
> 
> Ok, maybe I have misunderstood what transition pages are about because I was under the impression that it can only apply to file-backed memory. At least that's what I understood while reading the paper "Using Every Part of the Buffalo in Windows Memory Analysis" that is mentioned somewhere in this code base. But they don't really go into detail, so I might be wrong.

We can certainly leave it as-is and if it becomes an issue adjust later.

libvmi/arch/amd64.c
rageagainsthepc: I already have, please see my latest force push.

libvmi/arch/amd64.c
```diff
@@ -283,7 +296,7 @@ GSList* get_pages_ia32e(vmi_instance_t vmi, addr_t npt, page_mode_t npm, addr_t
             goto done;
 
         uint64_t pdpte_index;
-        for (pdpte_index = 0; pdpte_index < IA32E_ENTRIES_PER_PAGE; pdpte_index++, pdpte_location++) {
+        for (pdpte_index = 0; pdpte_index < IA32E_ENTRIES_PER_PAGE; pdpte_index++, pdpte_location += entry_size) {
```
tklengyel, line 299: This seems to be an unrelated change, different bug?

rageagainsthepc: Yes, that's why I've created a separate commit for this change. I've discovered it while looking for the other bugs. Would you like me to create a separate PR for it?

tklengyel: Not necessary, just checking

tklengyel (approved):

Issue #1052, opened 2022-12-12
`vmi_get_va_pages()` takes several minutes to complete on Windows


rageagainsthepc:
Hi!

I've noticed that `vmi_get_va_pages()` takes several minutes on Windows (Win10_21H2 specifically) in contrast to Linux where it is almost instant. If you count the entries in the result list it is close to one billion. If you count how many times a single physical page is mapped by the process it will yield a result similar to this output:
```
[    4] System (struct addr:ffffe1013b8a4040)
page addr 82837504 count 809529504
page addr 1052672 count 1572838
page addr 83886080 count 140567
page addr 84934656 count 8192
page addr 82845696 count 3070
page addr 85983232 count 270
page addr 983040 count 191
page addr 4227899392 count 129
page addr 4043309056 count 4
page addr 82849792 count 4
page addr 745472 count 3
page addr 671744 count 3
page addr 749568 count 3
page addr 753664 count 3
page addr 757760 count 3
page addr 761856 count 3
page addr 765952 count 3
page addr 770048 count 3
page addr 774144 count 3
page addr 778240 count 3
page addr 782336 count 3
page addr 663552 count 3
page addr 659456 count 3
page addr 667648 count 3
page addr 655360 count 3
page addr 737280 count 3
page addr 675840 count 3
page addr 679936 count 3
page addr 684032 count 3
page addr 741376 count 3
page addr 688128 count 3
page addr 692224 count 3
[...]
```
Note that while I've only provided the output for the system process other processes show similar behavior.
Peeking into the memory content of those pages that are on the top of this list reveals that they seem to consist of either null bytes or periodic patterns.

So, my question is: Does anybody know what feature of Windows causes this behavior? What would be the best way to reduce the performance cost?

tklengyel:
That doesn't sound right. Can you double-check what Volatility thinks are the available pages using https://github.com/volatilityfoundation/volatility/wiki/Address-Spaces#get_available_pages?

tklengyel:
That's your clue then. LibVMI's get_va_pages was originally modeled on Volatility's approach, they might have changed some things which makes their parser more accurate. Diff the two, find where things get off the rails in LibVMI and patch. ;)

rageagainsthepc:
Will do, as time permits.

rageagainsthepc:
[...]

PR #1055, opened 2022-12-20
Skip consecutive PD/PT entries for Windows 10 in vmi_get_va_pages

thomasdangl:
Drops duplicated entries in page tables at the lower two levels for x86_64.
Patch is inspired by https://github.com/volatilityfoundation/volatility/commit/6b9349b9f84b32ea7c19b8ab372e7a93efef926a.
Solves https://github.com/libvmi/libvmi/issues/1052.

### DIFF
--- a/libvmi/arch/amd64.c
+++ b/libvmi/arch/amd64.c
@@ -241,6 +241,12 @@ done:
     return status;
 }
 
+static inline
+bool is_duplicate(vmi_instance_t vmi, uint64_t *page, uint64_t index)
+{
+    return vmi->os_type == VMI_OS_WINDOWS && index > 0 && page[index] == page[index - 1];
+}
+
 GSList* get_pages_ia32e(vmi_instance_t vmi, addr_t npt, page_mode_t npm, addr_t dtb)
 {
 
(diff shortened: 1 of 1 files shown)


libvmi/arch/amd64.c
```diff
@@ -241,6 +241,12 @@ status_t v2p_ia32e (vmi_instance_t vmi,
     return status;
 }
 
+static inline
+bool is_duplicate(vmi_instance_t vmi, uint64_t *page, uint64_t index)
+{
+    return vmi->os_type == VMI_OS_WINDOWS && index > 0 && page[index] == page[index - 1];
```
Henning1, line 247: Is it safe to access `page[index - 1]` for index 0?

thomasdangl: `page[index - 1]` will not be accessed for index 0 due to short-circuit evaluation of the `&&` operator and the previous check against `index > 0`. See [ISO/IEC 9899 standard, section 6.5.13](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) for this:

> Unlike the bitwise binary & operator, the && operator guarantees left-to-right evaluation;
there is a sequence point after the evaluation of the first operand. If the first operand
compares equal to 0, the second operand is not evaluated.

PR #1056, opened 2023-01-09
Address to kernel symbol translation for windows

Henning1:
This adds translation of virtual adresses to kernel symbols based on volatility json profiles. A hash table is placed to the side of the json objects to enable lookups in the opposite direction.

The method `windows_address_to_kernel_symbol` (and its linux counterpart) expect an address parameter _and_ an access_context. This is redundant because the access_context already has an address. If it's preferred to only use the addr field from the context, I can adapt for that.

### DIFF
--- a/libvmi/json_profiles/json_profiles.c
+++ b/libvmi/json_profiles/json_profiles.c
@@ -76,7 +76,7 @@ bool json_profile_init(vmi_instance_t vmi, const char* path)
             return false;
     };
 
-    if (vmi->init_flags & VMI_INIT_SYMBOLLOOKUP) {
+    if (vmi->init_flags & VMI_INIT_V2SYM) {
         if (vmi->json.build_reverse_symbol_table == NULL) {
             dbprint(VMI_DEBUG_MISC, "Address to symbol translation not available for this json profile\n");
             return false;
--- a/libvmi/json_profiles/json_profiles.c
+++ b/libvmi/json_profiles/json_profiles.c
@@ -64,15 +64,28 @@ bool json_profile_init(vmi_instance_t vmi, const char* path)
             json->bitfield_offset_and_size = volatility_profile_bitfield_offset_and_size;
             json->get_os_type = volatility_get_os_type;
             json->struct_field_type_name = volatility_struct_field_type_name;
+            json->build_reverse_symbol_table = volatility_build_reverse_symbol_table;
             break;
         case JPT_REKALL_PROFILE:
             json->handler = rekall_profile_symbol_to_rva;
             json->bitfield_offset_and_size = rekall_profile_bitfield_offset_and_size;
             json->get_os_type = rekall_get_os_type;
+            json->build_reverse_symbol_table = NULL;
             break;
         default:
             return false;
     };
+
+    if (vmi->init_flags & VMI_INIT_SYMBOLLOOKUP) {
+        if (vmi->json.build_reverse_symbol_table == NULL) {
+            dbprint(VMI_DEBUG_MISC, "Address to symbol translation not available for this json profile\n");
+            return false;
+        }
+        if (vmi->json.build_reverse_symbol_table(vmi->json.root, &vmi->json.reverse_symbol_table) != VMI_SUCCESS) {
+            dbprint(VMI_DEBUG_MISC, "Reverse symbol table build failed\n");
+            return false;
+        }
+    }
 
     return true;
 }
@@ -85,6 +98,9 @@ void json_profile_destroy(vmi_instance_t vmi)
 
     vmi->json.path = NULL;
     vmi->json.root = NULL;
+
+    if (vmi->json.reverse_symbol_table != NULL)
+        g_hash_table_destroy(vmi->json.reverse_symbol_table);
 }
 
 json_object* vmi_get_kernel_json(vmi_instance_t vmi)
--- a/libvmi/json_profiles/volatility_ist.c
+++ b/libvmi/json_profiles/volatility_ist.c
@@ -213,6 +213,41 @@ exit:
     return ret;
 }
 
+status_t
+volatility_build_reverse_symbol_table(json_object *json, GHashTable** table)
+{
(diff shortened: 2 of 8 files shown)


libvmi/json_profiles/volatility_ist.c
```diff
@@ -213,6 +213,39 @@ volatility_ist_symbol_to_rva(
...
+status_t
+volatility_build_reverse_symbol_table(json_object *json, GHashTable** table) {
+    status_t ret = VMI_FAILURE;
+    json_object *symbols = NULL;
+    if (!json_object_object_get_ex(json, "symbols", &symbols)) {
+        dbprint(VMI_DEBUG_MISC, "Volatility IST profile: no symbols section found\n");
+        goto exit;
+    }
+
+    *table = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
```
tklengyel, line 225: You could just use `g_hash_table_new`

rageagainsthepc:
>The method windows_address_to_kernel_symbol (and its linux counterpart) expect an address parameter and an access_context. This is redundant because the access_context already has an address. If it's preferred to only use the addr field from the context, I can adapt for that.

I would argue that the access context makes no sense here since we can't influence how the address will be translated. Both the Windows and the Linux implementation use the internally stored base address of the kernel. So, I would drop the `ctx` rather then the `addr_t` value.

libvmi/json_profiles/volatility_ist.c
Henning1: Switched to `g_hash_table_new`

libvmi/os/windows/memory.c
```diff
@@ -89,6 +89,25 @@
     return ret;
 }
 
+char*
+windows_address_to_kernel_symbol(
+    vmi_instance_t vmi,
+    addr_t address,
+    const access_context_t *ctx)
+{
+    status_t ret = VMI_FAILURE;
```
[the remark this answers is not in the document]
Henning1: Removed unused variable

libvmi/json_profiles/json_profiles.c
```diff
@@ -100,6 +105,27 @@ status_t vmi_get_symbol_addr_from_json(vmi_instance_t vmi, json_object* json, co
     return vmi->json.handler(json, symbol, NULL, addr, NULL);
 }
 
+status_t vmi_address_to_symbol_from_json(vmi_instance_t vmi, addr_t addr, const char** symbol)
```
tklengyel, line 108: I'm not a 100% sure we would want a different API entry point for this, why not just use `vmi_translate_v2ksym`?

libvmi/json_profiles/json_profiles.c
```diff
@@ -100,6 +105,27 @@ status_t vmi_get_symbol_addr_from_json(vmi_instance_t vmi, json_object* json, co
...
+{
+    if(vmi->json.build_reverse_symbol_table == NULL)
+    {
+        dbprint(VMI_DEBUG_MISC, "Address to symbol not implemented for this type of json profile\n");
+        return VMI_FAILURE;
+    }
+
+    /* delay table build until feature is used */
```
tklengyel, line 116: Instead of this being a delayed step this should be performed by adding a new `VMI_INIT` flag and be built during init time.

libvmi/json_profiles/json_profiles.c
Henning1: I removed the method and added the ht lookup to the implementation used by `vmi_translate_v2ksym`

libvmi/json_profiles/json_profiles.c
Henning1: Removed delayed ht build and added `VMI_INIT_SYMBOLLOOKUP `

Henning1: I did not find a name for  `VMI_INIT_...` that's more expressive but still concise. Alternatively, what do you think about performing the ht build by default? The overhead is not too high.

These are execution times for a win10 volatility profile:
```
build_reverse_symbol_table() took 20.645000 ms to execute 
vmi_init_profile() took 215.098000 ms to execute 
```
In the measurement `build_reverse_symbol_table() ` was executed a part of` vmi_init_profile()`

tklengyel: I wouldn't want it on by default so either an initflag or a compile option would be required. I think the init flag is better. Let's go with `VMI_INIT_V2SYM`, that's more consistent with the API naming.

Henning1: Changed to `VMI_INIT_V2SYM`

libvmi/json_profiles/volatility_ist.c
```diff
@@ -213,6 +213,41 @@ volatility_ist_symbol_to_rva(
...
+            dbprint(VMI_DEBUG_MISC, "Volatility IST: no address found for %s\n", symbol_name);
+            continue;
+        }
+        addr_t rva;
+#ifdef JSONC_UINT64_SUPPORT
+        rva = json_object_get_uint64(address);
+#else
+        rva = json_object_get_int64(address);
+#endif
+        g_hash_table_insert(*table, GINT_TO_POINTER(rva), (gpointer)symbol_name);
```
tklengyel, line 243: Please use `GSIZE_TO_POINTER` for values that come from `addr_t`. In this case we are talking about RVAs so it's unlikely INT wrapping would be an issue but still.

Henning1: Changed to `GSIZE_TO_POINTER `

Henning1:
There was an empty method stub missing for compilation without volatility. It is a little bit difficult to handle all cases without direct CI feedback. I hope all cases are handled now.

Henning1:
> There was an empty method stub missing for compilation without volatility. It is a little bit tricky to handle all cases without direct CI feedback. I hope everything works now.

There was another one. Now works with and without `ENABLE_JSON_PROFILES`

PR #1058, opened 2023-01-16
Add option to disable v2p cache

skvl:
This option caused by bug in drakvuf with memory access on page boundary. Which looks like this
```
regmon Time=1669101675.260478,PID=2252,PPID=2704,TID=2960,ProcessName="\Device\HarddiskVolume2\Program Files\Google\Chrome\Application\chrome.exe",Method=NtCreateKey,Key="⬓碼炐耀梺荱炏"
```

It had been found that calling `vmi_v2pcache_flush` every time fixes the error. By default dravkfu flushes the cache in `vmi.c :: flush_vmi` every `VMI_FLUSH_RATE` (100) events.

Instead of fully disabling the v2p cache or flushing all caches on every event it been decided to add the compile time option.


tklengyel:
I don't think a separate option is needed for this. The main use of the cache is the v2p vTLB, the other caches are rarely used, so if you find the v2p cache to be in the way you should be fine with the other caches disabled along with it.

Also, perhaps the right solution is implementing a better detection technique for DRAKVUF when the v2p cache needs flushing. It has been on the TODO list since the beginning.

PR #1059, opened 2023-01-16
Add option to build with O3 level

skvl:


CMakeLists.txt
```diff
@@ -39,12 +39,13 @@ include(GNUInstallDirs)
...
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -g3 -O0 -Wall -Wextra -Wno-missing-field-initializers -Wno-missing-braces -Wno-unused-result -Wno-ignored-attributes")
 set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -Werror -O2")
+string(REGEX REPLACE "O2 " "O3 " CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -g -ggdb3" )
 
 # default hardening flags that have no performance hit
 set(HARDENING_LINKER_FLAGS "-Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now")
 # update default flags
-set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${HARDENING_LINKER_FLAGS}")
-set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${HARDENING_LINKER_FLAGS}")
+set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${HARDENING_LINKER_FLAGS} -flto")
```
rageagainsthepc, line 47: CMake supports the target property `INTERPROCEDURAL_OPTIMIZATION` for enabling LTO. https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html
There is also a module for checking whether LTO is supported. https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html

CMakeLists.txt
```diff
@@ -39,12 +39,13 @@ include(GNUInstallDirs)
 #-----------------------------------------------------------------------------
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -g3 -O0 -Wall -Wextra -Wno-missing-field-initializers -Wno-missing-braces -Wno-unused-result -Wno-ignored-attributes")
 set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -Werror -O2")
+string(REGEX REPLACE "O2 " "O3 " CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -g -ggdb3" )
```
rageagainsthepc, line 42: `-g` is already present in `CMAKE_C_FLAGS_RELWITHDEBINFO`. Also, why not simply use `-g3` instead of the gdb specific `-ggdb3`?

CMakeLists.txt
skvl: Great! Thanks!

CMakeLists.txt
skvl: Thanks!

configure.ac
```diff
@@ -262,6 +261,32 @@ LD="$saved_LD"
...
+
+saved_CC="$CC"
+saved_LD="$LD"
+flag_wrap="$srcdir/wrap-compiler-for-flag-check"
+CC="$flag_wrap $CC"
+LD="$flag_wrap $LD"
+
+OPTIMIZATION_CFLAGS=""
+
+AS_IF([test x"$hardening" != x"yes"], [
```
tklengyel, line 275: Why is this incompatibility with hardening?

configure.ac
```diff
@@ -262,6 +261,32 @@ LD="$saved_LD"
...
+  check_cc_flag([-O3], [OPTIMIZATION_CFLAGS="$OPTIMIZATION_CFLAGS -O3"])
+  check_cc_flag([-g -ggdb3], [OPTIMIZATION_CFLAGS="$OPTIMIZATION_CFLAGS -g -ggdb3"])
```
tklengyel, line 277: You generally want debug flags added only when building debug mode.

configure.ac
```diff
@@ -182,9 +182,9 @@ AM_CONDITIONAL([DEBUG], [test x$debug = xyes])
 
 AC_ARG_ENABLE([hardening],
   [AS_HELP_STRING([--enable-hardening],
-    [Enable compiler and linker options to frustrate memory corruption exploits @<:@yes@:>@])],
+    [Enable compiler and linker options to frustrate memory corruption exploits @<:@no@:>@])],
   [hardening="$enableval"],
-  [hardening="yes"])
+  [hardening="no"])
```
tklengyel, line 187: Blank disabling default option like this is a no-go.

configure.ac
skvl: Because with hardening enabled `-O2` is specified

CMakeLists.txt
skvl: I failed to use this option. I've tried something like this:
```
set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
```

But in build log there was no `-flto` options.

CMakeLists.txt
skvl: Switched to `-g3` option here

configure.ac
skvl: Fixed

skvl:
@tklengyel , the build error is not my fault:
изображение

configure.ac
tklengyel: Perhaps it would make sense to remove 02 from the harden flags.

CMakeLists.txt
rageagainsthepc: @skvl I have never tried the global approach myself, but I was able to make it work as a target property.  I think this is the cleaner approach anyways.

PR #1061, opened 2023-03-22
Add extraction for w string

K-Mayer:
We're trying to extract PWCHARs from Windows (for example the PathToFile  from [LdrLoadDll ](http://undocumented.ntinternals.net/UserMode/Undocumented%20Functions/Executable%20Images/LdrLoadDll.html)). Since those are UTF-16 strings with [16 bit wide chars ](https://learn.microsoft.com/en-us/windows/win32/extensible-storage-engine/wchar) we can't use the vmi_read_str or the os_read_unicode_struct functions.

### DIFF
--- a/libvmi/libvmi.h
+++ b/libvmi/libvmi.h
@@ -1293,6 +1293,19 @@ char *vmi_read_str(
     const access_context_t *ctx) NOEXCEPT;
 
 /**
+ * Reads a null terminated UTF-16 string from memory, starting at
+ * the given virtual address. The string will be converted to UTF-8.
+ * The returned value must be freed by the caller.
+ *
+ * @param[in] vmi LibVMI instance
+ * @param[in] ctx Access context
+ * @return String read from memory or NULL on error
+ */
+char *vmi_read_w_str(
+    vmi_instance_t vmi,
+    const access_context_t *ctx) NOEXCEPT;
+
+/**
  * Reads a Unicode string from the given address. If the guest is running
  * Windows, a UNICODE_STRING struct is read. Linux is not yet
  * supported. The returned value must be freed by the caller.
--- a/libvmi/read.c
+++ b/libvmi/read.c
@@ -24,9 +24,9 @@
  * along with LibVMI.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include <string.h>
-#include <wchar.h>
 #include <errno.h>
+#include <iconv.h>
+#include <string.h>
 
 #include "private.h"
 #include "driver/driver_wrapper.h"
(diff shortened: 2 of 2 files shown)


tklengyel:
> we can't use the vmi_read_str or the os_read_unicode_struct functions.

Why can't you use `vmi_read_unicode_str` exactly? I don't like having to introduce a separate API for this when by looks of it read_unicode_str should be already capable of returning UTF-16 just fine.

K-Mayer:
> > we can't use the vmi_read_str or the os_read_unicode_struct functions.
> 
> Why can't you use `vmi_read_unicode_str` exactly? I don't like having to introduce a separate API for this when by looks of it read_unicode_str should be already capable of returning UTF-16 just fine.

The problem is that libvmi tries to read a unicode struct with length, maxlength, buffer (add padding) which in this case simply does not exist. A PWCHAR is a pointer to a 0-terminated UTF-16 string. We have to find the length ourselves by finding the 0-terminator.

tklengyel:
How LibVMI handles the string internally is not what I'm asking about. Internally LibVMI can use heuristics to figure out what string format is being used and act accordingly. Why do you need a new external API for returning a UTF-16 Unicode string when the existing API is already sufficient to do that? The official Microsoft documentation considers wchar to be 16-bit Unicode https://learn.microsoft.com/en-us/windows/win32/extensible-storage-engine/wchar.

rageagainsthepc:
Let me weigh in here. This is not at all about internal libvmi string handling. In the windows API there are strings that are respresented by a unicode struct which stores a pointer to the data as well as the length. In order to extract the string you first have to read this struct and then the actual content pointed to by the contents of said struct. This is what the `read_unicode_str()` function already does. But there are also raw UTF-16 strings which have to be extracted differently since they come without a unicode struct. I suppose there was some confusion regarding the internal libvmi unicode struct and the unicode struct used by Windows. If you want to use a heuristic to figure out whether we need to extract a unicode struct or raw string data, I'd be all ears. I don't see a good solution to employ a heuristic here at first glance tbh.

rageagainsthepc:
> Are you telling me you can't fit the 16-bit unicode string into this structure and you must have a separate API to handle it?

Sure, we can most certainly fit the UTF-16 string into this vmi struct. I guess it makes sense to let the user decide whether the string should be converted to something else instead of forcing a conversion to UTF-8 (also, better code reuse). In any case I'm all for integrating this suggestion. But again, that's not the issue here:

> The LibVMI user says read me a unicode string from a given address. Well, does that address have a unicode structure with a pointer at the expected field, or not?

How are you going to reliably predict whether the address pointed to contains a unicode struct or string data? It's not like there is a struct header one would be able to parse. Maybe I'm missing something here, but I don't see how your idea should be realized.

tklengyel:
> How are you going to reliably predict whether the address pointed to contains a unicode struct or string data?

```
typedef struct _UNICODE_STRING {
  USHORT Length;
  USHORT MaximumLength;
  PWSTR  Buffer;
} UNICODE_STRING, *PUNICODE_STRING;
```

1. Check whether pwstr resolves when translated through the pagetable
2. Check if maxlength > length
3. Check if there is indeed a string at pwstr with a length followed by a terminating NULL
4. If either of 1, 2 or 3 fails, check if the address provided by the user has a string with a terminating NULL, up to some pre-defined max length (say 4k).

K-Mayer:
> 
1 has a problem with a really short (aka 1 characters long) w-string. I don't know if this is a "valid" case or a would be because of a bug in the analyzed program, but it can happen. You would then try to read a PWSTR from random memory.


I think point 3 does not work since the strings might not be NULL-terminated
"Buffer

Pointer to a wide-character string. Note that the strings returned by the various LSA functions might not be null-terminated."
https://learn.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings

4 Of course we can try to do that.

Don't get me wrong, I like the idea of having just one function to read the strings, but I see quite a significant performance (and complexity) impact when trying to decipher what kind of string we are looking at with all the different methods.

Maybe we could extend the `vmi_read_unicode_str ` function with some kind of hint parameter in which we provide the type of string we want to extract (if we already know which kind it will be)?

tklengyel:
> I think all 4 points together could work most of the time, but I'm rather sure there will be cases where this is not the case

Sure, it could be the case, but please show me an example where it doesn't work and we can go from there.

K-Mayer:
> 

Just to avoid any misunderstandings:
You want an example which verifies the quote from the Microsoft documentation which states "Note that the strings returned by the various LSA functions might not be null-terminated."? That seems to be pretty hard without running into it in the wild since the documentation does not get any more specific than that (at least to my knowledge).

rageagainsthepc:
> If the string is not null-terminated you must have a pre-set limit for how much you'll fetch max, how are you dealing with that otherwise?

We are dealing with that in the same manner as `vmi_read_str()` ;) Regular wide strings have to be null terminated as opposed to strings with a corresponding unicode struct. Of course it makes sense to set a predefined limit for all string extraction methods since data can't be trusted, but that's unrelated to the problem at hand. We don't think it makes sense to add a heuristic if the user already knows what type of string should be extracted. It just adds unnecessary complexity and it is potentially slower. On top of that, the heuristic would not work every time. We posted links to docs that back our claims. Testing your heuristic for some random example doesn't make it correct for all cases. If you are not satisfied with this solution, feel free to adjust it in your own time, but we are not going to invest more time into this PR at this point.

tklengyel:
> If you are not satisfied with this solution, feel free to adjust it in your own time, but we are not going to invest more time into this PR at this point.

I think the request to validate whether the addition of a new API is mandated is entirely reasonable, but of course, suite yourself. FOSS for the win, feel free to do whatever you think is right for your use-case.

PR #1072, opened 2023-06-26
Symbol RVA translation for ELF executables and shared libraries

thomasdangl:
This PR is recent work on enabling Symbol ⇔ RVA translations for Linux guests.
It provides similar functionality to what is already there for Windows. However, instead of linearly scanning through the table, we use `DT_GNU_HASH` (and `DT_HASH` as a fallback).

Unfortunately, `DT_GNU_HASH` is undocumented. But it is what the dynamic linker is using and what is present in the main memory for that reason. Accordingly, the implementation is pieced together from reading the source code and some old mailing list conversations I found.

Should this fail, we use `DT_HASH` as a fallback. Note that while this is still present in many binaries that also contain `DT_GNU_HASH`, the dynamic linker does not use it. Hence the page is not paged in. It is only intended for older systems that don't have `DT_GNU_HASH`. For newer systems, one could inject a page fault, but I have not bothered with this as the aforementioned `DT_GNU_HASH` implementation works fine for this case.


thomasdangl:
Okay, seems like this one needed some changes for autotools, but now it looks good.

configure.ac
```diff
@@ -229,8 +229,6 @@ AS_IF([test x"$hardening" != x"no"], [
   check_cc_flag([-fstack-protector-all],
    [check_link_flag([-fstack-protector-all],
      [HARDEN_CFLAGS="$HARDEN_CFLAGS -fstack-protector-all"
-      check_cc_flag([-Wstack-protector], [HARDEN_CFLAGS="$HARDEN_CFLAGS -Wstack-protector"],
-        [], [-fstack-protector-all])
```
tklengyel, line 233: Why do you need to remove stack protector?

thomasdangl: The PR is not removing the stack protector. It just turns off warnings related to variable-length arrays not being protected by the stack protector.

With `-Wstack-protector` we get some hard errors in `elfparse_assign_tables`, `linux_export_to_rva`, and `linux_rva_to_export`. The first one is trivial to fix by allocating the array on the heap instead. The second and third ones look like false positives as the functions do not contain variable length arrays.
Besides, the CMake build doesn't set this flag either, so we should be fine.

tklengyel: Well, this warning is in place so that with -Werror you can't introduce functions that are not protected by the stack protector. If you were tripping this warning, you are adding functions that aren't getting stack protection. The solution is not to just disable the warning. If you REALLY are sure you are getting false positives the proper way to handle would be to add something like:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstack-protector"
    foo(b);         /* no diagnostic for this one */
#pragma GCC diagnostic pop

Comment (tklengyel) [reply configure.ac:233]: Also, I just did a build of your branch with `-Wstack-protector` in place and it builds just fine.

File: libvmi/debug.h [hunk 2]
```diff
@@ -51,7 +51,8 @@ typedef enum {
     VMI_DEBUG_WRITE     = (1 << 13),
     VMI_DEBUG_DRIVER    = (1 << 14),
     VMI_DEBUG_PEPARSE   = (1 << 15),
-    VMI_DEBUG_BAREFLANK = (1 << 16),
+    VMI_DEBUG_ELFPARSE  = (1 << 16),
+    VMI_DEBUG_BAREFLANK = (1 << 17),
```
tklengyel, line 55: Please don't change existing definition numbers, introduce new ones with bits not yet used.

thomasdangl: Mhh, okay. I would have preferred to keep PEPARSE and ELFPARSE together, especially if it's only a private header. But having ELFPARSE at bit 17 is also fine for me.

configure.ac
thomasdangl: I don't find this convincing. Why clutter the code with multiple `#pragma` directives to address a compiler bug? It just makes it harder for everyone to read.

I don't think it's a problem if we have a few functions in the library that don't have a stack canary. Especially when the CMake build not only doesn't have this flag either but also uses a weaker model with `-fstack-protector-strong`. So if you currently build it with CMake, you already have functions without a stack protector (see https://lwn.net/Articles/584225/).

After all, compiler protections against stack buffer overflows are always just a best-effort approach. Now if it were the stated goal of the library to have the highest amount of stack protector coverage possible, I might see things differently. But as of right now, we don't even have consistent behavior across build tools.

 `-Wstack-protector` works fine on my side as well, but not on the CI (https://github.com/libvmi/libvmi/actions/runs/5380958803/jobs/9764402176). If you take a look at this input/output, you can clearly see that the behavior is incorrect:

```
addr_t hash = 0, gnu_hash = 0, str = 0, sym = 0;
uint8_t image[1024] = { 0 };

if (VMI_FAILURE == elfparse_get_image(vmi, ctx, sizeof(image), image))
    return VMI_FAILURE;

elfparse_assign_tables(vmi, ctx, image, &hash, &gnu_hash, &str, &sym);

if (!str || !sym || (!hash && !gnu_hash)) {
    dbprint(VMI_DEBUG_ELFPARSE, "--ELFParse: failed to locate dynamic tables\n");
    return VMI_FAILURE;
}

/* approach 1: attempt to resolve the symbol through DT_GNU_HASH. */
if (gnu_hash)
    return gnu_hash_lookup(vmi, ctx, str, sym, gnu_hash, symbol, rva);

/*
 * note that on modern systems, the fallback will fail.
 * most executables and shared libraries will still contain DT_HASH,
 * however, since the dynamic linker only uses DT_GNU_HASH, DT_HASH is never accessed.
 * hence, the corresponding page is not present in main memory.
 * thus, the fallback is only intended for legacy systems (2006 and earlier).
 */
dbprint(VMI_DEBUG_ELFPARSE, "--ELFParse: falling back to DT_HASH resolving\n");

/* approach 2: attempt to resolve the symbol through DT_HASH. */
if (hash)
    return elf_hash_lookup(vmi, ctx, str, sym, hash, symbol, rva);

/* unreachable. */
 return VMI_FAILURE;
```
```
libvmi/os/linux/elfparse.c: In function ‘linux_export_to_rva’:
libvmi/os/linux/elfparse.c:339:1: error: stack protector not protecting local variables: variable length buffer [-Werror=stack-protector]
  339 | linux_export_to_rva(
      | ^~~~~~~~~~~~~~~~~~~
```
Neither this function nor any of the functions called within (and potentially inlined) use variable length buffers on the stack (in the commit of the linked job, not the current commit).

thomasdangl:
Okay, so after changing the enum as requested, the CI broke apparently because ubuntu.com responded with a 503. Can you maybe manually rerun it or do I have to do another force-push?

configure.ac
tklengyel: Not sure what's happening here but seems like it only has this false positive with the scan-build CI test. I would rather just disable this warning for that specific CI instead of all compilers since it works fine otherwise. You should be able to just add a CFLAGS="-Wno-stack-protector" to that particular build. I'm also somewhat perplexed if we are the first to run into this bug or not and what actually triggers it. Should obviously be reported upstream to get it fixed properly.

thomasdangl: I don't think this is the case. Here you can see the error for the compile CI test:
https://github.com/libvmi/libvmi/actions/runs/5380884457/jobs/9764232105

From what I can see the CMake build is fine as expected since it doesn't have `-Wstack-protector` and I think clang + autotools is also fine. So most likely is some issue with gcc, maybe even the specific gcc version on the CI.

thomasdangl:
Okay, you win. Let's use pragma directives so everyone is happy and we can move on :smiley:

libvmi/os/linux/elfparse.c
```diff
@@ -0,0 +1,422 @@
...
+    vmi_instance_t vmi,
+    const access_context_t *ctx,
+    const uint8_t * const image,
+    addr_t *hash,
+    addr_t *gnu_hash,
+    addr_t *str,
+    addr_t *sym)
+{
+    struct elf64_ehdr *elf = (struct elf64_ehdr *) image;
+    struct elf64_phdr ph[elf->e_phnum];
```
tklengyel, line 167: I did a bit of testing on this and the reason why `-Wstack-protector` trips on this function is this above. You are allocating an array on the stack that's variable size. That's not a good thing. You aren't sanity checking that `elf->e_phnum` is within a reasonable range either. Considering the header may be entirely malicious or malformed, that's a concern. You should switch this to a `g_slice_alloc0` after sanity checking and making sure it gets `g_slice_free`d before the function returns.

libvmi/os/linux/elfparse.c
```diff
@@ -0,0 +1,422 @@
...
+    const char *symbol,
+    addr_t *rva)
+{
+    addr_t hash = 0, gnu_hash = 0, str = 0, sym = 0;
+    uint8_t image[1024] = { 0 };
+
+    if (VMI_FAILURE == elfparse_get_image(vmi, ctx, sizeof(image), image))
+        return VMI_FAILURE;
+
+    elfparse_assign_tables(vmi, ctx, image, &hash, &gnu_hash, &str, &sym);
```
tklengyel, line 350: After a bit of testing it seems the stack protector trips because of the number of arguments being passed to this function. If you can reduce the number of inputs so that the function can take all inputs via registers the warning disappears. So for example instead of passing individual pointers for `&hash, &gnu_hash, &str, &sym`, pass a pointer to a structure with those fields, or simply a pointer to an array.

libvmi/os/linux/elfparse.c
```diff
@@ -0,0 +1,422 @@
...
+    for (; h < ph + elf->e_phnum; h++) {
+        if (h->p_type == PT_DYNAMIC)
+            break;
+    }
+
+    if (h->p_type != PT_DYNAMIC)
+        return;
+
+    for (_ctx.addr = ctx->addr + h->p_vaddr;
+            d.d_tag != DT_NULL;
```
tklengyel, line 186: I think there should be an upper bound on this loop so that a malformed header can't make this spin excessively.

libvmi/os/linux/elfparse.c
```diff
@@ -0,0 +1,422 @@
...
+     * note that on modern systems, the fallback will fail.
+     * most executables and shared libraries will still contain DT_HASH,
+     * however, since the dynamic linker only uses DT_GNU_HASH, DT_HASH is never accessed.
+     * hence, the corresponding page is not present in main memory.
+     * thus, the fallback is only intended for legacy systems (2006 and earlier).
+     */
+    dbprint(VMI_DEBUG_ELFPARSE, "--ELFParse: falling back to DT_HASH resolving\n");
+
+    /* approach 2: attempt to resolve the symbol through DT_HASH. */
+    if (hash)
```
tklengyel, line 371: Superfluous if-check, you can just `return elf_hash_lookup` at this point and get rid of the unreachable VMI_FAILURE further down.

libvmi/os/linux/elfparse.c
```diff
@@ -0,0 +1,422 @@
...
+
+    return VMI_SUCCESS;
+}
+
+status_t
+elfparse_get_image(
+    vmi_instance_t vmi,
+    const access_context_t *ctx,
+    size_t len,
+    const uint8_t * const image)
```
tklengyel, line 62: I think it's strange that you declare this function to take `const` pointer for image just to immediately discard the `const` qualifier. Should adjust the function prototype to just take `uint8_t *image` as input.

tklengyel:
> Okay, you win. Let's use pragma directives so everyone is happy and we can move on 😃

I did a bit more digging because I wasn't happy with not understanding why the stack protector warnings were popping up. While some warnings still seem to be just buggy (passing more then 6 inputs to a function shouldn't trigger the warning) other logic in the code does seem to be problemtic. So silencing the warning with the pragma isn't the best solution either. I left some comments how it could be fixed, shouldn't be too much change and then we can merge.

configure.ac
tklengyel: It's not specific to GCC, even if using clang as the backend during scan-build the warnings trigger.

libvmi/os/linux/elfparse.c
thomasdangl: I had already changed that in 2cf503a but reverted it back since the other warnings were still there. I'm not too happy with allocating it on the heap. That seems quite inefficient to me. Would you be happy with the sanity check and a fixed size allocation on the stack?

libvmi/os/linux/elfparse.c
thomasdangl: Yeah, I guess checking against malicious input makes sense. We should also consider doing the same for the PE format. I might send a PR for that later.

libvmi/os/linux/elfparse.c
thomasdangl: Okay, then we need to do the same for the PE format as well.

libvmi/os/linux/elfparse.c
thomasdangl: Good catch!

thomasdangl:
I can't manage to reproduce the issue with passing more than 6 parameters to a function. Regardless, the last refactor seems to have fixed the issue by accident. Let me know if you spot any further problems.

PR #1073, opened 2023-07-06
Sanitize input for PE parser

thomasdangl:
This adds input sanitization for the PE parser as discussed in https://github.com/libvmi/libvmi/pull/1072.


libvmi/os/windows/peparse.c
```diff
@@ -228,7 +233,7 @@ get_aon_index(
 
 status_t
 peparse_validate_pe_image(
-    const uint8_t * const image,
+    uint8_t *image,
```
tklengyel, line 236: We ought to be able to keep the `const` qualifiers here, no?

libvmi/os/windows/peparse.c
```diff
@@ -286,7 +291,7 @@ peparse_get_image(
 
 void
 peparse_assign_headers(
-    const uint8_t * const image,
+    uint8_t *image,
```
tklengyel, line 294: Here as well, `const` could be kept.

libvmi/os/windows/peparse.c
thomasdangl: In https://github.com/libvmi/libvmi/pull/1072 you specifically asked for them to be removed, just keeping it consistent here.

libvmi/os/windows/peparse.c
thomasdangl: See comment above

libvmi/os/windows/peparse.c
tklengyel: Not for the validate function, for the get function.

tklengyel: In the get function the memory pointer to be image clearly gets modified, so no `const` qualifier should be used. Here, the memory read-only for the validation, so const can be used.

tklengyel:
Thanks!

PR #1074, opened 2023-08-17
implement Windows PTE resolving

1ndahous3:
There is no support for various non-trivial p2v address translation mechanics for Windows images. This affects all parts of the hypervisor when p2v translation is not working, but in fact the translation is through various types of PTEs.

The patch adds Windows-specific routines to parse PTEs and improve in-memory translation.

SA:
- https://web.archive.org/web/20230609032449/http://blog.rekall-forensic.com/2014/10/windows-virtual-address-translation-and.html
- https://insinuator.net/2021/12/release-of-pte-analysis-plugins-for-volatility-3/
- https://blog.thalium.re/posts/windows-full-memory-introspection-with-icebox/


1ndahous3:
@tklengyel did I forget something for "ubuntu-22.04, gcc, autotools"?

libvmi/libvmi_extra.h
```diff
@@ -166,7 +166,7 @@ status_t vmi_get_struct_member_offset_from_json(
  *
  * @return VMI_SUCCESS or VMI_FAILURE
  */
-status_t vmi_get_bitfield_offset_and_size_from_json(vmi_instance_t vmi, json_object *json,
+status_t vmi_get_bitfield_info_from_json(vmi_instance_t vmi, json_object *json,
```
tklengyel, line 169: This is a public header and renaming an existing API function like this is no-go. If you strongly think the new API name should be used please introduce the new one and add a comment to the old one to say it's deprecated and to be removed in a future version. Then in a couple releases we can remove the old API.

tklengyel:
> @tklengyel did I forget something for "ubuntu-22.04, gcc, autotools"?

I can't spot what the issue is from the logs. I suggest you setup a docker to test it, it's reproducible.

tklengyel:
The remaining error is due to using APIs that are only available when LibVMI is compiled with either Rekall or Volatility JSON support, ie. if you compile with `--disable-volatility-ist --disable-rekall-profiles` the json related APIs won't be available. You have to adjust how you call those functions to make sure those are available.

libvmi/debug.h
```diff
@@ -35,27 +35,27 @@
...
-    VMI_DEBUG_READ      = (1 << 12),
-    VMI_DEBUG_WRITE     = (1 << 13),
-    VMI_DEBUG_DRIVER    = (1 << 14),
-    VMI_DEBUG_PEPARSE   = (1 << 15),
-    VMI_DEBUG_BAREFLANK = (1 << 16),
-    VMI_DEBUG_ELFPARSE  = (1 << 17),
-    VMI_DEBUG_OSX       = (1 << 18),
-
-    __VMI_DEBUG_ALL    = ~(0ULL)
+     VMI_DEBUG_MISC        = (1 << 0),
```
tklengyel, line 38: Extra space here before `VMI`

1ndahous3:
@tklengyel finally tested and ready to merge :)

libvmi/libvmi.h
```diff
@@ -687,13 +687,15 @@ typedef struct page_info {
 
     union {
         struct {
+            addr_t pte_value_prev;
```
tklengyel, line 690: Could you add a comment as to when this field is set? Perhaps adding a short description into`notes` would help as well where you put the information you have in the PR description. Thanks!

tklengyel:
Looks like some of the CI tests are failing for unrelated reasons. Could you still fix those up as part of this PR? Just need to add the glib2.0-dev package to list of dependencies that get installed.

PR #1076, opened 2023-09-06
Remove cached config entry as part of vmi_destroy function

thehoustonian:
When initializing and destroying a `vmi_instance` for the same VM multiple times in a program (such as in a loop), a warning message is printed: 
```
Duplicate config for application-vm found, using most recent
```
This comes from [this line](https://github.com/libvmi/libvmi/blob/8a79057d60ea8a8491230dda3284180fb2c811ca/libvmi/config/grammar.y#L222). 

This merge request cleans up this entry when calling `vmi_destroy`. 

This may seem like a weird workflow, since the example programs included with LibVMI run and then exit. However, if (for example) a program wanted to periodically calculate the process list for a VM, it could initialize a `vmi_instance` associated with that VM, read the process list, and destroy it all within a loop. This might be done to handle a VM rebooting and getting a different domain ID in Xen, which would require a re-initialization of the `vmi_instance` struct. In this case, I believe this warning message would appear. It might be limited to cases where `VMI_CONFIG_STRING` is used as the config mode though.

I don't fully understand the config parsing logic/workflow, so if this fix doesn't make sense just let me know.

### DIFF
--- a/libvmi/config/config_parser.h
+++ b/libvmi/config/config_parser.h
@@ -33,6 +33,7 @@
 #include <libvmi/libvmi.h>
 
 int vmi_parse_config(const char *target_name);
+void vmi_destroy_config();
 GHashTable* vmi_get_config();
 
 #endif /* CONFIG_PARSER_H_ */
--- a/libvmi/config/config_parser.h
+++ b/libvmi/config/config_parser.h
@@ -33,7 +33,6 @@
 #include <libvmi/libvmi.h>
 
 int vmi_parse_config(const char *target_name);
-void vmi_destroy_config();
 GHashTable* vmi_get_config();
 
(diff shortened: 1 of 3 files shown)


tklengyel:
I think the logic you are pointing to is supposed to optimize this exact workflow so that you don't have to reparse the config file every time and recreate the hash tables. I think it would just be sufficient to change that `fprintf` to be a ` dbprint(VMI_DEBUG_MISC,...`. Can you give that a try?

thehoustonian:
@tklengyel Thanks for pointing that out, I didn't see that. I changed that message to log as a debug message instead, and it works like I would expect. It doesn't show up on the terminal unless you have debug logging enabled

PR #1077, opened 2023-09-15
fix KPCR read on Windows x86

alex-pentagrid:
During initialization, the virtual machine can be stopped either in ring0 or in ring3. But on 32-bit windows if we are in ring3 - we cannot read `KPCR` from `fs_base` register, we must instead read them from `GDT`.

So, i add `CPL` check for 32-bit Windows, and if we are in ring3 - `KPCR` would be read from `GDT`. If we are in ring0 - `KPCR` would be read like before, from `fs_base` register.

For 64-bit Windows old behavior also persists.

### DIFF
--- a/libvmi/os/windows/core.c
+++ b/libvmi/os/windows/core.c
@@ -701,6 +701,50 @@ done:
     return ret;
 }
 
+static status_t kpcr_get_x86(vmi_instance_t vmi, reg_t *kpcr_reg)
+{
+    registers_t regs;
+    if (VMI_FAILURE ==  driver_get_vcpuregs(vmi, &regs, 0)) {
+        dbprint(VMI_DEBUG_MISC, "** driver_get_vcpuregs(..) failed.\n");
+        return VMI_FAILURE;
+    }
+
+    // check current privelege level for 32-bit Windows
+    if (regs.x86.cs_sel & 0x3) {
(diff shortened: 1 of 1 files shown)


libvmi/os/windows/core.c
```diff
@@ -834,10 +834,33 @@ init_from_json_profile_real(vmi_instance_t vmi, reg_t kpcr_register_to_use)
 
     if (kpcr_register_to_use) {
         reg_t kpcr_reg = 0;
-        dbprint(VMI_DEBUG_MISC, "** Trying kpcr_register_to_use to get KPCR.\n");
-        if (VMI_FAILURE == driver_get_vcpureg(vmi, &kpcr_reg, kpcr_register_to_use, 0)) {
-            dbprint(VMI_DEBUG_MISC, "** driver_get_vcpureg(..) failed.\n");
-            goto done;
+
+        if ((vmi->page_mode == VMI_PM_LEGACY || vmi->page_mode == VMI_PM_PAE) && kpcr_register_to_use == FS_BASE) {
```
tklengyel, line 838: Please split this logic into a separate function.

alex-pentagrid: Split logic into two separate methods, `kpcr_get` and `kpcr_get_x86`.

PR #1082, opened 2023-10-16
driver interface, xen: added possibility to map memory with specified prot flags

j3su5cr1st:


tklengyel:
Thanks, I think in this case an API change of `vmi_mmap_guest` would be perhaps a better way to go about it.

j3su5cr1st:
Sure, but I have choose this way to avoid any breaks in projects currently using libvmi. There's one another option, to make `prot` flags defaulted as last argument, but it's seems to me that this will add some implicity to code base, what do you think?

tklengyel:
Maintaining API compatibility is certainly important but in cases where the API is clearly in need of a fixup I prefer just doing that. Whether we default to PROT_READ or not in case prot is passed in as 0 is a matter of choice, it wouldn't really help with backwards compatibility (I think).

tklengyel:
Thanks!

PR #1083, opened 2023-11-13
New physical page api

rageagainsthepc:
Needs https://github.com/libvmi/libkvmi/pull/1

### DIFF
--- a/libvmi/libvmi.h
+++ b/libvmi/libvmi.h
@@ -2466,6 +2466,31 @@ status_t vmi_set_vcpuregs(
     unsigned long vcpu) NOEXCEPT;
 
 /**
+ * Allocates a new physical page.
+ *
+ * @param[in] vmi LibVMI Instance
+ * @param[in] gfn gfn of physical page to be allocated
+ */
+status_t vmi_alloc_gfn(
+	vmi_instance_t vmi,
+	uint64_t gfn) NOEXCEPT;
+
+
+/**
+ * Frees a physical page that has previously been allocated via
+ * vmi_alloc_gfn.
+ * Note that the maximum physical address cached by libvmi
+ * will not be updated accordingly. Freeing a page that has not
+ * been allocated via vmi_alloc_gfn is undefined behavior.
+ *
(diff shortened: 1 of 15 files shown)


libvmi/libvmi.h
```diff
@@ -2465,6 +2477,29 @@ status_t vmi_set_vcpuregs(
...
+ * @param[in] vmi LibVMI Instance
+ * @param[in] gfn gfn of physical page to be allocated
+ */
+status_t vmi_alloc_gfn(
+    vmi_instance_t vmi,
+    uint64_t gfn) NOEXCEPT;
+
+
+/**
+ * Frees a with vmi_alloc_gfn allocated physical page.
```
tklengyel, line 2492: Typo, `with` is unnecessary. Technically this could free any gfn, not just the one allocated with vmi_alloc_gfn, right?

libvmi/libvmi.h
```diff
@@ -2465,6 +2477,29 @@ status_t vmi_set_vcpuregs(
...
+ * Note that the maximum physical address cached by libvmi
+ * will not be updated accordingly.
```
tklengyel, line 2494: Why not adjust it if the gfn being removed is exactly where the `max_physical_address >> page_shift` is?

tklengyel:
Thanks, looks largely good, just a couple minor things I pointed out above.

libvmi/libvmi.h
rageagainsthepc: On kvm, pages allocated via `vmi_alloc_gfn` are tracked, actually. Calling `vmi_free_gfn` on a page that has not been allocated via `vmi_alloc_gfn` will fail.

libvmi/libvmi.h
rageagainsthepc: I think that decreasing the cached value in this specific case might even be more confusing than not decreasing it at all. In order to correctly adjust the value, we would either have to keep track of freed pages, or always retrieve a new one via the hypervisor API. If you care about having a correct value, I would suggest the latter approach. Personally, I didn't think it was important enough to be implemented.

PR #1089, opened 2023-12-19
CMake: Adjust scope of linked libraries

rageagainsthepc:
In CMake, targets should propagate public dependencies so the dependent target doesn't have to link against those manually.


libvmi/CMakeLists.txt
```diff
@@ -146,9 +146,10 @@ if (ENABLE_KVM)
...
             endif ()
         endif ()
-        target_include_directories(vmi_shared PRIVATE ${JSON-C_INCLUDE_DIRS})
+        target_include_directories(vmi_shared PUBLIC ${JSON-C_INCLUDE_DIRS})
         # CMAKE_DL_LIBS -> dlopen* lib
-        target_link_libraries(vmi_shared PRIVATE ${CMAKE_DL_LIBS} ${JSON-C_LIBRARIES})
+        target_link_libraries(vmi_shared PRIVATE ${CMAKE_DL_LIBS})
+        target_link_libraries(vmi_shared PUBLIC ${JSON-C_LIBRARIES})
         list(APPEND VMI_PUBLIC_HEADERS events.h)
         list(APPEND VMI_PUBLIC_DEPS ${CMAKE_DL_LIBS} ${JSON-C_LIBRARIES})
```
rageagainsthepc, line 154: Any idea why `libdl.so` is listed as a public dependency? As far as I am aware it should be private, right?

Wenzel: Indeed, it shouldn't be a public dependency, we only use dlopen internally
https://github.com/libvmi/libvmi/blame/master/libvmi/CMakeLists.txt#L150

Wenzel:
LGTM

PR #1090, opened 2023-12-19
Xen: fix calculation of ram size

rageagainsthepc:
I've noticed that `vmi_get_memsize()` reports vastly more memory than the VM should actually have. Afais the correct value is stored in `xen->info.tot_pages` instead of `xen->info.max_pages`. I suppose `xen->info.max_pages` reports the maximum amount of RAM available for ballooning? I did not look into it too closely though.

### DIFF
--- a/libvmi/driver/xen/xen.c
+++ b/libvmi/driver/xen/xen.c
@@ -942,16 +942,12 @@ xen_get_memsize(
     if ( xen->major_version == 4 && xen->minor_version < 6 )
         xen->max_gpfn = (uint64_t)xen->libxcw.xc_domain_maximum_gpfn(xen->xchandle, xen->domainid);
     else
-        xen->libxcw.xc_domain_maximum_gpfn2(xen->xchandle, xen->domainid, (xen_pfn_t*)&xen->max_gpfn);
+        xen->libxcw.xc_domain_maximum_gpfn2(xen->xchandle, xen->domainid, &xen->max_gpfn);
 
     if ( !xen->max_gpfn )
         return VMI_FAILURE;
 
-    // note: may also available through xen_get_instance(vmi)->info.max_memkb
-    // or xenstore /local/domain/%d/memory/target
-    uint64_t pages = xen->info.max_pages + xen->info.shr_pages;
-
-    if ( !pages )
+    if ( !xen->info.tot_pages )
         dbprint(VMI_DEBUG_XEN, "--Xen reports no pages being allocated for the domain\n");
 
     /* For Xen PV domains, where xc_domain_maximum_gpfn() returns a number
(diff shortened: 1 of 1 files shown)


rageagainsthepc:
Tested on xen version 4.14.5, if that helps.

tklengyel:
This is incorrect. `vmi_get_memsize` needs to return the maximum addressable memory for the guest, not the currently allocated number of pages.

rageagainsthepc:
Just to clear up any confusion for myself: If I give my VM 4 GB of RAM, what value would you expect `max_pages` to be? I would have expected it to be around 4GB because that's the maximum amound of RAM I have set. But it is actually much higher than that. This can't be right, can it?

rageagainsthepc:
Ok, so `max_pages` contains `0xffffffff` apparently, so there is probably something broken.

rageagainsthepc:
I can trace it down to this PR: https://github.com/libvmi/libvmi/pull/1081. Before that `vmi_get_memsize()` reports the correct value, afterwards it's broken.

tklengyel:
Memory of a virtual machine is not continuous. You have holes in the memory reserved for various devices, hypervisor pages, etc. So you have to distinguish between "what's the maximum physical address this VM may use" and "currently allocated memory". You can assign a VM 4GB of memory, but that won't translate to physical addresses 0-4gb. Check https://github.com/libvmi/libvmi/blob/master/notes/memory_map.txt. The get_memsize function is primarily used for the first purpose. Switching it to `tot_pages` is incorrect as it would just add up the pages that are not holes. It is possible `max_pages` returns 0xffffffff right now, would need to check the Xen source why it would do that.

tklengyel:
Also FYI, with DRAKVUF we always set max_pages to ~0 because we don't care and want to avoid LibVMI sanity checks failing because the added shadow pages are not included in the initial memory size cached during startup. Make sure you are not testing on a VM you ran DRAKVUF on cause there it would be expected behavior.

rageagainsthepc:
> Memory of a virtual machine is not continuous. You have holes in the memory reserved for various devices, hypervisor pages, etc. So you have to distinguish between "what's the maximum physical address this VM may use" and "currently allocated memory". You can assign a VM 4GB of memory, but that won't translate to physical addresses 0-4gb. Check https://github.com/libvmi/libvmi/blob/master/notes/memory_map.txt. The get_memsize function is primarily used for the first purpose.

I am well aware of this, but if `vmi_get_memsize()` is intended for the first case then its documentation would be very confusing because it currently states:
"This is the amount of RAM allocated to the guest, but does not necessarily indicate the highest addressable physical address;"
Also, how do you explain that the behavior changed with https://github.com/libvmi/libvmi/pull/1081? This is either a breaking change or a regression, depending on whether this was intentional or not.

> Make sure you are not testing on a VM you ran DRAKVUF on cause there it would be expected behavior.

I am using plain libvmi.

tklengyel:
> I am well aware of this, but if `vmi_get_memsize()` is intended for the first case then its documentation would be very confusing because it currently states: "This is the amount of RAM allocated to the guest, but does not necessarily indicate the highest addressable physical address;"

My bad, mixed up which API function is which. We've split this function into two to have `get_max_physical_address` as a separate function. So in this case you can use tot_pages but you also have to add shared pages to it, otherwise forks would report a memory size of 0.

rageagainsthepc:
Ok, so does it make sense to reopen this PR then? Or should I create a new one?

tklengyel:
You can reopen this one

rageagainsthepc:
@tklengyel I can't. Afaik you need to be a maintainer to reopen a PR.

tklengyel:
Since you deleted the branch it won't let me, just open a new one

PR #1092, opened 2023-12-26
Xen: fix calculation of ram size

rageagainsthepc:
Reopened as discussed in https://github.com/libvmi/libvmi/pull/1090


rageagainsthepc:
Btw, I am still not sure whether the calculation for PV domains is correct or not.

PR #1093, opened 2024-01-05
Add alternative memory mapping API

rageagainsthepc:
When mapping large memory regions that are only sparsely populated `vmi_mmap_guest()` may take a long time (> 30s) because it tries to resolve the virtual address of each page separately. Roughly 2/3 of the execution time is spent doing hashtable lookups because of the v2p cache (assuming it is enabled). Also, the API requires the caller to allocate an array of size `num_pages` and will allocate an array with the same size for storing `pfns` on top of that. This results in a lot of wasted memory for large, sparsely populated memory regions.
This PR introduces an alternative API which directly iterates over any set of page tables and finds chunks of consecutive memory. These regions will be mapped and returned as an array of memory region descriptors. In my case, mapping everything that is accessible via the kernel page tables took about a second (4GB Windows VM).
However, the new API may not produce the exact same results as the old API due to differences in the translation mechanism. For example, on x86 Windows consecutive entries pointing to the same `pfn` will be discarded.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -263,7 +263,8 @@ endif
                        examples/breakpoint-emulate-example \
                        examples/read-disk-example \
                        examples/io-event-example \
-                       examples/switch-view-example
+                       examples/switch-view-example \
+                       examples/memory-mapping-example
 
     examples_map_symbol_SOURCES = examples/map-symbol.c
     examples_map_addr_SOURCES = examples/map-addr.c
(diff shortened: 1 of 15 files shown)


rageagainsthepc:
Hi, just chiming in, mostly to say I don't have any stakes in this anymore. ;) I don't even have access to the branch.
However, the only lint was the `Poorly documented large function` one. I have only slightly modified the function in question which was poorly documented to begin with. The changes I made where rather trivial and didn't need any comments imho. The other option would habe been to add random comments to umodified lines which I deemed out of scope for this PR. Honestly, I didn't think that this lint would be a deal breaker. I thought it was fine to ignore it in this specific case.
Anyways, you guys have fun with this PR. I am sure you will figure it out. 😊

PR #1098, opened 2024-07-21
Makefile.am: Add logic to prevent a duplicate call to event.h.

SigmaSquadron:
Currently, if both KVM and Xen are enabled during build time, `event.h` is called twice, often ending in an error like the following:
```bash
install: will not overwrite just-created <installPath>/event.h with libvmi/event.h
```

This PR removes the duplicate call in KVM's `if` condition, and adds an extra `if` condition at the end of the `drivers =` stanza that calls `event.h` if it has not already been called by Xen, and if Legacy KVM support is not enabled.

This patch was first used in [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs/pull/328873) to build `pkgs.libvmi`. If this PR is merged, we can drop our downstream patch.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -79,7 +79,6 @@ if WITH_KVM_LEGACY
     drivers     += libvmi/driver/kvm/kvm_legacy.c
 
 else
-    h_public    += libvmi/events.h
     drivers     += libvmi/driver/kvm/kvm.c \
                    libvmi/driver/kvm/kvm_events.c \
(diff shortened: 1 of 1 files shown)


tklengyel:
Thanks for the PR but I would prefer something more like https://github.com/libvmi/libvmi/commit/f04fbb9158fcef31a69dc304a49c1e4a2ce63c0e. Could you check if that works for you?

SigmaSquadron:
That looks great, thank you! Closing.

PR #1106, opened 2024-08-26
Add vmi_get_domain_status and Xen driver implementation

thehoustonian:
This PR adds a `vmi_get_domain_status()` function to `libvmi_extra.h` which allows users to retrieve a domain's status via introspection. This is particularly useful when a DomU reboots on Xen (at least on ARM) because the DomU will have a different domain ID when it comes up after reboot. Xen still maintains a record of the domain, so its status can be queried and flags checked to determine that the DomU is no longer running. This can allow introspection programs to destroy and re-initialize a `vmi_instance_t` when the original domain it was associated with reboots or otherwise becomes unavailable.

### DIFF
--- a/libvmi/libvmi_extra.h
+++ b/libvmi/libvmi_extra.h
@@ -192,6 +192,48 @@ status_t vmi_get_struct_field_type_name_from_json(
     const char **member_type_name) NOEXCEPT;
 #endif
 
+/**
+ * @struct xen_domain_status
+ * @brief Struct contain Xen-specific status flags
+ * Field values are set based on the values in the
+ * xen_domctl_getdomaininfo_t struct's flag field.
+ */
+typedef struct xen_domain_status {
+    bool dying:1;                     /**< set if XEN_DOMINF_dying is set */
+    bool shutdown:1;                  /**< set if XEN_DOMINF_shutdown is set */
+    bool paused:1;                    /**< set if XEN_DOMINF_paused is set */
+    bool blocked:1;                   /**< set if XEN_DOMINF_blocked is set */
+    bool running:1;                   /**< set if XEN_DOMINF_running is set */
+    bool debugged:1;                  /**< set if XEN_DOMINF_debugged is set */
+    bool xs_domain:1;                 /**< set if XEN_DOMINF_xs_domain is set */
+    bool hardware_assisted_paging:1;  /**< set if XEN_DOMINF_hap is set */
+} xen_domain_status_t;
+
+/**
+ * @struct domain_status
+ * @brief Struct containing status flags for a domain.
+ * Also contains hypervisor-specific flags in sub-structs
+ */
+typedef struct domain_status {
+    bool missing:1; /**< Indicates whether the domain was found */
+    vmi_mode_t type; /**< Hypervisor type, used to determine which fields exist in the union below */
+    union {
+        xen_domain_status_t xen_domain;
+    };
+} domain_status_t;
+
+/**
+ * Get the execution status of the domain associated with the provided vmi_instance
+ *
+ * @param[in] vmi LibVMI instance
+ * @param[out] domain_status domain_status_t containing the status flags for the given domain
+ *
+ * @return status_t result of retrieving the domain's status.
+ */
+status_t vmi_get_domain_status(
+        vmi_instance_t vmi,
+        domain_status_t *domain_status);
+
 #pragma GCC visibility pop
 
 #ifdef __cplusplus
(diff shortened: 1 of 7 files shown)


tklengyel:
I have no technical issue with the PR but it begs the question what's the point of doing this if it really just passes through the same information you could get by calling libxc directly? There is no abstraction/unification achieved in the LibVMI API with this. If instead of having a union and then hypervisor specific status bits you actually defined a common set of statuses that can be reused among the various supported hypervisors I would see more reason to add this to the library. As-is, it seems to be just a superfluous layer with no added benefit/utility.

thehoustonian:
@tklengyel That is a fair point. I had chosen the struct with a union instead of just a plain bitmask with #defines for the fields  in case it wasn't possible to determine all of the state flags for a given hypervisor (for example, if other hypervisors don't have the concept of dying or blocked like Xen does). This seemed like the cleanest way to convey which status fields are supported for a given hypervisor type. However, maybe a better approach would have been to define a struct of `int`s which are set to 0 or 1 for the status, or-1 if the given field is not supported for that hypervisor type? That way it's not tied to Xen specifically and could just expose the more common statuses of dying, shutdown, paused, blocked, and running. 

The only other benefit with this approach and not calling libcx directly is that programs using LibVMI would not have to also link against libxc directly, but that may be a marginal benefit at best. 

Either way, thanks for the feedback. If you'd be open to this functionality with common status fields on a struct directly, that seems like a worthwhile refactor to me

tklengyel:
> If you'd be open to this functionality with common status fields on a struct directly, that seems like a worthwhile refactor to me

Yea, I think that sounds like a worthwhile addition. It would probably be also fine to not incorporate all Xen status bits in the LibVMI abstraction if it's not useful definition for a "status" API (I'm looking at xs_domain & hap and both seem like a weird thing to have for status).

libvmi/libvmi_extra.h
```diff
@@ -192,34 +192,20 @@ status_t vmi_get_struct_field_type_name_from_json(
...
+ * Fields that are true are set to 1.
+ * Fields that are false are set to 0.
  */
 typedef struct domain_status {
-    bool missing:1; /**< Indicates whether the domain was found */
-    vmi_mode_t type; /**< Hypervisor type, used to determine which fields exist in the union below */
-    union {
-        xen_domain_status_t xen_domain;
-    };
+    int dying:2;     /**< Set if the domain is dying */
```
tklengyel, line 203: Please use `unsigned int` in the bitfield

thehoustonian: Could it be `signed int`, since I was thinking `-1` could be used in the event that a field was not supported on another hypervisor type?

tklengyel: How would that work?

thehoustonian: I wasn't sure if all of these fields would make sense for all hypervisor types. Setting one of the fields to `-1` could be used to indicate that status flag is not applicable for a given hypervisor type. Maybe that's overthinking it though? I don't have a problem changing these to 1-bit unsigned int fields instead

libvmi/libvmi_extra.h
```diff
@@ -192,6 +192,34 @@ status_t vmi_get_struct_field_type_name_from_json(
...
+
+/**
+ * Get the execution status of the domain associated with the provided vmi_instance
+ *
+ * @param[in] vmi LibVMI instance
+ * @param[out] domain_status domain_status_t containing the status flags for the given domain
+ *
+ * @return status_t result of retrieving the domain's status.
+ */
+status_t vmi_get_domain_status(
```
tklengyel, line 219: I also just noticed that you placed this into `libvmi_extra.h`, generally that header only contains APIs that require an additional library to be linked with in the caller (like libjson or libglib). As this API doesn't have such external dependency you can just place it in `libvmi.h`

libvmi/libvmi_extra.h
tklengyel: That sounds fine although I'm not sure under what circumstance would you actually check for that unsupported value for a status. Doing every time during run-time sounds like an unnecessary overhead. If its something you need to know it probably would be something you want to figure out at initialization time and set things up accordingly, no?

thehoustonian: Yeah, that's a good point. It also seems likely that any additional hypervisor drivers that implement support for this function would set status bits appropriately, even if (for instance) `dying` doesn't directly translate. I will update this to be unsigned int bitfields.

libvmi/libvmi_extra.h
thehoustonian: Oh, that makes sense. I will move it into the `libvmi.h` header then

